PR #75, opened 2015-01-05
improved assembly control

jgangemi:
after much delay, i'm putting this up for review so i get it finished. i still need to add in 'user' support and update the docs for that but that shouldn't take more then a couple hours.

i also think i forgot to rebase before i pushed, so i may need to recreate this request. :)

happy new year!


rhuss:
I did a quick review, and it looks very good to me (thanks a lot !)

Some remarks:

* In [DockerAssemblyConfigurationSource](https://github.com/rhuss/docker-maven-plugin/blob/7af57fd958679ddad84aa3315a803ef32531d186/src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyConfigurationSource.java#L37-37), would it be possible to prepend the source directory only if the given descriptor path is **not** an absolute file path ? That way we could easily support all use cases.
* IMO `docker.target.dir` is not necessary to be able to specify since I consider this directory as a temporary, internal directory within `target`. Do you have a use case where this configuration option is required ?
* /wrt `dryRun` option: Are you sure that will work ? I have the feeling (though not verified) that the directory with the Dockerfile wouldn't be created but `docker:build` still would try to make a Docker Image (then on an empty directory). Also, I don't see a good use case for that, so I would skip it (and set `dryRun` to always `false`).
* As you said, the `user` functionality is currently not implemented. Are you currently working on it ? If not I could have a look at it.

Please let me know when you considered to be finished with the PR, so that I can integrate it.

jgangemi:
hey - sorry for the late response, i was traveling again. let me get settled and i'll respond.

rhuss:
Do you have a rough timeframe for me ? I ask, because I want to release a 0.11.0 release ASAP.

If you don't have any time, no problem. I would do then the following:

* Check for absolute paths in DockerAssemblyConfigurationSource
* Leave out docker.target.dir, dryRun and user. We can add this then in a later release.

Is this ok for you ?

jgangemi:
hey - sorry, still trying to settle back in. i was going to tackle this over the weekend, that ok?

rhuss:
No worries. I plan for a release on Monday then. 

Nice weekend .....

jgangemi:
closed in favor of #81, which should be properly rebased

PR #76, opened 2015-01-11
pull base images (from) before building

kuhnroyal:
Currently I did not manage to build an image from a base image when the base image is not available locally and the private registry requires authentication.
For example building my.registry.com/example:latest from my.registry.com/dependant:latest.

The problem is that the BuildMojo is not adding the authConfig to the request. I tried adding it but to no effect on the docker daemon side. Thus now the "from" image is being pulled before the actual image is being build.
After that I had the problem, that during pull the "registry" parameter was not being honored by the docker daemon, prefixing the image name with the registry name works however.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/BuildMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/BuildMojo.java
@@ -61,6 +61,10 @@ public class BuildMojo extends AbstractDockerMojo {
 
     private void buildImage(ImageConfiguration imageConfig, DockerAccess dockerAccess)
             throws DockerAccessException, MojoExecutionException {
+        String fromName = imageConfig.getBuildConfiguration().getFrom();
+        info("Pulling base image " + fromName);
+        dockerAccess.pullImage(fromName, prepareAuthConfig(fromName));
+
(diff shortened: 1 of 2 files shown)


rhuss:
Thanks for the pull request !

You are right, a base image should be pulled if the option `autoPull` is enabled. This is currently not supported. Your solution to unconditionally pull the base image when doing the build is a  bit to much, though. I'll open an issue for this.

The registry handling however has been improved quite a bit alread, although not yet released. You can try out `0.11.0-SNAPSHOT` (which is available in Maven central), which already includes the private registry support as documented in the [manual](https://github.com/rhuss/docker-maven-plugin/blob/integration/doc/manual.md#registry-handling). I would be curious, whether this works for you, so that we could tune it before the release.

kuhnroyal:
Sounds good, i'll give it a try.

kuhnroyal:
I ran into 2 problems:
- PushMojo does not use docker.registry or imageConfig.registry for building the authConfig, thus the push failes
- autoPull doesn't seem to be true by default also it is documented that way

Also I think the global registry could get problematic when the "from" image is in a different registry, maybe something like the following is needed.
```xml
<from>
   <name></name>
   <registry></registry>
</from>
```

Issue #78, opened 2015-01-13
With big user IDs, an error "user id is too big" is thrown


thomas-endres-tng:
When using user IDs bigger than 2097151, the Maven assembly plugin throws the following error:

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.jolokia:docker-maven-plugin:0.10.5:build (default-cli) on project customer-lookup-web: Execution default-cli of goal org.jolokia:docker-maven-plugin:0.10.5:build failed: user id '148234567' is too big ( > 2097151 )

The problem is due to the fact that the TarArchiver uses short user IDs when not set to mode TarLongFileMode.posix. This has to be done in DockerArchiveCreator.createDockerBuildArchive().

This causes the docker:build process to throw an error and stop.

rhuss:
Thanks a lot. I will fix this issue ASAP. A release is scheduled for Friday (0.11.0) which will include this fix. The will be some minor configuration syntax changes for this release, though. But the migration is trivial (and documented).

Greetings to Tsvetomir, BTW ;-)

Thanks again ....

brmakana:
Hi rhuss,

Could you merge this to the main branch and make it available from maven central?

rhuss:
I will do a release on Monday. In the meantime, you can refer to 0.11.0-SNAPSHOT, which I just pushed to Maven central (might take a bit until is available). Please let me know if there any issues.

It is also available on branch `integration`.

brmakana:
Thanks! I'm hitting refresh all day on maven central hoping to see it, but in the meantime have been using your 0.10.6-SNAPSHOT in local repository.

PR #79, opened 2015-01-13
long file mode POSIX

thomas-endres-tng:
The tar archiver uses long file mode POSIX so that users with big UID get no error anymore (fixes issue #78)


rhuss:
I added the fix manually to branch `integration`. Thanks again !

PR #81, opened 2015-01-18
improved assembly control redux

jgangemi:
closed #75 in favor of this (rebased, at least i think) pull request. :) i know you want to get out a release and i've just been a little busier then expected.

    i took out dryRun, you're right that wasn't really going to gain anything.
    i did leave in the target dir override. the use case is if you build w/ tycho which uses the default eclipse (bin) directory structure. it's not really intended to override the internal docker name. this is how the build element works, so i was just sticking w/ that convention.

i'm a little confused w/ the absolute paths - i thought doing that would make the build non-portable and you were against that idea. i'm not going to argue against it so feel free to add it in.

i also realized i had a question about the user functionality...where should that command be called? before adding files? before env is called? might be best to delay this one a little pending further discussion.


rhuss:
Sounds all good. For the absolute vs. relative path thingy, I mean to use sth like `${project.baseDir}/src/main/resources/myDocker`, which is an absolute path when translated. So you are free to change the default (and this is still portable).

Now, I'm a bit busy ;-), but I'll try to cut a release on wednesday. After that we can discuss about the "user" stuff.

Thanks a lot ....

jgangemi:
gotcha - ok - i can do that real quick, it's just adding an `isAbsolute` check.

jgangemi:
now who's the busy one? :)

rhuss:
;-) I already integrated everything into branch `integration`. Still busy, hope I can to it today, if not, then on Wednesday. 

But it will be better soon ...

jgangemi:
haha - no rush on my side, the snapshot version works just fine for now.

rhuss:
0.11.0 released. Enjoy it and thx for the fish ;P

PR #86, opened 2015-02-04
fix for stopping containers using the 'latest' tag

jgangemi:
containers tagged with `latest` are not properly handled by the `docker:stop` goal.


rhuss:
I added the change to `integration`, however did also some tweaks. Could you please verify whether it works for you ? (we still need unit tests for the DockerAccessClient so its not so easy to verify this by hand).

jgangemi:
this is working for me, sorry i missed this before.

Issue #77, opened 2015-01-11
Respect "autoPull" when building images


rhuss:
When building images, the base image specified with `from` should be pulled as well if `autoPull` is set. Currently `autoPull` works only for `docker:start`.

See also #76

jgangemi:
where does this and #76 fall on your priority list / what you are currently working on? 

i am definitely going to need this in the coming days/weeks and don't want to duplicate efforts if you're already working on it.

jgangemi:
lol - looks like someone else has tackled this :)

kuhnroyal:
Yea, please check if that fixes your issues.

rhuss:
Fixed in 0.11.1

PR #88, opened 2015-02-05
fix #76 #77

kuhnroyal:
* allows autoPull to be specified on build
* make sure DockerAccess.hasImage() respects the tag and thus finds the correct image, before it only looked for the image name (repository) and this failed if the correct tag was not present
* BuildMojo now always does a manual pull of the 'fromImage' if it does not already exist or autoPull is enabled. This is necessary because the automatic pull that docker does itself during the create fails to authenticate against a private registry.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/BuildMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/BuildMojo.java
@@ -11,6 +11,7 @@ import org.jolokia.docker.maven.access.DockerAccessException;
 import org.jolokia.docker.maven.assembly.DockerAssemblyManager;
 import org.jolokia.docker.maven.config.BuildImageConfiguration;
 import org.jolokia.docker.maven.config.ImageConfiguration;
+import org.jolokia.docker.maven.util.ImageName;
 import org.jolokia.docker.maven.util.MojoParameters;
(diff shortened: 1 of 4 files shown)


kuhnroyal:
Currently the documentation states that autoPull is enabled by default. As far as I can tell by the code and tests, this is not true. I would like to see it enabled by default.

kuhnroyal:
Found the problem with the default value for autoPull.
Changed 'defaultValue' to 'default-value'.

jgangemi:
:+1:

kuhnroyal:
Rebased into one commit and signed.
Any thoughts on the log statements? Too chatty?

rhuss:
Thanks a lot !

Give me a bit, I will review it while integrating into branch `integration`. Will do it probably this weekend, but no promises ;-)

rhuss:
Thanks a lot. I applied the PR, (tuned a bit the logging ;-) to branch `integration`. I have however still one open question. Please see this [comment](https://github.com/codecrafters/docker-maven-plugin/commit/cbfcd4b09622182271867cd3bfc339f34fee8942#diff-94a065d53f48d5a87e52cc100e578cdbR78).

PR #89, opened 2015-02-05
fixed plugin can not find assembly descriptor

kuhnroyal:
The plugin could not find the assembly.xml file in the default source directory in multi-module maven projects.
It works if docker:build is executed from the sub-module but not if execute from the parent module.
When building the absolute path to the assembly.xml or Dockerfile the project baseDir is now respected.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyConfigurationSource.java
+++ b/src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyConfigurationSource.java
@@ -32,9 +32,9 @@ public class DockerAssemblyConfigurationSource implements AssemblerConfiguration
     @Override
     public String[] getDescriptors() {
         String descriptor = assemblyConfig.getDescriptor();
-        
+
         if (descriptor != null) {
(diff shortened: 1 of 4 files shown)


rhuss:
Awesome. Thanks a lot for the fix. Could you please do me a favor and sign-off your commits as explained [here](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md) ? Please also for you other PRs which I'm going to integrate today.

You simply need to do a 

```
git commit -s --amend
```

on this bug fix branch (works for the last commit, if you need to sign-off more commits in retrospective, you need an interactive rebase (git -i rebase))

Thanks a lot !

jgangemi:
do you want to do a quick bugfix release w/ these and maybe slate new functionality for 0.12.0?

rhuss:
Yes, think this is a good idea. I will do a a 0.11.2 this weekend.

jgangemi:
was there a 0.11.1 that i missed?

rhuss:
Ehm, no. I was wrong, no 0.11.1 yet, so that's the next one.

rhuss:
Applied and available on branch `integration`

PR #94, opened 2015-02-16
allow hostnames to be specified in bind configuration

jgangemi:
allows hostnames to be specified in the bind configuration. this lets me specify the prod hostname in the configuration and override it w/ `/etc/hosts` locally instead of having to do it via maven properties which becomes a little unwieldy at times.

also - looks like i've got changes from master that aren't in integration yet.


jgangemi:
going to resubmit...

PR #95, opened 2015-02-18
restart policy bug fix

jgangemi:
i did the other pull request on my master by accident and i don't have the time to sort that out right now which is why this contains both. once you resolve the other, this should be ok...or just pick this one. :)


jgangemi:
closing this - going to fix my mistake now that there are other issues to tend to.

Issue #97, opened 2015-02-18
docker:build to use a temporary directory inside the ${basedir}/target to avoid multi-project build issues


jstrachan:
I've just hit an issue where a second docker image in a build got jars from a previous docker build :) 

It seems like the temporary folder docker:build is using doesn't take into account the current ${basedir} from the project module.

It'd be good to make the default use ${basedir}/target/docker/... or something?

jstrachan:
I think we need to add ${basedir}/ to these properties
https://github.com/jstrachan/docker-maven-plugin/blob/master/src/main/java/org/jolokia/docker/maven/BuildMojo.java#L50-50
https://github.com/rhuss/docker-maven-plugin/blob/master/src/main/java/org/jolokia/docker/maven/BuildMojo.java#L45

jgangemi:
i thought `$basedir` was always assumed if the path is relative

jstrachan:
It doesn't seem to be in a multi-maven build; I can see the docker plugin using the exact output folder for every module ;)

jgangemi:
yeah - i just noticed that myself but adding in `$basedir` where you suggested doesn't work but i have another idea.

jgangemi:
ok - i have a fix for this. let me put together a pull request.

jstrachan:
great thanks!

FWIW as a workaround I added this to my pom.xml:
```
        <docker.target.dir>${basedir}/target/docker-plugin</docker.target.dir>
```
https://github.com/fabric8io/quickstarts/blob/master/pom.xml#L79

which seems to be a work around

jgangemi:
fixed - if you want to build a local copy, feel free to pull from my branch.

jstrachan:
thanks @jgangemi

rhuss:
fixed in 0.11.2

PR #98, opened 2015-02-19
append $basedir to all output directory paths

jgangemi:
fix for issue #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Flattr](http://api.flattr.com/button/flattr-badge-large.png)](http://flattr.com/thing/73919/Jolokia-JMX-on-Capsaicin)
 
 This is a Maven plugin for managing Docker images and containers for your builds.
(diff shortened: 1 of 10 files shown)


kuhnroyal:
Noticed this as well today, I only fixed it for the source directory before.
LGTM

jgangemi:
yeah - i actually just realized something else is potentially wrong w/ this - if you build two images, the data from one will override the data from the other in the output dir, so you loose the ability to see what was done for previous images.

i'm going to add an additional enhancement to this that will include the image name as part of the output directory path.

jgangemi:
to prevent annoying merge conflicts this should be pulled after #105

kuhnroyal:
Yea nice, ran into the problem you described last night. Lucky for me I have different folder names inside the 3 images so it is just added weight but not overridden.

jgangemi:
this was closed w/o a merge - was that correct? re-opening for now...

rhuss:
ah, yes. sorry.

PR #103, opened 2015-02-19
Fixes for two JUnit tests on Windows

Burre83:
Fixed two JUnit-test that were failing on Windows due to a difference in path separator chars (was using "/", was expecting "\"). It is now platform independent.


rhuss:
Thanks a lot. I applied you PR to branch `integration`. I assume also, that the APL for your [contribution](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md) is fine, too.

Burre83:
APL is fine.

PR #104, opened 2015-02-19
Tagging during build/push

kuhnroyal:
This may not be complete and might just be a starting point for discussions.


rhuss:
IMO this looks good to me. Just doing a small review and will then integrate into `integration`.

PR #105, opened 2015-02-19
allow the assembly configuration block to be optional

### LINKED ISSUE #66: assembly descriptors should be optional
The target image may just want to set a few env entries or a different command
 
```
[ERROR] Failed to execute goal org.jolokia:docker-maven-plugin:0.10.4:build (build-docker-image) on project wildfly-camel-itests-openshift-domain: Error reading assembly: No assembly descriptors found. -> [Help 1]
```

jgangemi:
forgot to do this along side the volume enhancements, resolves #66.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Flattr](http://api.flattr.com/button/flattr-badge-large.png)](http://flattr.com/thing/73919/Jolokia-JMX-on-Capsaicin)
 
 This is a Maven plugin for managing Docker images and containers for your builds.
-The current version ist **0.11.0** and works with Maven 3.2.1 and Docker 1.3.0 or later.
+The current version ist **0.11.1** and works with Maven 3.2.1 and Docker 1.3.0 or later.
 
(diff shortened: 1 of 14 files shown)


jgangemi:
btw, i've created some extremely crude pom test cases in this project (https://github.com/jgangemi/docker-maven-test) and i've added you as a collaborator. 

if you'd like to use them, please feel free. you're also more then welcome to push the repository if you want. i don't think it's necessary to follow the same workflow for that, but we can if you want.

rhuss:
Thanks, I will have a look in this test project. Indeed, we desperately need some integration tests, especially for securing against Docker version upgrades. Maybe we can have a look at the setup of the integration tests of wouterd/docker-maven-plugin ?  (AFAIR they are quite comprehensive)

jgangemi:
i'll have a look - i think we could come up w/ some fairly good individual integration tests, the hard part is going to be getting the environment setup correctly to run. perhaps creating a simple, dedicated container that can be pushed/pulled from the hub would help then setup of the test can make sure the environment is good.

Issue #108, opened 2015-02-21
return error messages from docker daemon


jgangemi:
this will alleviate the need to re-run the maven command using `-X`.

rhuss:
applied in 0.11.2

PR #109, opened 2015-02-21
refactor logging code

jgangemi:
the first in a series of pull requests to address issue #108. 

i want to write a test that can run against the real docker client in an on demand fashion (will be `@Ignored` by default). it made more sense to refactor then create another implementation that was basically going to do the same thing.


jgangemi:
also, from a merge standpoint, this should probably be first in line to ease potential conflicts.

PR #110, opened 2015-02-22
return docker daemon log messages

jgangemi:
this is the 2nd in a series of pull requests to address #108. this contains the changes from #109 and is intended to be merged after it is.

the apache client is now encapsulated in it's own class along w/ the url building and response parsing. this should make it easier to introduce unit-tests for things going forward. 

there's also some warnings cleanup as well. i still want to refactor the log handler to use the delegate class as well, but that can wait until after this is merged in.


rhuss:
sorry, my fault did some real stupid thing now (re-created branch integration instead of rebasing).

Uh, holidays were too long ;-[

I will apply all the PRs manually now.

jgangemi:
lol - that's the point of a vacation :) hope you had a good time!

rhuss:
@jgangemi , first of all, thanks a lot for splitting up DockerAccess ! 
This was long overdue.

Here are some things I observed during the integration:

* ChunkedResponseHandler is to generic. It serves two purposes:
	* Collecting text chunks from responses with chunkedencoding
	* Processing JSON objects from this junks
	Both classes uses the same ChunkedResponseHandler interface with different types. In order to empahsize there different tasks I'd prefer different interfaces. Also, the entry point is only the <String> variant, the <JSONObject> instance is used inside.
* LogRequestor still uses RequestUtil.newGet() which should not be needed anymore because of ApacheHttpDelegate.
* In `DockerAccessWithHttpClient`: I think instead of directly logging `HttpRequestException` when throwing an `DockerAccessException` we should put the HttpRequestException into DockerAccessException as a nested exception and let bubble it up until it is finally logged by the user of the method.
* Instead of `statusCode, ... aditional` I would use `... statusCodes` in the signature and check that statusCodes contains at least one element (`ApacheHttpDelegate` (which is probably better named `HttpClientDelegate` to fit `DockerAccessWithHttpClient`))
* Why do you truncate the container id to 12 chars within `createContainer()` ? We should use the full id internally and only truncate when printing them on screen. You'll neve know what will happen on the Docker side with the id, and changing a server side Id is IMO always a bad idea.
* A `ApacheHttpDelegate.getAsJson()` would avoid duplicate code in DAWHC when doing an HTTP Get for a JSONObject.
* Instead of an `assert` we should use proper `DockerAccessExceptions` (`getNewestImageForContainer()`)
* Thanbks for starting ITs for DockerAcces. Integration tests for `DockerAccess`should go in a separate src-level directory `it` or `itest` instead of `test` to separate them nicely, though.
* Just out of curiosity: Why do you always remove the 'else' when the 'if' part contains a `return` statement at the end ? Because of taste or  could it be that you use an IDE which can not handle this properly even when the else also contains a undconditional `return` at the end ? (IDEA detects this properly, btw).

In order to not cause too much trouble I suggest that I  push your PR nearly unchanged to branch integration, make a release, rebase integration and then you  integrate your other stuff and I then would do some more intense refactoring afterwards for the next release. Is this ok for you ?

Before you integrate your other changes: Please note, that I did a slight refactoring in `ImageName` with moving `getFullNameWithTag()` to `getFullName()` and `getFullName()` to `getNameWithoutTag()` since this loooks more reasonable for me (a "full name" is complete).

jgangemi:
i'll just try to comment on these in order ;)

the response handler interfaces seemed to be the exact same thing w/ the exception of the object they worked w/, so it seemed cleaner and easier to understand w/ just a single interface that used a generic. 

i know the LogRequestor still needs to be refactored. i was just holding off on this until this set of changes was merged.

i originally thought about this but i liked how the overall output looked doing it this way over trying to construct the proper `MojoException` or checking to see what the nested exception may be. given this is a closed ecosystem, what do you see is the benefit to bubbling it up only to be logged by whomever called it? 

the `(int, int...)` pattern is something i've seen a lot when requiring one or more params. why would you take the other approach? feel free to rename classes however you'd like - personally i'd prefer to see `DockerAccessWithHttpClient` be named `DockerHttpClient`, but that's just me. :)

docker states only the 1st 12 characters are needed to correctly identify a container, image, etc. no issues if you want to change that but we should make sure we're only ever display the first 12.

sounds like a good addition.

pretty sure this came from work done not by me, but i agree w/ you. :)

place wherever you'd like.

so that registers a compiler warning for me in eclipse - the `else` is unnecessary b/c of the `return` in the `if` and i'm a big proponent of clean code, so by habit i fix them as i find them. it's also why you'll see me list out individual packages i'm using instead of importing wildcard packages. i can turn off that warning specifically for the project if it bothers you - if nothing's drawing my attention to it i'll be less likely to notice. :)

as always, you are free to make whatever changes you want. i appreciate (and enjoy) being allowed to contribute on the level i do and at the end of the day, as long as the issue/feature gets in there, i'm happy.

i will pull from integration and merge my changes into that branch and fix whatever conflicts. shouldn't take long.

Issue #96, opened 2015-02-18
Provide forcePull option


jegr78:
It would be great to have an option, to force pulling an image. I'm currently having the issue, that my images already exist and so the pull option doesn't download the latest image from my private registry. I'd like to perform some kind of testing during CI (Jenkins with slaves) and need the possibility to force pulling of the latest availiable images from my private registry before running/starting the containers.

jgangemi:
the docker api already has an option to allow this so it just needs to be added in. depends on #108.

kuhnroyal:
This should work in 0.11.1.
See #88 and d77b597dc88a6caf25813595384a833452ae4b1e

jgangemi:
hrm - yeah - it looks like that option was added to the 1.17 api and doesn't exist in 1.15, so your solution handles more cases.

kuhnroyal:
What do you guys think about this?

```xml
<autoPull>
    <enable>true</enable><!-- defaults to true -->
    <force>true</force><!-- defaults to false -->
</autoPull>
```

jgangemi:
i would be ok with having a single tag that had multiple options (enum) of `off`, `enabled`, `force` where `force` implies `enabled but let's see what @rhuss says.

kuhnroyal:
True, that would work.

rhuss:
I'm also more for the enum approach (but more like `true`/`false`/`always` or `on`/off`/`always`) since having two independent options leads to non sensible combinations (like enable == false, force == true)

rhuss:
As mentioned in #112, `autoPull` with `always` is probably not  a good idea for the image which are build by this maven build when doing a `docker:start`.  With `autoPull=always` you will integration test images from the registry (if any is already pushed) and not the one just build (and stored in the docker host). 

So I wonder, whether a global 'forceMode' is a good idea anyway. It is probably more suited for being part of the `<image>` and `<run>` and `<build>` sections. 

If used globally I could imagine the following algorithm, when `autoPull == always`:

* For `<build>` always pull any base image.
* For `<run>` only pull images which don't have a `<build>` section, since these are supposed to be created outside this build and can safely be pulled always.

What's your opinion on this ?

Since I'm going to make a release today, I will save this feature for the next until we decided on the feature (and it would be cool to have autoPull support on all levels: (global), (image), (build and run) with more specific configs overwriting the more general ones.

rhuss:
[...]

PR #112, opened 2015-02-24
implemented #96 allow forcing off autoPull

kuhnroyal:
Implemented #96

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -3,6 +3,7 @@
 * **0.11.2**
   - Fix maven parse error when specifying restart policy (#99)
   - Allow hostnames to be used in port bindings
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks a lot. Going to refactor it a bit and merge into integration.

rhuss:
There's still an issue with `autoPull == always`: If this is set, for integration tests maven also tries to pull all images to start from the registry. 

I don't think this is the desired behaviour since you want to integration test the image just build locally and not the one stored in the registry. With 'always' it will pull the image from the registry overwriting the just created image.

Shouldn't autoPull always be part of the image configuration itself instead of a global flag ?

rhuss:
I added this to `integration` which will be then part of 0.13.1. Thanks for the initial PR!

PR #115, opened 2015-03-01
allow containers to be named using its 'alias' 

jgangemi:
- allow container to be named using `alias` when created/started


rhuss:
Do you intend to implement the other namingSchemes, too ? If not I'll try to work on them (or at least open an issue so they are not forgotten ;-) Currenlty, an error occurs if `docker:start` is called twice (which is probably expected).

jgangemi:
for now i think we should just file a separate issue for the other schemes. no one else seems to be clamoring for this and what's been implemented meets my needs. 

an error is expected if `docker:start` is called twice. this is inline w/ how the `docker` binary works.

Issue #36, opened 2014-10-24
More output when  building


rhuss:
While building, only when pulling an image (which takes some time) log output is printed. Instead it would be nice to have all build steps printed.

Probably in some grey color so that it doesnot get into the way (use advance ANSI here).

rhuss:
Added a `dockver.verbose` property, which, when enabled, will print out the build steps.

PR #123, opened 2015-03-11
Added conversion to system local path-separators for the test.

jburell:
Fixes two tests that fail on master-branch.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/access/chunked/BuildResponseHandler.java
+++ b/src/main/java/org/jolokia/docker/maven/access/chunked/BuildResponseHandler.java
@@ -21,7 +21,7 @@ public class BuildResponseHandler implements ChunkedResponseHandler<JSONObject> 
             }
         } else if (json.has("stream")) {
             String message = json.getString("stream");
-            log.debug(message.trim());
+            log.info(message.trim());
         } else if (json.has("status")) {
             String status = json.getString("status").trim();
             String id = json.has("id") ? json.getString("id") : null;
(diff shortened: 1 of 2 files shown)


jgangemi:
i think this has already been corrected on the `integration` branch.

also - i don't necessarily want to see all the build steps, so i think the fix for #36 should be controlled via some kind of property so it can be turned on/off.

jburell:
About the integration branch, you're right. It was me who did the PR. Forgot that it got accepted because I had a fresh clone from master. My bad.

On the logging issue, I can see why you would like to have sparse logging sometimes. However I think the plugin, in general, does too many steps in the background without telling the user (like how it pulls a template Dockerfile if none was provided). Some of these steps are not obvious and should perhaps be communicated. Do you have a preference for the name of the flag? Also, should the flag be used to turn off logging (i.e. a "quiet"-flag) or to turn it on (a "verbose"-flag)?

rhuss:
`integration` and `master` doesn't differ yet (since it just after a release), so indeed the path issue in the unit test still needs to be fixed (sorry don't have a windows box for testing). Will apply that ASAP.

/wr to the build logging I agree that it should be easier visible instead of using `mvn -X` which is super noisy. If a new option is needed, I would opt for a *verbose* flag (for backward compatibility issues). Still looking for a smarter coloring option (low contrast grey maybe).

rhuss:
Thanks, I applied the bug fix for the path issue in the unit test on branch integration. For the logging, I'm still looking for a more fine grained solution. But switching on building log for everything can become quite noisy. I will work on #36 for the next release, ok ?

rhuss:
Should fixed now.

PR #137, opened 2015-04-02
Do port/parameter substitution in environment variables

sverhagen:
I had a need for port/parameter substitution in environment variables. I wanted to set the database URL on my second container, that had the port as determined for the first container.

E.g., the second container had something like this:

    <env>
        <DATABASE_URL>jdbc\:mysql\://${env.DOCKER_HOST_HOST}:${database.port}/</DATABASE_URL>
    </env>

The code in the merge request does this, plus a few minor other snippets.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.3.2</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.json</groupId>
       <artifactId>json</artifactId>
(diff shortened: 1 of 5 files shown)


sverhagen:
Added the *sign off* as required [here](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md). Is there anything else I could do to move this forward? Thank you.

rhuss:
I will have a look at it ASAP (probably at the end of this week). I think its a quite nice feature, so you can be pretty sure that the PR will be applied after a review.

rhuss:
Sorry for the delay. Applied the PR to `integration`, will be available in 0.11.4 soon.

Issue #51, opened 2014-11-21
file permissions not set when 'fileMode' set in assembly descriptor


jgangemi:
i have the following in my assembly descriptor (it's just the relevant snippet)

```
<fileSet>
  <directory>${project.build.scriptSourceDirectory}</directory>
  <outputDirectory>/</outputDirectory>
  <filtered>true</filtered>
  <includes>
    <include>startup.sh</include>
   </includes>
   <fileMode>755</fileMode>
</fileSet>
```

when my container gets built, the startup script doesn't have the correct permissions. i'm not sure if this something that should work in the context of this plugin or not.

i also tried setting the permission directly on the local filesystem to see if they carry over but was unsuccessful there as well. right now i'm working around this by specifying `bash startup.sh` as the command but that somewhat less then idea.

is this a bug? if not, do you have any suggestions on how this could be handled? i want this contained within the data container that gets built rather then mucking w/ the image that will be running it (which in this case is the 'official' tomcat instance).

jgangemi:
this was fixed in 0.11.0

dcalmbach:
Hi, i'm using version 0.11.2 and it seems that this issue is not fixed. i explored the temporary tar from assembly plugin which takes care of filemode, the tar created by maven-docker-plugin does not.

dcalmbach:
Perhaps its a windows issue since i'm building on a windows box

rhuss:
Thanks for reporting, I'll try to have a look for this.

jgangemi:
i'm thinking this is windows related b/c the above works for me on a mac and linux machines.

jgangemi:
also, what is the value of `fileMode`? i just realized what i posted above it slightly incorrect and it should be `<fileMode>0755</fileMode>`

trying to google but not coming up w/ much - it's possible this is a bug in the assembly plugin as well.

dcalmbach:
my assembly.xml was :<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
    <id>docker</id>
    <fileSets>
        <fileSet>
            <directory>${project.basedir}/src/main/docker/tagr</directory>
[...]

PR #139, opened 2015-04-06
Output to dir is now an archive.

efollea:
The temporary assembly is now outputdir/build/maven.tgz instead of a directory (was outputdir/build/maven/

Aim to fix the files/directory rights defined in assembly issue which appear for some windows users.
The problem is that the fat/ntfs or any other file systeme like this ones which does not support linux like rights)
The solution is to use a system that does not rely on underling File system. choosen to use tgz archive.

Fixed some unit test issues under windows.


efollea:
Dockerfile in outputdir should not be used when dockerFileDir is used, and a Dockerfile is present.

efollea:
Dockerfile in outputdir should not be used when dockerFileDir is used, and a Dockerfile is present.

efollea:
this is a proposal for bug file permissions not set when 'fileMode' set in assembly descriptor #51

rhuss:
Thanks, applied it finally (sorry for the long delay). Using a tar as intermediate format is also nice so that arbitray files in the output dir does not end up in the image by accident. Will be in 0.11.4

PR #145, opened 2015-04-14
Add parameter to allow skipping of tagging

twz123:
Naive approach to be able to skip tagging of images during `docker:build` using a property. When `docker.skipTags` is `true`, then built images won't be tagged.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,5 +1,8 @@
 # ChangeLog
 
+* **0.12**
+  - Add a new parameter 'docker.skipTags' that suppresses tagging of images that have been built.
+
 * **0.11.5**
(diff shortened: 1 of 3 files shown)


rhuss:
Looks good to me. Could you add some documentation to manual.md and add to changelog.md ?

Just out of curiosity: What is your use case for skipping configured tags ?

twz123:
This used to be a CI issue when creating docker images tagged with the git commit ID. The current solution is a special profile in the POM for that. Not a big deal, but with the skipTags property there wouldn't have been the need to change the POM.

twz123:
Added some lines to the docs. I wasn't sure if I did it right for the change log, though.

rhuss:
Applied for 0.12.0

PR #156, opened 2015-05-07
Add CMD and ENTRYPOINT (fixes #130)

### LINKED ISSUE #130: Support shell syntax
Right now, the CMD is specified in exec form no matter what.
This means, that you can't really use variables in you command (adding `sh -c`to the beginning doesn't always work).

It would be nice if you could choose between shell and exec form.

osigida:
introduced cmd and entrypoint with exec and shell form support.
the change is not backward compatible to the previous version.

Signed-off-by: Oleg Sigida <oleg.sigida@gmail.com>

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -228,8 +228,10 @@ of an image configuration. The available subelements are
 
 * **assembly** specifies the assembly configuration as described in
   [Build Assembly](#build-assembly)
-* **command** is the command to execute by default (i.e. if no command
-  is provided when a container for this image is started).
+* **cmd** A command to execute by default.
+  [Start-up Arguments]
+* **entrypoint** An ENTRYPOINT allows you to configure a container that will run as an executable.  
+  [Start-up Arguments]
 * **env** hold environments as described in
   [Setting Environment Variables](#setting-environment-variables). 
 * **from** specifies the base image which should be used for this
@@ -263,7 +265,16 @@ Here's an example:
   <volumes>
     <volume>/path/to/expose</volume>
   </volumes>
-  <command>java /opt/demo/server.jar</command>
+  
+  <entryPoint>
+    <!-- exec form for ENTRYPOINT -->
+    <params>
+      <param>java</param>
+      <param>-jar</param>
+      <param>/opt/demo/server.jar</param>
+    </params>
+  </entryPoint>
+
   <assembly>
     <mode>dir</mode>
     <basedir>/opt/demo</basedir>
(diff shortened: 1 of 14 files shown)


alan-czajkowski:
please fix the documentation: manual.md mentions entrypoint but still shows ```<cmd>``` tag

doc/manual.md
```diff
@@ -262,7 +265,30 @@ Here's an example:
...
+    <!-- or exec form for command -->
+    <params>
+      <param>java</param>
+      <param>-jar</param>
+      <param>/opt/demo/server.jar</param>
+    </params>
+  </cmd>
+
+  <!-- or you can use entry point -->
+  <cmd>
```
alan-czajkowski: please fix the documentation, mentions entrypoint but still shows ```<cmd>``` tag

src/main/java/org/jolokia/docker/maven/assembly/DockerFileBuilder.java
```diff
@@ -14,6 +16,8 @@
  */
 public class DockerFileBuilder {
 
+    private static final Joiner JoinOnComma = Joiner.on("\",\"");
```
alan-czajkowski: should use lowercase J? Java coding conventions?

alan-czajkowski:
can CMD and ENTRYPOINT configuration in the plugin be used in conjunction? the Docker documentation mentions that if ENTRYPOINT is used then CMD becomes the parameters to ENTRYPOINT, just want to make sure the plugin will handle both simultaneously

osigida:
@alan-czajkowski yes, if both cmd and entrypoint are specified they will be printed out to docker file, so one an use this feature (gonna test it soon, sounds like a good simplification to my use cases)

doc/manual.md
osigida: oops, thanks a lot! fixed

src/main/java/org/jolokia/docker/maven/assembly/DockerFileBuilder.java
osigida: actually it should be JOIN_ON_COMMA as a static constant, renamed.

rhuss:
Thanks a lot for the PR, it adds really a very useful feature and is well done. 
However, since it will change the configuration syntax in such a way that a migration is necessary, I would like to schedule it for 0.12.0 (second version number is reserved for such things). I need to do a 0.11.4 before for some important fixes, though. But 0.12.0 is the next one ;-)

I added some commit comments, i.e. it would be perfect if you could adapt the PR a bit:

* Remove the dependency to guava, it's not (yet) really needed and only adds to the baggage.
* Adapt the example in the documentation so that it is directly usable ;-) 
* Squash commits to a single one, please. This makes commenting on commits much easier. Tipp: Check for 'git rebase -i' and search for "git squash", e.g. [this](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) one)

As I said, I really like the code, its great, these are only minor things. Thanks a lot !

rhuss:
Ah, I forgot. Could you please also rebase againt branch `integration`. This is where I integrate all PRs. Danger of conflict is quite low, though, the current PRs are quite orthogonal.

osigida:
@rhuss, sure, will update the PR soon.
Regarding guava, I found out that guava 10 is already in the dependency list, then I decided to keep it 
So dependency wise it is the same, but it is visible now.

So what will be the conclusion, keep or drop?

rhuss:
Ah, ok, its a transitive dep of maven-plugin-api. No problem to keep it then ;-)

osigida:
@rhuss, please have a look
* as we agreed, guava is still used
* restructured documentation a bit, is it better?
* commits are squashed
* rebased on recent integration

rhuss:
thanks, will review and add this for 0.12.0 (there will a quick bugfix release 0.11.5 before btw).

osigida:
...rebased on integration branch...

rhuss:
Thanks for the rebase. I will review it ASAP, but probably not this week anymore, hopefully next week.

rhuss:
Merged in and tuned it a bit. Sorry for the long delay.

BTW, I renamed `<params><param>` to `<exec><arg>` since it better matches the Dockerfile syntax.

Thanks again !

alan-czajkowski:
when will a release be made for 0.12.0?

rhuss:
Next week, promised. Just have to prepare a talk for Thursday, sorry, i'm in 'slide machine' mode at the moment.

Issue #157, opened 2015-05-07
add ability to inline assembly section 


osigida:
Currently I have to specify external assembly file, even if I want to add couple of files.
It would be great to inline it in to assembly tag

PR #158, opened 2015-05-07
Adding ability to inline assembly into pom file (#157)

osigida:
Change is not backward compatible: descriptor is renamed into descriptorFile

Signed-off-by: Oleg Sigida <oleg.sigida@gmail.com>

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -275,6 +275,8 @@ Here's an example:
 * **basedir** depicts the directory under which the files and
   artifacts contained in the assembly will be copied within the
   container. The default value for this is `/maven`.
+* **assemblyDef** inlined assembly descriptor as
+  described in the section [Docker Assembly](#docker-assembly) below. 
 * **descriptor** is a reference to an assembly descriptor as
   described in the section [Docker Assembly](#docker-assembly) below. 
 * **descriptorRef** is an alias to a predefined assembly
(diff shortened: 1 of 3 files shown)


jgangemi:
i'm going to say that i'm not a fan of the `descriptor` element changing here. `descriptor` and `descriptorRef` were specifically chosen b/c they match what is used by the assembly plugin itself.

i would rather see this functionality added using a different element.

osigida:
@jgangemi any hint what would be a good name for it? 
I cannot make up any better naming on myself...

jgangemi:
how about `assemblyDef`?

osigida:
@jgangemi, done, please review

rhuss:
I wonder, whether it works correctly with XML namespaces (the assembly XSD is different from Maven's pom, but this shouldn't matter). I will try this soon. This change is in fact backwards compatible since it only adds a new tag (and doesn't modify or remove one), so we could put it into 0.11.4 as well (which I want to release this week because of some quite important fix)

osigida:
@rhuss, not sure for complicated assembly, but this works for me quite well, testing it for a week already  
```
<assembly>
    <basedir>/</basedir>
    <descriptor>
        <files>
            <file>
                <source>${project.build.directory}/${project.build.finalName}-with-dependencies.jar</source>
                <outputDirectory>/</outputDirectory>
            </file>
            <file>
                <source>${project.basedir}/src/main/config/active.json</source>
                <outputDirectory>/</outputDirectory>
            </file>
        </files>
    </descriptor>
</assembly>
```

rhuss:
Sounds good, I will try to experiment a bit with namespaces tomorrow a bit (which is somewhat cool since a decent IDE can then autocomplete on the tags properly). Will add this as a profile to the sampe app.

But if it works with Maven, we are save even without namespaces. Unfortunately I still have not real integration tests checking this automatically.

rhuss:
Added it to branch `integration`. Renamed the tag to `inline` though since the context "assembly" is clear from the parent tag. Also added an example in `data-jolokia-demo` sample under the profile "inline".

PR #159, opened 2015-05-07
Correct the documentation on what range of ports is used for random selection.

breser:


rhuss:
It in release 0.11.4 and now online.

PR #161, opened 2015-05-07
Allow dynamic ports to be override by System properties.

breser:
Previous behavior only allowed project properties to override them.
Even the documentation suggested that you could override them by passing
-Dfoo=bar on the mvn cli, but this wasn't actually the case.

Signed-off-by: Ben Reser <ben.reser@cdk.com>


rhuss:
Thanks, you are right. The behaviour is not as described in the manual. Your fix is straight forward, I will polish it a bit and it will make it into 0.11.4.

PR #163, opened 2015-05-12
Fix log messages getting cut off in the build

dankeast:
Hi,

I was getting log messages being cut off in the build:

[ERROR] DOCKER> Invalid log format for '201' (expected: "<timestamp> <txt>")

I switched this method:

https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[],%20int,%20int)

"An attempt is made to read as many as len bytes, but a smaller number may be read. The number of bytes actually read is returned as an integer."

for this method:

https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/IOUtils.html#read(java.io.InputStream, byte[], int, int)

"This implementation guarantees that it will read as many bytes as possible before giving up"

Cheers,
Dan


rhuss:
Thanks a lot ! 'will be in the next release.

dankeast:
Fantastic! It's a really great project, we're using pretty much all of it's
features in my project now.

rhuss:
Thanks ;-) And thanks for the PR ...

PR #168, opened 2015-05-15
Chunked response errors should fail plugin execution

timkral:
This is a pull request to fix this issue: https://github.com/rhuss/docker-maven-plugin/issues/167

I saw that there was already some processing around chunked responses to parse them and log their messages. This just takes it one small step further and throws an exception if the chucked response contains an error (effectively causing the plugin execution to fail).

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/access/chunked/ChunkedResponseHandler.java
+++ b/src/main/java/org/jolokia/docker/maven/access/chunked/ChunkedResponseHandler.java
@@ -1,5 +1,7 @@
 package org.jolokia.docker.maven.access.chunked;
 
+import org.jolokia.docker.maven.access.DockerAccessException;
+
 public interface ChunkedResponseHandler<T> {
-    void process(T toProcess);
+    void process(T toProcess) throws DockerAccessException;
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/jolokia/docker/maven/access/chunked/ChunkedResponseHandler.java
```diff
@@ -1,5 +1,7 @@
 package org.jolokia.docker.maven.access.chunked;
 
+import org.jolokia.docker.maven.access.DockerAccessException;
+
 public interface ChunkedResponseHandler<T> {
-    void process(T toProcess);
+    void process(T toProcess) throws DockerAccessException;
```
timkral: Updating the interface contract to allow for exception throwing. `DockerAccessException` may not be the best choice based on its description, but everything seemed to funnel into it anyway, so I went with that instead of creating a whole new class of exception.

rhuss:
Looks good to me. Yes, currently there's only `DockerAccessException`, which I think is ok as an indicator that _something_ during the communication with the docker host failed. I'm not a big fan of large exception hierarchies because tend then to always choose the wrong one anyway ;-). 

The fix is applied for 0.11.4 which will be released quite soon.

PR #170, opened 2015-05-16
force a tag so that docker:push doesn't barf if not using a registry …

jstrachan:
…in the docker container name


rhuss:
thx, closing it for now :)

Issue #148, opened 2015-04-24
UNIX socket support


rassie:
I don't know if this is intended (documentation implies that it might be), but I'm sorely missing UNIX socket support for docker. One of the reasons is that we are running Jenkins instances in containers and mount `/var/run/docker.sock` into each container, so that Jenkins could build other docker images. Working with TCP becomes vastly more complicated in this situation.

vauvenal5:
+1

jgangemi:
i don't know what the likelihood of this is going to be given it's not portable across build environments, etc, and i know that is one of the goals of this plugin.

have you tried using the ambassador pattern as a work around for this: https://docs.docker.com/articles/ambassador_pattern_linking/

rhuss:
In fact, Unix socker support is not so easy when coming from Java. As @jgangemi said, portability is one of the main key features. However, I also think that it shouldn't harm to allow unix socket access for Linux plattforms as an optional way for accessing the Docker daemon as long it doesn't require some strange packaging of the plugin because of native libs to be included.

Maybe we can collect here what are the technical possibilities to do so ? (I know, there are some libs for this out there, but never checked them really)

jgangemi:
marking as closed

PR #179, opened 2015-05-25
Add Support for communication to docker hosts using UNIX domain sockets

twz123:
This is my draft to a solution for #148. The same technique is used by the Spotify plugin. No obscure packaging involved.

  * Use jnr-unixsocket for native OS-level access
  * Use a pseudo socket class that mimics a Java network socket and delegates to the jnr channel where applicable
  * Provide a special ApacheUnixSocketHttpDelegate that enables the Apache HTTP client to use that pseudo socket for HTTP communication

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,12 @@
     </dependency>
 
     <dependency>
+        <groupId>com.github.jnr</groupId>
+        <artifactId>jnr-unixsocket</artifactId>
+        <version>0.8</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
       <version>3.3.2</version>
--- a/src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
@@ -1,8 +1,8 @@
 package org.jolokia.docker.maven;
 
 import java.io.IOException;
-import java.net.MalformedURLException;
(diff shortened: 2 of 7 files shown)


rhuss:
Awesome! Thanks a lot. Do you know any limitations of jnr-unixsocket ? (i.e. does it run on **any** Unix or are there are limitations ? Just to add it to the docs in case).

BTW, whould you mind to signoff the PR ? (as explained [here](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md)). Shouldn't be much more than a `git commit -s --amend` on this branch. Thanks !

I will review and integrate it ASAP, probably this week.

twz123:
No, I don't have any deeper knowledge of jnr-unixsocket. The only two things I can state about it:

# Spotify uses it in their docker plugin
# It simply just worked on first shot when implementing this PR

Added the signoff.

rhuss:
Thanks ! Give me some days for the review, won't make it this week but hopefully the week after.

Thanks again ...

rhuss:
Merged it into the integration with some slight modifications:

* Detangled inheritance hierarchy
* Avoid encoding the socket path as an URL host name. Instead set the path directly in the constructor of `UnixSocketFactory`. The base URL (host & port) is now meaningless and set to a dummy value.
* If no DOCKER_HOST is set or configured, and a file `/var/run/docker.sock` exists, takes this as a default.

Again, thanks a lot ! Especially for finding out how Apache HC works nicely with a Unix Socket. 

The next release 0.12.0 containing the changes will come next week.

PR #180, opened 2015-05-26
fixed an issue (ArrayOutOfBoundException) with long log lines

fbertola:
Some application tend to dump huge log string (eg. a list of all system variables) for debug purpouses. These log exceeds the predefined buffer length of 8129. I think that the buffer should be created dynamically per line.


rhuss:
Looks good to me, thanks !

PR #188, opened 2015-06-05
Convert host machine path to virtual machine path for those who use boot2docker

0x7364:
Hacky workaround like this: https://github.com/boot2docker/boot2docker/blob/1bf1671c0f5c103ac4e2229b96a21a04b6d99043/rootfs/rootfs/etc/rc.d/automount-shares#L34
Pros: now you can use the variable ${project.build.directory} in run.volumes.bind.volume

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/access/ContainerHostConfig.java
+++ b/src/main/java/org/jolokia/docker/maven/access/ContainerHostConfig.java
@@ -17,6 +17,7 @@ public class ContainerHostConfig {
             JSONArray binds = new JSONArray();
 
             for (String volume : bind) {
+                volume = volume.replace("\\", "/").replaceAll("^C:", "/c");
                 if (volume.contains(":")) {
                     binds.put(volume);
                 }


rhuss:
Thanks for your PR.

Please help me a bit to understand this PR: `${project.build.directory}` contains backslashes on Windows,  so this mangling is needed on windows only, right ?

What exactly is your use case ? To mount a directory within you source in the container when running it (and using boot2docker) ?

Could you give me an example how a build directory on Windows really looks like (sorry, don't have one available) ? I know that boot2docker can map directories within a user home directory to the the VM, but I think `${project.build.directory}` should have a defined format. Your regexps looks a bit like a shot gun approach. 

To nail this down to a single expression would be cool (should be really possible). It really would help if you can provide some examples what kind of paths should be mapped to what. 

Thanks !

0x7364:
1. yes
2. yes
3. ${project.build.directory} on Windows looks like this one: "C:\Users\0x7364\Projects\test\target" and mount point looks like this one: "/c/Users/0x7364/Projects/test/target". I was trying to write a regular expression like boot2docker automount-shares in order to get a clear understanding of why this is necessary. Of course it can be simplified.

rhuss:
I added the patch (manually) for release 0.12.1 (due this friday)

Thanks !

PR #199, opened 2015-06-22
resolves issue 196

### LINKED ISSUE #196: allow 'extraHosts' to accept hostname as destination ip
this is similar to #101.

i want to specify a hostname as the destination address in the maven configuration that will be resolved to an ip address at runtime, eg:
```xml
<run>
  <extraHosts>
    <host>dev-db.hostname.pvt:docker-dev.hostname.pvt</host>
  </extraHosts>
</run>
```
this lets me inject the correct ip address by way of `/etc/hosts` instead of using maven properties. 

if the specified host cannot be resolved at start time, the plugin will throw an error.

jgangemi:
- allow 'extraHosts' ip addresses to be resolved at runtime

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,5 +1,8 @@
 # ChangeLog
 
+* **0.12.1**
+  - allow `extraHosts` ips to be resolved at runtime
+
 * **0.12.0**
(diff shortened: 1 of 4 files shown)


rhuss:
Applied to integration and pushed. Thanks !

PR #200, opened 2015-06-23
wait command fails due to empty data being returned from docker logs

tszpinda:
After upgrading to Docker 1.7 we started to have problems with wait command which was timing out as it was giving up reading from logs due to first two reads in header not containing expected data (the declaredLength being 0). PR fixes this issue.


rhuss:
Thanks ! Applied path to integration

PR #204, opened 2015-06-24
docker:build - add support for WORKDIR

ttiurani:
This is a pretty straightforward feature addition of the WORKDIR Dockerfile parameter:
https://docs.docker.com/reference/builder/#workdir
Would it be possible to pull?


rhuss:
Adapted and applied to integration.

PR #212, opened 2015-07-06
fix #211 - wait url can contain user information

### LINKED ISSUE #211: Start wait url does not support user information in URL
Currently _wait_ option times out always when you try to refer to protected URL.
Specifying  user auth info in URL does not make any difference.

I added some logging to HttpPingChecker and reason is http error 401:
```
...
[DEBUG] DOCKER> wait url 'http://user:pass@host/someurl' exception: Read timed out
[DEBUG] DOCKER> wait url 'http://user:pass@host/someurl' response code: 401
...
```

*WaitUtil.HttpPingChecker* uses *java.net.HttpURLConnection* and this is probable reason.

I think there are two options to allow user auth info:
* use apache HttpClient lib which support user info by default (library is already in plugin depedencies).
* keep using *java.net.HttpURLConnection* but:
  * extend configuration with authorization user / password (required configuration format change), or
  * extract auth info from specified url before using it.

I tested HttpClient option and can supply pull request for that.

tdomzal:
Replace ping implementation from java.net.HttpURLConnection
to Apache HttpClient. Generate ping errors diagnostics info
when maven debug level is active.
Fixes #211.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/StartMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/StartMojo.java
@@ -114,7 +114,7 @@ public class StartMojo extends AbstractDockerMojo {
             ArrayList<String> logOut = new ArrayList<>();
             if (wait.getUrl() != null) {
                 String waitUrl = StrSubstitutor.replace(wait.getUrl(), projectProperties);
-                checkers.add(new WaitUtil.HttpPingChecker(waitUrl));
+                checkers.add(new WaitUtil.HttpPingChecker(waitUrl, log));
                 logOut.add("on url " + waitUrl);
             }
             if (wait.getLog() != null) {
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks a lot ! Could you please sign-off your PR as described [here](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md) ? 

I will integrate it for 0.13.2 which will be released probably end of next week.

tdomzal:
Sure. 
Take a look at updated commit message if it is OK.

rhuss:
perfect, thanks !

rhuss:
Merged into branch integration. 'hope you don't mind that I removed the debugging again (in order to keep the API a bit cleaner). 

If there's a good reason we can add it again.

Thanks !

tdomzal:
I thought logging may be handy when visible with maven -X flag but of course I don't mind if you just clean it out.
Thanks!

PR #216, opened 2015-07-09
Ignore zero-length log frames

stevenschlansker:
Sometimes the log watch gives you a zero length frame.  I have no idea why,
but ignoring it allows it to work fine, and is better than bailing...


stevenschlansker:
Whoops, already fixed as #200

PR #217, opened 2015-07-09
If the StartMojo fails, stop all started containers

stevenschlansker:
We're about to fail the build anyway, so let's not leave heavyweight resources lying around.  ( HubSpot/Singularity#607 )

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/StartMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/StartMojo.java
@@ -60,6 +60,8 @@ public class StartMojo extends AbstractDockerMojo {
         getPluginContext().put(CONTEXT_KEY_START_CALLED, true);
 
         LogDispatcher dispatcher = getLogDispatcher(dockerAccess);
+
+        boolean success = false;
         try {
             for (StartOrderResolver.Resolvable resolvable : runService.getImagesConfigsInOrder(getImages())) {
                 final ImageConfiguration imageConfig = (ImageConfiguration) resolvable;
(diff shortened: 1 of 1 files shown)


tpetr:
:+1:

stevenschlansker:
@jgangemi Sure why not :smile:

stevenschlansker:
I'm sad that this didn't make 0.13.1, maybe we could get a 0.13.2?

rhuss:
Sorry, I had to get out 0.13.1 very quickly because of a nasty behaviour of the special version of Docker Fedora/Centos uses and 'docker:push'. 0.13.1 adds a workaround.

I highly appreciate you contribution and 0.13.2 will follow next week. Please don't be disappointed, typically it takes a bit (one to four weeks) until a PR is integrated. 

Thanks a lot ...

rhuss:
@stevenschlansker, in the meantime could you do me a favour an sign-off the PR ? How this is done is explained [here](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md)

And if you want to go fancy, you could also rebase it to be a single commit (git rebase -i). But that's not so important.

stevenschlansker:
No worries.  I added the sign-off.  I think squishing the commits does not make sense, as they are truly unrelated fixups, so they should be two separate commits.

Looking forward to 0.13.2 :+1:

rhuss:
Intgerated it into branch 'integration', will be part of 0.13.2 end of this week.

Thanks again !

Issue #73, opened 2014-12-29
volume from doesn't appear to work


taylodl:
I start a maven container and give it the name 'maven' using docker's --name option. In my pom.xml I've configured my image as follows:

```xml
<image>
   <name>taylodl/wls-mydomain</name>
  <run>
   <volumes>
      <from>maven</from>
   </volumes>
         .
         .
         .  
```
When I run maven I get the following error: 

```
[ERROR] Failed to execute goal org.jolokia:docker-maven-plugin:0.10.6-SNAPSHOT:start (start) on project jaxRS-maven: No container start order could be found: Cannot resolve image dependencies for start order
[ERROR] Unresolved images:
[ERROR] taylodl/wls-mydomain depends on maven
```

If I run taylodl/wls-mydomain using the docker run option "--volumes-from maven" then everything runs as I expect and the volumes are mounted where I expect. I was expecting <volumes><from> to accomplish the same thing.

jgangemi:
that functionality doesn't currently exist but i think that could be easily added. an additional element (maybe `external`) could be added to the existing `volumes` element that could be used to reference volumes created outside the container.

jgangemi:
fixed in 0.13.2

Issue #207, opened 2015-06-27
allow containers to be skipped during startup


jgangemi:
during the `verify` phase, i start a postgres container to run all my integration tests against but i don't want the application container to start at the same time b/c it's not involved w/ any of the tests, etc. 

i thought about doing this via profiles (i already use one so the tests don't run as part of the 'normal' build process) but i don't believe that is going to work w/o having to duplicate parts of the container configuration across various profile. instead, i'd like to see an additional element added to the run configuration called `startup` (open to other suggestions) that defaults to `true` but if it's false, that container will get skipped.

then i can define all my container configurations as part of the 'main' build definition and just define the property to disable startup in the appropriate profile(s).

rhuss:
I think this is already possible by using the top level `image` configuration parameter (or property `docker.image`) which can contain a comma separated list of images to build/start. So you can easily add a profile and set this filter-like configuration value on your own.

What I think could be quite helpful is to support Docker labels. So we can attach labels to images (when building) and then provide a similar top-level config `docker.label` for selecting the images to process.

But first: Does `docker.image` solves your use case ?

rhuss:
I think, this could work (not tried): 

```
<profile>
  <id>db-only</id>
  <plugin ....>
     ...
     <configuration>
         <image>postgres</image>
     </configuration>
 </plugin>
</profile>
```

jgangemi:
lol - you're right and it clearly states that in the docs (although it can only be done by using the property to list the images). guess i should have rtfm a little more carefully when i was looking to see if there were any options to do this last night.

```xml
    <profile>
      <id>integration-tests</id>
      <properties>
        <docker.image>postgres:9.3</docker.image>
      </properties>
      <build>
        <plugins>
          <plugin>
            <groupId>org.jolokia</groupId>
            <artifactId>docker-maven-plugin</artifactId>
            <executions>
              <execution>
                <id>start</id>
                <phase>pre-integration-test</phase>
                <goals>
                  <goal>start</goal>
                </goals>
              </execution>
              <execution>
                <id>stop</id>
                <phase>post-integration-test</phase>
                <goals>
                  <goal>stop</goal>
                </goals>
              </execution>
            </executions>
          </plugin>
          <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-failsafe-plugin</artifactId>
          </plugin>
        </plugins>
      </build>
    </profile>
```
[...]

PR #219, opened 2015-07-12
allow linking / volumesFrom external containers

### LINKED ISSUE #195: allow linking to containers not managed by the plugin
i thought i had seen an issue regarding this before, but perhaps it was for referencing volumes from other containers.

in any case, when starting a container via the plugin, i'd like to be able to link to an existing container that has been started via some other mechanism.

in an effort not to break existing configuration, my thought is to just add a new xml block to the run section:
```xml
<linkExternal>
  <link>container_name:env_var_prefix</link>
</linkExternal>
```
thoughts? i'm open to different element names, etc.

jgangemi:
resolves #195 and #73 

- pulled fine grained query methods out of DockerAccess into QueryService
- added coarse grained list/inspect methods to DockerAccess
- introduced model objects
- refactored RunService to accept DocerAccess instance via 'initialize' method
- allow linking/mounting from external containers

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -4,6 +4,8 @@
   - Allow autoPull to be forced on docker:build and docker:start (#96)
   - Respect username when looking up credentials for a Docker registry (#174)
   - Add "force=1" to push for Fedora/CentOs images allowing to push to docker hub
+  - Allow linking to external containers (#195)
+  - Allow volume mounting from external containers (#73)
   
 Note that the default registry has been changed to `docker.io` as docker hub doesn't use 
 `registry.hub.docker.com` as the default registry and refused to authenticate against this 
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -702,6 +702,9 @@ DB_PORT_5432_TCP_PORT=5432
 DB_PORT_5432_TCP_ADDR=172.17.0.5
 ```
 
+If you wish to link to existing containers not managed by the plugin, you may do so by specifying the container name
+obtained via `docker ps` in the configuration.
+
 ##### Volume binding
 
 A container can bind (or "mount") volumes from various source when starting up: Either from a directory of
@@ -743,6 +746,9 @@ should even work for boot2docker:
   </bind>
 </volumes>
 ````
+
+If you wish to mount volumes from an existing container not managed by the plugin, you may do by specifying the container name
+obtained via `docker ps` in the configuration.
 
 ##### Container restart policy
 
--- a/src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
@@ -3,9 +3,17 @@ package org.jolokia.docker.maven;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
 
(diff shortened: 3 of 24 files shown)


jgangemi:
ideally the `UrlBuilder` would be updated to return `DockerUrl`s for everything, but baby steps :)

rhuss:
Thanks a lot, will have a review soon.

jgangemi:
cool - i had a better idea on how to deal w/ the state/services this morning.

rhuss:
Thanks, Jae ! 

Quite some changes, need a bit to walk through it ;-)

jgangemi:
haha - don't you always when i submit something? :)

rhuss:
Yeah, always gate keeping ;-)

rhuss:
I worked mostly through it and cleaned up a bit (you know what that means ;-P)

I like the refactoring in the QueryService and the ServiceFactory (which I now call ServiceHub since all services are created upfront). Thanks !

Also, you know that I didn't want to introduce a fancy 'model' in order to avoid bloat. However using a model more or less as a 'proxy' to the JSON object (without copying things over) looks fine. Also, it hides a bit the inconsistence model that Docker uses in their Remote API. I introduced a `Container` interface which can be nicely implemented by both the elements returned by 'container list' and 'container inspection'. What I also did, though, I removed everything currently not needed. We can introduce this always later on.

I didn't test the functionality yet, but I think there were some gotchas (see my commit comment). Also, you looked up the name in the link config as a container name.

I pushes my changes to `195-external-linking` so it might be good idea to check this out.

rhuss:
Hmm, just saw that 'Names' is really a part of an list element when listing containers. But it is nowhere documented and also, why it is a list ? AFAIK a container can have only one name, or ?

Strange, strange ...

rhuss:
I asked here https://github.com/docker/docker/issues/14675

jgangemi:
the `Names` field is undocumented at the moment but i posted an issue in the docker project and was told that when a container gets linked, it gets an additional name, eg: `['/database', '/database/application-linked-to-it'` (i may have the linkee-linker backwards, but you get the idea.

i also switched everything over to using container name instead of container id b/c that's how the cli works and if you inspect a container created that way, it shows the name and not the container id. everything worked for me when i pushed, so i'll pull down the new branch and try it out.

did you strip out all of the fields from the original `Container` object? i added all of them to pave a little way to adding a `docker:ps` mojo as those are fields that would get displayed.

rhuss:
@jgangemi ah, thanks for the explanation. I understand now. I just pushed to the branch (with some rebasing, so please pull it a fresh). For me it works, too (after i inserted back your stuff ;-)

Yes, I stripped off on the container stuff since I didn't see any use up to now (I do this always when I see unused code ;-). Sorry. I like that you don't copy over data but convert it 'on-the-fly' making the model more  a proxy than a data object.

Also, since the container elements were quite similar for 'list' and 'inspect' I tried to unify them behind an interface. This might cease to work when you add more from the 'inspect' side. One could then either decided to throw UnsuporrtedOperationExceptions or, maybe better, extend a `ContainerDetails` on top of `Container`. What do you think ?

jgangemi:
lol - i was originally going to call `DependencyTracker` `ContainerTracker`

jgangemi:
everything looks good to me.

i initially thought about trying to make `Container` and `ContainerDetails` share some kind of commonality (either via an interface or one inheriting from the other) but after peeking at some other clients that exist and not seeing that done, i went the two different classes route.

glad you like the model as a proxy. i know you wanted to shy away from that but it seemed cleaner.

once you get this merged i'm gonna deal w/ #207.

rhuss:
I will cut a release 0.13.2 tomorrow containing this fix and merge everything to master for that

thanks ...

Issue #90, opened 2015-02-06
allow a 'global' port properties file configuraiton option


jgangemi:
right now the `portPropertyFile` needs to be specified on a per `run` basis for images which means i need to load/read multiple files instead of just one.

as an enhancement, we should support defining a global `portPropertyFile` element for all images.

the behavior will be similar to that of the registry configuration, if the image specifies the element it will take precedence over what is globally defined.

rhuss:
If both are given, should the image specific property file override to the global one or should both be merged (and only a specific container port mapping overrides a global mapping) ?

Maybe we can define a 'default' `<run>` configuration on the global level so that also for other properties defaults can be specified ?

jgangemi:
i based this on how the registry is handled where the more specific value takes precedence, so i think from a consistency standpoint we should stick w/ that.

i'm also not sure i understand the other questions - i'm only proposing the global `portPropertyFile` element here that will override the properties file that the dynamic port values get written to.

are you suggesting we offer a way to default values for various `run` options, such as defining a global 'wait' that will apply to all containers that could then be overwritten? 

if yes, i like that idea a lot b/c i think `env` vars would be a perfect fit. if no, please explain. :)

rhuss:
Ok, understand. Its about *writing out* the port properties only (so, of course merigng is not needed here). Sorry, misunderstood my own plugin ;-). Think thats a good idea. I have still other things with higher prio, so if you like it, a PR is welcome (as always ;-)

For the global `<run>` stuff it's probably a bit over the top (one has to explain it and debugging becomes harded), but a global `<env>`  might be nice (but then again, for `build`, `run` or both ?)

PR #221, opened 2015-07-18
resolves issue 90

jgangemi:
- support for a global 'portPropertyFile' value

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,6 +2,7 @@
 
 * **0.13.3**
   - Allow dangling images to be cleaned up after build (#20)
+  - Introduced global `portPropertyFile` setting (#90)
 
 * **0.13.2**
   - "run" directives can be added to the Dockerfile (#191)
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -136,6 +136,10 @@ parentheses.
   container logs. This configuration can be overwritten by individual
   run configurations and described below. The format is described in
   the [section](#log-configuration) below. 
+* **portPropertyFile** if given, specifies a global file into which the
+  mapped properties should be written to. The format of this file and
+  its purpose are also described [below](#port-mapping). Please note, this field takes precidence
+  over any `portPropertyFile` value specified in an `image` configuration.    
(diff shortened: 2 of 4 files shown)


jgangemi:
i really don't like submitting this w/o a test but i'm still not sure what the proper why to test this type of thing.

i did validate by hand using this pom: https://github.com/jgangemi/docker-maven-test/tree/master/issue-90

jgangemi:
oh - i also fixed the default api version. not sure how that got missed from the other PR but that needs to match what's defined in the constant.

rhuss:
Some questions:

* Why does the global property file overwrites a local one given in the run configuration ? Couldn't it be implemented so that, if both are given, both property files are written ? Also, in #90 it is described inversely, a `<run>` propertyFile overwrites the globale one ...
* Why is the `portPropertyFile` attached to `AbstractDockerMojo` ? It makes only sense for the `StartMojo`, or ?
* Isn't it that an existing property file is overwritten ? I.e. when you have multiple images with port mappings, I would suspect that you would like to have all mappings in this property file, not only the last oner.

jgangemi:
`portPropertyFile` is part of the global configuration, so it has to be on `AbstractDockerMojo` otherwise the other goals will complain they can't find the element.

looks like i flipped the behavior of who overrides who, i can fix that. if we want to write to both locations in the event both are specified, that can be done.

ultimately the use case i'm trying to solve here is what you described, i have multiple images and i want all the port mappings to end up in a single file instead of multiple files. 

let me know how you want to proceed and i'll make the appropriate changes.

rhuss:
Thanks for the quick response. Actually I think it would be best, when we would keep both files if two are given.

Also please check if the the port properties are really collected, it seems to me that the global property file is overwritten for each image's mapping. The first image of course should write out a fresh property file (to overwrite the one from the last run).

Coudl you do this ? Thx a lot ....

jgangemi:
will do

rhuss:
thx!

jgangemi:
closed in favor of  #240

PR #225, opened 2015-07-18
resolves issue 207

jgangemi:
- allow 'build' and/or 'run' configuration to be skipped during build
- fixed model classes to only return first 12 of container id

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ target/
 .classpath
 .project
 .settings
+.*.md.html
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -170,6 +170,8 @@ The `<image>` element can contain the following sub elements:
   identifying the image within this configuration. This is used when
   linking images together or for specifying it with the global
   **image** configuration.
+* **mode** allows the `build` and/or `run` bevhavior for an image to be skipped. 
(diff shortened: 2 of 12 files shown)


rhuss:
I wonder whether it wouldn't be better to have a `<skip>` configuration on both `<run>` and `<build>`. This would be closer to the corresponding feature and feels more natural IMO (also it provides more decoupling between build and run). The global config section is more for generic aspects which is valid in general for the image (like the name).

What do you think ?

jgangemi:
well - there is already the global `skip` which skips execution of the entire plugin, so i guess this falls inline there...

i'm indifferent - if you think it should be on the `build` and `run` configurations, i can move it. i based this on the discussion in #207 which had this being in the 'global' image section.

rhuss:
Yeah, I'm just brainstorming. I know, I suggested a mode but somehow <skip> feels more natural. It also wouldn't interfere with the global `skip` and we would have nice consistent configuration syntax for skipping stuff on different levels.

So I opt for having a skip on <run> and <build> if you don't mind.

jgangemi:
ok - i'll take care of it.

rhuss:
Thanks !

jgangemi:
closing in favor of a new request w/ squashed commits

PR #227, opened 2015-07-21
Change docker http client to use handlers instead of results to avoid leaving open connections

rabelenda:
The http client was not correctly used which was leaving connections open not being able to reuse them, which caused at some point to reach the max connections per route.

We faced this particular issue when trying to build 2 images with 2 tags each: since the default limit of connections per route host is 2 (when using unix socket, not https since https has a hardcoded config of 10 connections per route host) when we tried to build the images the plugin was getting stuck trying to get more connections.

The change consists mainly in changing existing httpClient execute using result as return with proper ResponseHandlers so http client library automatically closes httpResponse s to release open connections.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>org.jolokia</groupId>
   <artifactId>docker-maven-plugin</artifactId>
-  <version>0.13.2</version>
+  <version>0.13.3-SNAPSHOT</version>
   <packaging>maven-plugin</packaging>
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks a lot ! W

ould you mind to sign-off your PR as described [here](https://github.com/rhuss/docker-maven-plugin/blob/master/doc/contributing.md)

I'm going to integrate that for 0.13.3

PR #239, opened 2015-07-27
Fix in command -> cmd

javapapo:
A small fix for the example from command to cmd


rhuss:
Thanks, I just merged it into branch `integration`

PR #240, opened 2015-07-27
resolves issue 90 redux

jgangemi:
- refactored 'queryContainerPortMapping' into 'getPortBindings'
- update dynamic ports in the run service
- cleaned up compiler warnings
- allow 'HostIp' to be bound as maven property and exported
- introduced 'PortMapping.Writer' to handle property file writing


rhuss:
@jgangemi do you think we can get this into 0.13.3 which I want to cut today ? If not thats not a big issue because I can do a quick release next week for 0.13.4, too.

BTW, I found out, when a variable is not yet set, it remains unresolved so that one could specify `${jolokia.port}` instead of `jolokia` in the portmapping. Should be easy to support both, I will add this when you are done. However, if `${jolokia.port}` is already set it gets resolved before injected to the plugin, so for your use case where you want to write out properties, its probably not useful. 

Except when you could live with a property format like 

```
<image-or-alias-name>.8080.port=312224
<image-or-alias-name>.8080.host=172.28.128.4
```

i.e. the fixed container port as part of a calculated property name. If this is ok, you then could use direct property syntax in your mappings like:

````
${myhost.ip}:${myport}:8080
````

so we don't need any artificial prefix for distinguishing host and property names. From a user interface perspective I tend to say that's the best solution (also you then don't have a direct coupling with pom variables and your code using these properties since they are calculated after a fix scheme).

What do you think about this ?

rhuss:
Ah, just saw that you already committed. sorry. 

Sorry for coming up late with my latest suggestion ;(. When the calculated syntax would be ok for you too, I could do the implementation, too.

rhuss:
Sorry for the confusion (I'm confused myself ;-)

I will add your changes to 0.13.3 and I think I can add the support for `${vars}` in mappings easily, too.

If we want to go for another property file format we could still introduce that later, maybe supporting both.

rhuss:
Merged in, thanks a lot for your efforts !

jgangemi:
lol - welcome :)

PR #241, opened 2015-07-27
resolves issue 236

jgangemi:
- fixed 'NullPointerException' thrown when no assembly was specified
- only return first 12 of image id


jgangemi:
i recommend applying this before #240

PR #248, opened 2015-08-01
fix docs: replace 'command' by 'cmd' (since v0.12.0)

### LINKED ISSUE #243: command param has been changed to cmd but some parts of the doc still refers to command
I stumbled upon that today.
I will provide a PR to fix that

looztra:
the 'command' configuration parameter has been replaced by 'cmd' since v0.12.0

fixes #243

### DIFF
--- a/doc/intro.md
+++ b/doc/intro.md
@@ -74,14 +74,20 @@ Let's look at a plugin configuration example:
       <name>jolokia/docker-demo:${project.version}</name>
 
       <build>
-         <from>java:8</from>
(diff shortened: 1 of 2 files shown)


rhuss:
Sorry, slipped through. I'm just about thinking to streamline the `cmd` in run with the variant used for creating the Dockerfile, since it seems they share the same semantics (its not so clear from to Docker docu whether this is really the case). 

I'll apply your PR for now, thanks !

PR #258, opened 2015-08-13
Added functionality to specify which http method to use when 'waiting'

DennisDenuto:
A signed PR of #255 for the 'integration' branch this time 

Signed-off-by: Dennis Leon <dennis.leon39@gmail.com>

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/StartMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/StartMojo.java
@@ -43,7 +43,7 @@ public class StartMojo extends AbstractDockerMojo {
      * @parameter property = "docker.follow" default-value = "false"
      */
     protected boolean follow;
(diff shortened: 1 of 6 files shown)


rhuss:
Thanks a lot for the PR ! I integrated it slightly modified so that the configuration looks like 

```xml
<wait>
    <http>
       <url>....</url>
       <method>GET</method>
       <status>200..399</status>
    </http>
 </wait>
```

I also added a status (range) which can be specified.

'hope that is ok for you.

This will be available in 0.13.3 which I want release today.

thanks again ...

DennisDenuto:
The status range is a good idea. Thanks!

rhuss:
thx ;) One could extend that to allow comma separated multiple ranges but I'm going to wait until the first one needs that :)

PR #269, opened 2015-08-23
#253: explicit exec arguments for run.cmd and run.entrypoint

### LINKED ISSUE #253: run.cmd - support explicit exec form
At the moment it is not possible to run a more complicated command on a container start, e.g. `docker run -it --rm busybox /bin/sh -c "while true; do echo Hello world; sleep 1; done"` as the cmd is split by spaces at  [ContainerCreateConfig.command](https://github.com/rhuss/docker-maven-plugin/blob/master/src/main/java/org/jolokia/docker/maven/access/ContainerCreateConfig.java#L35), and the following configuration: 

````xml
              <image>
                  <name>busybox</name>
                  <run>
                      <cmd>/bin/sh -c "while true; do echo Hello world; sleep 1; done"</cmd>
                      <wait>
                          <log>Hello world</log>
                      </wait>
                  </run>
              </image>
````
produces container startup error `true;: line 1: syntax error: unterminated quoted string`. 

It would be nice to have an ability to specify the exec form of the command explicitly, e.g.

````xml
              <image>
                  <name>busybox</name>
                  <run>
                      <cmd>
                          <params>
                              <param>/bin/sh</param>
                              <param>-c</param>
                              <param>while true; do echo Hello world; sleep 1; done</param>
                          </params>
                      </cmd>
                      <wait>
                          <log>Hello world</log>
                      </wait>
                  </run>
              </image>

````

in line with [recent changes to ENTRYPOINT/CMD](https://github.com/rhuss/docker-maven-plugin/commit/cc6f348324d958c2943cb25d91b2ab4cb8c4e1e )

### PR DESCRIPTION
**Pull request by mgurov**
fixes #253 

* explicit exec arguments for run.cmd and run.entrypoint
* refactored Arguments validation: invoked before mojo executions; includes deprecation message; on the downside revalidated for every mojo.
* refactored Arguments to support shortcuts `<cmd>blah</cmd>` and `<cmd><arg>blah</arg><arg>fooe</arg></cmd>`

Feel free to ask for documentation changes and better tests coverage. 

pom.xml used for integration (with maven) testing: https://gist.github.com/mgurov/8c9c991d95b5a95b91ff

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,7 @@
 # ChangeLog
 
 * **0.13.4**
+  - Support explicit exec arguments for `start.cmd` and `start.entrypoint`. (#253) 
   - Fix processing of split chunked JSON responses (#259)
   - Fix for default registry handling. Again and again. (#261)
   - Allow `runCmds` to be compressed into a single command with the build config option `optimise`. (#263)
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -394,6 +394,24 @@ or
  </entryPoint>
 ```
 
+Shorter forms supported:
+
+```xml
+<!-- shell form  -->
+ <entryPoint>java -jar $HOME/server.jar</entryPoint>
+```
+
+or 
+
+```xml
+ <entryPoint>
+    <!-- exec form  -->
+    <arg>java</arg>
+    <arg>-jar</arg>
+    <arg>/opt/demo/server.jar</arg>
+ </entryPoint>
+```
+
 ##### Docker Assembly
 
 With using the `inline`, `descriptor` or `descriptorRef` option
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by mgurov** [comment]
updated with the sign-off

**Comment by rhuss** [comment]
Thanks for the PR. An update of `manual.md` and `changelog.md` would be cool, indeed ;-)

Looks good to me, for the `Arguments` changes I have some questions, though (see comments).

**Comment by rhuss** [comment]
Ah, ok, some advanced Maven (or Plexus) fu (which I was not aware of ;) Do you have some pointer to documentation of this behaviour ? (just curious and always looking hard for good Maven docu)

So, if I understand your comments right:

```xml
<cmd>ls /tmp</cmd>
```

is the same as 

```xml
<cmd>
   <shell>ls /tmp</shell>
</cmd>
```

and 

```xml
<cmd>
  <arg>ls</arg>
  <arg>/tmp</arg>
</cmd>
```
is the same as

```xml
<cmd>
  <exec>
    <arg>ls</arg>
    <arg>/tmp</arg>
  </exec>
</cmd>
```

Right ?

Sounds goo to me, maybe you can add some inline comments (with examples) in `Argument` ? would be cool, because then I could reuse that as reference if I need similar advanced configuration constructs ;)

mgurov:
yes, your understanding is correct. I will add some pointers in javadocs and update the manuals shortly. Meanwhile you can find some of the related maven-fu described [here](http://blog.sonatype.com/2011/03/configuring-plugin-goals-in-maven-3/#.VeR3JbQ56Rv) .

jgangemi:
:+1: for the maven docs, there is some useful stuff in there!

rhuss:
Thanks for the link, really useful ;-). But I think we still need the `/** @parameter */` comments for the old behaviour, or ?

mgurov:
@rhuss the `/** @parameter */` isn't really necessary as the bean utils used by the maven would use the setters available. I have restored those comments for the sake of consistency though. 

Please find the new commit (pushed forcedly) with the mentioned changes and the docs updated.

rhuss:
Thanks a lot ! I applied to branch `integration` will be available in the next release which I plan for this week.

Issue #266, opened 2015-08-19
New feature docker exec


ginbalin:
Hello,

It would be nice to have an equivalent maven goal to the docker exec command. I would need in some case to be able to run a docker exec command previously started containers.

Or does it already exist?

rhuss:
No, it doesn't exist yet. Do you have a use case for me where you would need this (and can't easily use `docker exec` itself ?)

ginbalin:
Well I have a docker container that I want to start at the beginnig of my integration test then run a docker exec to start some extra stuff in the docker containers then run my tests and finally stop and delete everything.

rhuss:
I understand.

I think this will be a useful addition, though it might take a bit until its get implemented (however, we are happily integrating pull requests ;-)

jgangemi:
in the interim, couldn't you just build the container w/ an startup script that starts the extra services?

rhuss:
you could also use a different build configuration in a dedicated Maven profile which you use only for the integration tests.

Timing could also be an issue for when to execute the `docker:exec` task (would need to wait for the startup of the container, too)

DennisDenuto:
I don't mind giving this a go. 
Regarding the timing, I am thinking 3 'hooks' into the maven docker lifecycle. With an option to run a command based on a time interval or log message.
```
<exec>
   <preWait>command goes here</preWait>
   <postWait>command goes here</postWait>
   <preStop>command goes here</preStop>
   <wait>
      <time>10 seconds</time>
      <msg>log message to match against</msg>
      <cmd>command goes here</cmd>
   </wait>
</exec> 
```

However I think initially I will work on just getting 'postWait' tag working. (as it seems to fit the use case above.). 

What do you think?

jgangemi:
i think it looks ok. you can probably leverage much, if not all, of the existing `wait` code for this.

i also advise working off the `integration` branch of the project instead of `master`.

DennisDenuto:
cool thanks

rhuss:
I wonder whether we shouldn't include that into the `docker:start` goal and the extend the `wait` config accordingly section. 

The benefits would be:

* No additional goal (yes, that's a benefit ;-)
* No duplication in `<wait>` configurations. Also, the `exec` goal would only be callled after or before `start` and hence might later than necessary.

What I can think about would be somethings like extension for the wait section in `<run>` configuration:

```xml
<wait>
  <http>
    <url>http://localhost:${host.port}</url>
[...]

PR #272, opened 2015-08-26
Add logic to specify exec commands during postStart and preStop

DennisDenuto:
PR for #266
e.g. 
```
<wait>
  <http>
    <url>http://localhost:${host.port}</url>
    <method>GET</method>
    <status>200..399</status>
  </http>
  <time>10000</time>
  <shutdown>500</shutdown>
  <exec>
    <postStart>/opt/init_db.sh<postStart>
    <preStop>/opt/notify_end.sh</preStop>
  </exec>
</wait>
```

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -13,6 +13,7 @@
   - HTTP method and status code can be specified when waiting on an HTTP URL (#258)
   - Introduced global `portPropertyFile` setting (#90)
   - Allow the container's host ip to be bound to a maven property and exported
+  - Add logic to specify exec commands during postStart and preStop (#272)
 
 * **0.13.2**
   - "run" directives can be added to the Dockerfile (#191)
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -806,7 +806,10 @@ some condition is met. These conditions can be specified within a
 * **shutdown** is the time to wait in milliseconds between stopping a container
   and removing it. This might be helpful in situation where a Docker croaks with an
   error when trying to remove a container to fast after it has been stopped.
-
+* **exec** Specifies commands to execute during specified lifecycle of the container. It knows the following sub-elements:
+  - **postStart** Command to run after the above wait criteria has been met
+  - **preStop** Command to run before the container is stopped.
+  
 As soon as one condition is met the build continues. If you add a
 `<time>` constraint this works more or less as a timeout for other
 conditions. The build will abort if you wait on an url or log output and reach the timeout. 
@@ -823,6 +826,10 @@ Example:
   </http>
   <time>10000</time>
   <shutdown>500</shutdown>
+  <exec>
+     <postStart>/opt/init_db.sh</postStart>
+     <preStop>/opt/notify_end.sh</preStop>
+  </exec>
 </wait>
 ```` 
 
--- a/src/main/java/org/jolokia/docker/maven/StartMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/StartMojo.java
@@ -93,6 +93,10 @@ public class StartMojo extends AbstractDockerMojo {
 
                 // Wait if requested
                 waitIfRequested(dockerAccess,imageConfig, projProperties, containerId);
+                WaitConfiguration waitConfig = runConfig.getWaitConfiguration();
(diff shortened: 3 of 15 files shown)


rhuss:
Thanks for the PR, I just did a quick review. Looks real good overall. Some minor remarks:

* Could you please add some documentation to manual.md (with example) and to changelog.md ?
* /wr to the possible options when attaching to a running container: I never would expect that a tty would be set for a maven build, so I would remove `Tty` all together. Also I always would expect to attach to StdErr and StdOut and currently its also not possible to set it from the configuration. So I would remove that, too, and then maybe the whole `ExecContainerConfiguration` since all what is left is the command. We could make it a `org.jolokia.docker.maven.config.Arguments` as it is for the other shell-command related config options, too (though I don't know by heart whether there are also two forms (exec/shell) for calling the exec container.
* I would remove `Optional` in ShutdownAction because this can be realized easily with a null-check, too and wouldn't introduce a new dependency.

A general principle I try to follow is to implement only what is needed (and don't fall in the trap *we might it need it sometime* as I did top often and frankly nearly never happened) and even when I think things like `Optional` are a good thing, I want to introduce that only when there is a good reason weighting out the costs of an extra dependency. From my feeling this is not yet the case yet.

DennisDenuto:
Thanks for the feedback!
From what I can see the /exec endpoint only accepts 'exec' not 'shell'. (Not sure if that changes your opinion on whether Arguments should still be used since it maintains both shell and exec.)

wrt Optional, I figured it was ok to use since guava was on the classpath and was only a private helper method. But I agree looking back now it looks out of place considering null checks are mainly used. :-)

DennisDenuto:
Hey Rhuss,
were you able to check the changes made to this PR? (following your feedback).

rhuss:
Ooops, didn't get any update by GitHub that the PR was changed. Will have a look ASAP.

rhuss:
Thanks again ! I'm going to cleanup the duplicate code in `ShutdownAction` and `RunService`, too, since there is now some overlap here.

DennisDenuto:
Cool Thanks heaps Rhuss

robzet:
Can I use this to run more than one command? I cant seam to get it working. I have tried with two commands separated by ; or && (as &amp;&amp;). Only one command is ran.I can see both commands getting sent over rest interface as a "Cmd": "[]" but I get errors related to it and it does not look like its interpreted as two commands correctly. Any ideas?

As i believe the first answer will be "put the commands in a script and execute that in one command" I want to also add some background to the problem. My setup runs well on Linux, but not on Windows. The command is supposed to run a script file with several commands in. On Windows this file looses its permissions somehow during transfer to the docker container and this is why I need to run two commands. One to restore +x and one to execute the script.

rhuss:
There is a known issue with Windows which looses the execution permission. For a complete discussion please refer to #477, but in short with 0.15.9 you have the following options:

* You use an assembly with an assembly descriptor which adds your `postExec` script. In this assembly descriptor you can add a file mode for the entries in a fileSet. Then you create the assembly with a `<mode>zip</mode>` (or tar, but not `<mode>dir</mode>` since you will loose the permission again).
* You add a `<permissions>exec</permissions>` to the assembly descriptor which will add the exec bit to every file. A somewhat ugly workaround but the way that Docker recommends.

Does this help ?

As a workaround you could also use 'sh /path/to/my/script' as command to explicitly call your script, too, so you dont have to rely on the execution bit.

robzet:
@rhuss Thank you very much! That worked well. I can now run the script because the execute permissions are kept on windows as well after changing the assembly mode to 'tar' in the pom-file.

burtsevyg:
My configuration
 ```
                                        <wait>
                                            <time>30000</time>
                                            <exec>
                                                <postStart>/opt/postStart.sh</postStart>
                                            </exec>
                                        </wait>
```
/opt/postStart.sh return 1 exit code.
Why postStart script did not fail maven build?
How can I do this.

rhuss:
Currently the exit status of a `postStart` is not checked at all (would require an extra API call). That's probably not that difficult, but its simply not implemented yet.

If you think this would be a useful addition, then please open an extra issue for that, describing your use case. thanks !

PR #281, opened 2015-09-14
fix to add support for multiple docker services

nkrendel:
I ran across this problem with trying to use the plugin to stop/remove/build/start images on two docker hosts.  ServiceHub was only initialized once and so the second docker configuration used the services that were initialized with the old docker url.  This is an attempt to fix that.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -525,7 +525,7 @@ The `<run>` configuration knows the following sub elements:
   the container.
 * **capDrop** (*v1.14*) a list of `drop` elements to specify kernel parameters to remove
   from the container.
-* **command** is a command which should be executed at the end of the
+* **cmd** is a command which should be executed at the end of the
   container's startup. If not given, the image's default command is
   used. See [Start-up Arguments](#start-up-arguments) for details.
 * **domainname** (*v1.12*) domain name for the container
@@ -602,7 +602,7 @@ Example:
     <date>ISO8601</date>
     <color>blue</color>
   </log>
-  <command>java -jar /maven/docker-demo.jar</command>
+  <cmd>java -jar /maven/docker-demo.jar</cmd>
 </run>
 ````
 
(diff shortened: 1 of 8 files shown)


rhuss:
Interesting use case (which I never thought about ;-). SInce the Docker URL is a plugin-global param, I guess you have a multi-module setup ? Would you mind to post some example for this setup so that I can reproduce it ?

nkrendel:
Yes it's a multi-module setup.  The root pom defines the plugin and has <skip>true</skip>.  Then there are two sub-modules that each build a war.  One sub-module builds the regular api endpoint war and another sub-module builds the admin api endpoint war.  These are deployed to different hosts so each sub-module has a <configuration> with a different dockerHost.  e.g.:

```
                <configuration>
                    <dockerHost>tcp://api-host.company.com:2376</dockerHost>
                    <verbose>false</verbose>
                    <skip>false</skip>
                    <sourceDirectory>src/main/docker</sourceDirectory>
                    <images>
                        <image>
                            <name>wildfly-api</name>
                            <alias>service</alias>
                            <build>
                                <cleanup>true</cleanup>
                                <tags>
                                    <tag>latest</tag>
                                    <tag>${project.version}</tag>
                                </tags>
                                <assembly>
                                    <dockerFileDir>wildfly-api</dockerFileDir>
                                </assembly>
                            </build>
                            <run>
                                <ports>
                                    <port>8080:8080</port>
                                    <port>9990:9990</port>
                                </ports>
                            </run>
                        </image>
                    </images>
                </configuration>
```
Please let me know if you'd like me to post more pieces from this example.

rhuss:
Thanks, I will tackle the issue for the next (0.13.6) release since its a bit more involved (and needs some extra thoughts)

rhuss:
I refactored the usage of the `ServiceHub` so it is recreated for each run. Multiple Docker hosts should be now be easily possible. This was added in course of #320 .

I pushed a version `0.13.7-SNAPSHOT` with this fix. Could you please check whether this solves your use case ?

nkrendel:
Support for multiple docker hosts seems to be working in ```0.13.7-SNAPSHOT```.

rhuss:
cool. I'll close this PR, feel free to reopen it or create an issue if there are still problems.

PR #287, opened 2015-09-16
Add logconfig option to write logs to specified file

ntrvic:
We are using this plugin at Conductor and would really benefit from this addition. This PR will allow the user to pass the 'file' log configuration option for a given image and specify a location for the logs to be written to. For example:

```
<log>
    <file>${basedir}/app.out</file>
</log>
```

If the file option is empty, it will print to System.out, otherwise it will use FileOutputStream and print to the specified location.


rhuss:
Thanks !

If you don't mind, I'm just polished it a bit:

* No color if logging to a file
* Append to file (user must rotate on its own)
* Add documentation 
* Refactoring (the map holding the streams is not needed)
* Add a property `docker.logStdout` to overwrite any configured file (useful when doing a `docker:log`)

rhuss:
Will be available in the next release.

Issue #285, opened 2015-09-16
Extend assembly to allow extracted inclusion of dependencies


rhuss:
The use case is, that the run-script from https://github.com/fabric8io/run-java-sh which is also deployed as a maven artefact can be included in a Docker image easily.

Maybe there is already sth like this (e.g. with a special packaging). Some R&D is needed here probably.

Issue #284, opened 2015-09-16
Expose Tar-File generated for docker:build


rhuss:
The tar file which we are creating for `<docker:build>` could be easily exposed. That way it can be carried around and, if properly done, the code in the tar could be used standalone, too, without creating a docker image.

Maybe even introduce an own packaging ("docker") so that this tar can be deployed to a maven repository. This would work nicely in a multi module setup with one dedicated module only for building docker image and creating this tar archive.

Issue #283, opened 2015-09-16
Add option <createClasspath> to <build>


rhuss:
If given within a `<build>` section like in 

```
<build>
  <!-- ... -->
 <assembly>
   <createClasspath>run-classpath</createClasspath>
   ....
    <basedir>/opt/demo</basedir>
    <descriptor>assembly.xml</descriptor>
  </assembly>
</build>
```

the plugin will add a classpath to the given path within the basedir and add it to the Dockerfile. This file should also respect any filename mappings happening in the assembly.xml.

This can be done by re-using the classpath resolution used in the `maven-dependency-plugin` (unfortunately not easy reusable except when subclassing) and the [MappingTrackingArchiver](https://github.com/rhuss/docker-maven-plugin/blob/master/src/main/java/org/jolokia/docker/maven/assembly/MappingTrackArchiver.java) we are using for the watch feature. If this project is of packaging `jar` or `war` then it should be added to the classpath, too.  

The use case is for flat classpath apps which come with a startup script which can use this generated file for setting up the classpath properly.

rhuss:
There is a super easy solution by using the `maven-dependency-plugin`:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-dependency-plugin</artifactId>
  <version>2.10</version>
  <executions>
    <execution>
      <id>attach-deps</id>
      <phase>package</phase>
      <goals>
        <goal>build-classpath</goal>
      </goals>
      <configuration>
        <prefix>.</prefix>
        <includeScope>runtime</includeScope>
        <outputFile>${project.build.directory}/classpath</outputFile>
      </configuration>
    </execution>
  </executions>
</plugin>
```

This will create a file `classpath` in `target/` which will be automatically picked up by the `docker-maven-plugin` when using the assembly descriptor `artifactWithDependencies` (for the next version).

osigida:
I've solved it with existing tool set like that
```
                                    <assembly>
                                        <basedir>/</basedir>
                                        <inline>
                                            <dependencySet>
                                                <outputDirectory>jars</outputDirectory>
                                            </dependencySet>
                                        </inline>
                                 </assembly>
```

this will copy all the dependencies to `/jars` folder

then start up script looks like
```
                                    <entryPoint>
                                        <exec>
                                            <arg>java</arg>
                                            <arg>-cp</arg>
                                            <arg>jars/*</arg>
                                            <arg>${mainclass}</arg>
[...]

PR #288, opened 2015-09-16
Support docker:watch correlating maven jars that end up in it's assem…

chirino:
…bly dir so that a docker container update can be triggered if transitive dep jar is installed in the mvn local repo.

Mostly a proof of concept.  Using this dev:watch works on hawt-app based assemblies.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/WatchMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/WatchMojo.java
@@ -15,16 +15,21 @@ package org.jolokia.docker.maven;/*
  * limitations under the License.
  */
 
+import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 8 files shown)


rhuss:
Awesome, thanks, Hiram ! I will look at it ASAP (today, hopefully) .... 

btw, do you have any comments on #284, #283 or #285 ? I really want to solve our gordian knot asap, not necessarily by implementing these features (although they would be a solution for the offline use case).

rhuss:
I had a look at the PR:

* The extraction for the maven coordinates is quite smart, like it :)
* However: Dependencies in the Maven assembly are already have the local repository path as source. I guess you had an other use case in mind where you need to find out this path on your own (i.e. not jar dependencies)
* For the artefact of this pom.xml, I think it is not needed to watch the repository variant but the one created in `target/` (so that a `mvn package` would also trigger a rebuild)

Some improvement coming to mind, too:

* We could ignore all non SNAPSHOT dependencies from watching, or ? That would speed up things a bit. 
* I injected the local repository directly into the tracker (via a session) so no need to pass it down as argument (done).

I will refactor the PR a bit (have seen some redundant directory name movememnt, too) and merge it into branch `integration` so we can continue from there on.

rhuss:
I merged the changes in by hand. thanks again.

chirino:
cool looks good!

PR #291, opened 2015-09-17
Fixed NullpointerException ...

christophd:
... when plugin is executed in Maven module without image configuration. This can be the case when using the plugin in a multi module Maven project. I was not able to call docker:build on parent Maven module as there was no image configuration but in sub modules there were.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
@@ -1,19 +1,6 @@
 package org.jolokia.docker.maven;
 
-import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jolokia/docker/maven/AbstractDockerMojo.java
```diff
@@ -183,8 +167,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     }
 
     private void validateConfiguration(Logger log) {
-        for (ImageConfiguration imageConfiguration : images) {
-            imageConfiguration.validate(log);
+        if (images != null) {
```
mgurov: Auch, my  fault via 3a040e00f5b7e3d5760960e8f161ade82ac995ae . Apologies.

rhuss:
Thanks ! will do a release soon, probably early next week.

PR #294, opened 2015-09-18
add killafter parameter to image:run:wait

### LINKED ISSUE #293: Add a parameter to configure the number of seconds to wait before killing the container
I would like to expose the **t** parameter of the stop call (see https://docs.docker.com/reference/api/docker_remote_api_v1.20/#stop-a-container) to the docker-maven-plugin configuration.

This defaults to 0 in docker. Therefore, for the process running inside there is no change of a proper shutdown, when receiving SIGTERM.

In my use case the process writes some log data, when SIGTERM is received before exiting gracefully.

With the current docker-maven-plugin this file is empty.

I extended docker-maven-plugin in a fork and it is working fine there.

fiji-flo:
Sets the delay between docker stop and kill. This resolves #293
Signed-off-by: Florian Merz <florian.merz@goeuro.com>

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -822,6 +822,7 @@ some condition is met. These conditions can be specified within a
 * **log** is a regular expression which is applied against the log
   output of an container and blocks until the pattern is matched.
 * **time** is the time in milliseconds to block.
+* **killafter** is the time in seconds between sending SIGTERM and SIGKILL when stopping a container
 * **shutdown** is the time to wait in milliseconds between stopping a container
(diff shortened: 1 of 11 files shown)


rhuss:
Thanks a lot !

I adapted it a bit for consistencies sake:

* Used `<kill>` instead of `<killAfter>` to make it simpler. Since the context is the `<wait>` section it should be somewhat clear what is meant.
* Used milliseconds instead of second to be consistent with the other time values (time, shutdown).

fiji-flo:
Nice, thank you.

PR #297, opened 2015-09-25
Disable Apache HTTP Client retry in WaitUtil

james-bright-helix:
WaitUtil has it's own retry mechanism and by disabling the HTTP Client retries it is easier to debug issues with containers starting.

Also HTTP Client retries provide superfluous logging considering it is highly likely the system will be unavailable when the first check occurs.

Signed-off-by: James Baker <james@brighthelix.co.uk>


rhuss:
Thanks, looks good.

PR #306, opened 2015-10-12
Added support for TCP port wait condition

### LINKED ISSUE #304: Add port wait condition
An additional wait condition, which waits for specific ports, would be useful. The configuration could look like this:

```xml
<wait>
  <port>
    <tcp>3306</tcp>
    <pollIntervall>500</pollIntervall>
    <pollLimit>20</pollLimit>
  </port>
</wait>
```

I want to work on this and make a pull request and I am interested in your opinions.

svenwltr:
Example:

```xml
<wait>
   <time>60000</time>
   <tcp>
      <ports>
         <port>3306</port>
         <port>9999</port>
      </ports>
</wait>
```

This will setup a TCP port checker for ports 3306 and 9999. The test will time out after 60 seconds and the poll interval is hard coded, just like the HTTP wait condition.

This configuration differs from the proposed one in the issue #304:

* I remove the `<interval>` property, because this is already hardcoded with `org.jolokia.docker.maven.util.WaitUtil#WAIT_RETRY_WAIT` in `org.jolokia.docker.maven.util.WaitUtil#wait()`.
* Added support for multiple port checks.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/StartMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/StartMojo.java
@@ -61,14 +61,14 @@ public class StartMojo extends AbstractDockerMojo {
         getPluginContext().put(CONTEXT_KEY_START_CALLED, true);
 
         Properties projProperties = project.getProperties();
-        
+
         QueryService queryService = serviceHub.getQueryService();
         RunService runService = serviceHub.getRunService();
 
         LogDispatcher dispatcher = getLogDispatcher(dockerAccess);
 
         PortMapping.PropertyWriteHelper portMappingPropertyWriteHelper = new PortMapping.PropertyWriteHelper(portPropertyFile);
-        
+
         boolean success = false;
         try {
             for (StartOrderResolver.Resolvable resolvable : runService.getImagesConfigsInOrder(queryService, getImages())) {
@@ -104,7 +104,7 @@ public class StartMojo extends AbstractDockerMojo {
                 runService.addShutdownHookForStoppingContainers(keepContainer,removeVolumes);
                 wait();
             }
-            
+
             portMappingPropertyWriteHelper.write();
             success = true;
         } catch (InterruptedException e) {
@@ -159,13 +159,18 @@ public class StartMojo extends AbstractDockerMojo {
                     host = projectProperties.getProperty("docker.host.address");
                 }
 
-                if ("localhost".equals(host)) {
+                if ("localhost".equals(host) && container.getIPAddress() != null) {
                     host = container.getIPAddress();
                     ports = tcpConfig.getPorts();
                     log.info(String.format("%s: Waiting for ports %s directly on container with IP (%s).", imageConfig.getDescription(), ports, host));
                 } else {
                     for (int port : tcpConfig.getPorts()) {
                         Container.PortBinding binding = container.getPortBindings().get(port + "/tcp");
+                        if (binding == null) {
+                            throw new MojoExecutionException(String.format(
+                                    "Cannot watch on port %d, since it was not bind by Docker.", port
+                            ));
+                        }
                         ports.add(binding.getHostPort());
                     }
                     log.info(String.format("%s: Waiting for exposed ports %s on remote host (%s), since they are not directly accessible.", imageConfig.getDescription(), ports, host));
--- a/src/main/java/org/jolokia/docker/maven/StartMojo.java
+++ b/src/main/java/org/jolokia/docker/maven/StartMojo.java
@@ -9,6 +9,7 @@ package org.jolokia.docker.maven;
  */
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 import java.util.regex.Pattern;
 
@@ -25,6 +26,7 @@ import org.jolokia.docker.maven.config.LogConfiguration;
 import org.jolokia.docker.maven.config.RunImageConfiguration;
 import org.jolokia.docker.maven.config.WaitConfiguration;
 import org.jolokia.docker.maven.log.LogDispatcher;
+import org.jolokia.docker.maven.model.Container;
 import org.jolokia.docker.maven.service.QueryService;
 import org.jolokia.docker.maven.service.RunService;
 import org.jolokia.docker.maven.util.StartOrderResolver;
@@ -144,6 +146,39 @@ public class StartMojo extends AbstractDockerMojo {
         if (wait.getLog() != null) {
             checkers.add(getLogWaitChecker(wait.getLog(), docker, containerId));
             logOut.add("on log out '" + wait.getLog() + "'");
+        }
+        if (wait.getTcp() != null) {
+            WaitConfiguration.TcpConfiguration tcpConfig = wait.getTcp();
+            try {
+                Container container = docker.inspectContainer(containerId);
+                String host = tcpConfig.getHost();
+                List<Integer> ports = new ArrayList<>();
+
+                if (host == null) {
+                    // Host defaults to ${docker.host.address}.
+                    host = projectProperties.getProperty("docker.host.address");
+                }
+
+                if ("localhost".equals(host)) {
+                    host = container.getIPAddress();
+                    ports = tcpConfig.getPorts();
+                    log.info(String.format("%s: Waiting for ports %s directly on container with IP (%s).", imageConfig.getDescription(), ports, host));
+                } else {
+                    for (int port : tcpConfig.getPorts()) {
+                        Container.PortBinding binding = container.getPortBindings().get(port + "/tcp");
+                        ports.add(binding.getHostPort());
+                    }
+                    log.info(String.format("%s: Waiting for exposed ports %s on remote host (%s), since they are not directly accessible.", imageConfig.getDescription(), ports, host));
+                }
+
+                WaitUtil.TcpPortChecker tcpWaitChecker = new WaitUtil.TcpPortChecker(host, ports);
+                checkers.add(tcpWaitChecker);
+                logOut.add("on tcp port '"+tcpWaitChecker.getPending()+"'");
+
+            } catch (DockerAccessException e) {
+                throw new MojoExecutionException("Unable to access container.", e);
+            }
+
         }
 
         if (checkers.isEmpty()) {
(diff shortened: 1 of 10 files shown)


svenwltr:
My current approach isn't compatible with `docker-machine`. So I have take a look on this, first.

rhuss:
If you mean, that you can't use `localhost` as default address, then you are right ;-). I would use `${docker.host.address}` as default, which points to the IP of the docker daemon (which is almost always the right choice).

rhuss:
I'll postpone this feature to the next release since I'm going to release d-m-p 0.13.6 today. hope you don't mind.

svenwltr:
I have problems detecting port status with `${docker.host.address}`. It points to `127.0.0.1`, which is basically correct. However, the port check will always success, when I use the host `127.0.0.1`, even if the port isn't opened, yet. Whereas the check on `172.17.42.1` is correct, although this is the same host. This must be a bug with Docker, since I get the same result with a command line check.

The used host on this PR falsely is `0.0.0.0`, but it is magically working :flushed:

svenwltr:
As mentioned in docker/docker#16976 a port check on `localhost` isn't possible without disabling `userland-proxy`. I don't like this solution since everybody would have to change the default option for the docker daemon.

Possible solutions:

* Access docker containers directly, but I have no idea whether and how this is possible using `docker-machine`.
* Detect docker bridge IP from host  (ie `172.17.42.1`).

rhuss:
I see the point (userland proxy will hopefully [vanish soon](https://github.com/docker/docker/issues/14856)). 

Yet another solution could be to not rely on connect but send some also some package and detect (maybe with some timeout ?) whether it is received. Could work for tcp but I'm really not network expert and don't know whether this is possible at all and whether tools like hping2 does something similar.

svenwltr:
I have now implemented the following behavior:

* Added `<host>` configuration, which defaults to `${docker.host.address}`.
* When the given `host` is `localhost`, then the containers will accesses directly with their respective IPs and the original port (ie. `3306` for MySQL).
* When the given `host` isn't `localhost`, then this `host` will be used with the exposed ports.

What do you think about this? Also, I can squash the commits before merge. They got a little messy.

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,43 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
             logOut.add("on log out '" + wait.getLog() + "'");
         }
+        if (wait.getTcp() != null) {
+            WaitConfiguration.TcpConfiguration tcpConfig = wait.getTcp();
+            try {
+                Container container = docker.inspectContainer(containerId);
+                String host = tcpConfig.getHost();
+                List<Integer> ports = new ArrayList<>();
+
+                if(host == null) {
```
bjoernhaeuser: I think there are missing spaces before the `(`

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,43 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
+                    // Host defaults to ${docker.host.address}.
+                    host = projectProperties.getProperty("docker.host.address");
+                }
+
+                if(host == "localhost") {
```
bjoernhaeuser: you have to use equals here :)

src/main/java/org/jolokia/docker/maven/util/WaitUtil.java
```diff
@@ -176,6 +181,68 @@ public void cleanUp() { }
...
+
+                try {
+                    Socket s = new Socket();
+                    s.connect(address, TCP_TIMEOUT);
+                    s.close();
+
+                    iter.remove();
+
+                } catch (IOException e) {
+                    // Ports isn't opened, yet. So don't remove von queue.
```
bjoernhaeuser: from*

rhuss:
Thanks, I will have a look asap (but being busy quite a bit at the moment ;-(

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,43 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
+                    host = container.getIPAddress();
+                    ports = tcpConfig.getPorts();
+                    log.info(String.format("%s: Waiting for ports %s directly on container with IP (%s).", imageConfig.getDescription(), ports, host));
+
+                } else {
+                    for (int port : tcpConfig.getPorts()) {
+                        Container.PortBinding binding = container.getPortBindings().get(port + "/tcp");
+                        ports.add(binding.getHostPort());
+                    }
+                    log.info(String.format("%s: Waiting for exposed ports %s on remote host (%s), since they aren't accessible directly.", imageConfig.getDescription(), ports, host));
```
bjoernhaeuser: Would make that logmessage ending with:

`since they are not directly accessible.`

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,43 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
+
```
bjoernhaeuser: Suggesting to remove that empty line here :)

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,43 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
+                }
+
+
+                WaitUtil.TcpPortChecker tcpWaitChecker = new WaitUtil.TcpPortChecker(host, ports);
+                checkers.add(tcpWaitChecker);
+                logOut.add("on tcp port '"+tcpWaitChecker.getPending()+"'");
+
+            } catch (DockerAccessException e) {
+                throw new MojoExecutionException("Unable to access container.", e);
+
```
bjoernhaeuser: also this line can be removed ;)

src/main/java/org/jolokia/docker/maven/model/ContainersListElement.java
```diff
@@ -76,6 +76,12 @@ public String getName() {
     }
 
     @Override
+    public String getIPAddress() {
+        // IP address isn't provides by container list action.
```
bjoernhaeuser: `is not provided`:)

bjoernhaeuser:
There are some minos code-style violations (missing space), but nothing serious :)

:+1: from me, that extension is pretty useful :)

bjoernhaeuser:
Is there anything we can do to make that go live? :)

rhuss:
I plan to make a release this week and this PR is the next I'm going to integrate. So chances are good that it will live before santa claus  ;-)

src/main/java/org/jolokia/docker/maven/model/ContainerDetails.java
```diff
@@ -62,6 +63,18 @@ public String getName() {
     }
 
     @Override
+    public String getIPAddress() {
+        if (json.has(NETWORK_SETTINGS) && !json.isNull(NETWORK_SETTINGS)) {
+            JSONObject networkSettings = json.getJSONObject(NETWORK_SETTINGS);
+            if (!networkSettings.isNull(PORTS)) {
```
rhuss: Shouldn't we check here for `IP` ? Returning a null IP address when no ports are configured although there would be one seems not so intuitive.

src/main/java/org/jolokia/docker/maven/util/WaitUtil.java
```diff
@@ -176,6 +181,68 @@ public void cleanUp() { }
 
     // ====================================================================================================
 
+    /**
+     * Check whether a given TCP port is available
+     *
+     */
+    public static class TcpPortChecker implements WaitChecker {
+        private static final int TCP_TIMEOUT = 2000;
```
rhuss: Is 2 seconds not a bit long ? For the HTTP checker it 0.5 seconds. Remember, this time out is only for one try (there can be multiple tries until the overall timeout has reached).

src/main/java/org/jolokia/docker/maven/util/WaitUtil.java
```diff
@@ -176,6 +181,68 @@ public void cleanUp() { }
...
+
+        public List<InetSocketAddress> getPending() {
+            return pending;
+        }
+
+        @Override
+        public boolean check() {
+            Iterator<InetSocketAddress> iter = pending.iterator();
+
+            while (iter.hasNext()) {
```
rhuss: For a better performance the checks should be done in parallel with threads. This check is perfectly suitable for that since most of the time its blocking.

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,39 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
+                String host = tcpConfig.getHost();
+                List<Integer> ports = new ArrayList<>();
+
+                if (host == null) {
+                    // Host defaults to ${docker.host.address}.
+                    host = projectProperties.getProperty("docker.host.address");
+                }
+
+                if ("localhost".equals(host)) {
+                    host = container.getIPAddress();
```
rhuss: Host can be null (e.g. when no ports are exposed). Could this be an issue here ?

src/main/java/org/jolokia/docker/maven/StartMojo.java
```diff
@@ -145,6 +147,39 @@ private void waitIfRequested(DockerAccess docker, ImageConfiguration imageConfig
...
+                    ports = tcpConfig.getPorts();
+                    log.info(String.format("%s: Waiting for ports %s directly on container with IP (%s).", imageConfig.getDescription(), ports, host));
+                } else {
+                    for (int port : tcpConfig.getPorts()) {
+                        Container.PortBinding binding = container.getPortBindings().get(port + "/tcp");
```
rhuss: binding can be null, too, when there configuration doesn't fit the port binding configuration. (leading to a NPE in the following line). This should be dealt with here (probably by throwing an exception because its an configuration mismatch).

src/main/java/org/jolokia/docker/maven/util/WaitUtil.java
svenwltr: I don't think the performance impact justifies the complexity of using a concurrent check. It is true, that this test is blocking the most of the time, but these checks are waiting for the container to be ready, anyway.

rhuss:
Thanks for the fix ! I will do a release over the weekend with this included.

nice weekend ....

PR #324, opened 2015-11-13
Update dynamic host prop even without dynamic port prop

caspark:
Fixes #323 (see that for background).

PR #331, opened 2015-11-16
Update dynamic host prop even without dynamic port prop

### LINKED ISSUE #323: Dynamic host property in port mapping is not filled in if port is specified
I have a port mapping config like the following:

```
<run>
    <ports>
        <port>+docker.ondemand.postgres.host:docker.ondemand.postgres.port:5432</port>
    </ports>
</run>
```

Ordinarily, I can access `${docker.ondemand.postgres.host}` & `${docker.ondemand.postgres.port}` after the docker container has started.

But if I specify `-Ddocker.ondemand.postgres.port=5432` then `docker.ondemand.postgres.host` is not dynamically set.

This appears to be because of the following snippet in `org.jolokia.docker.maven.service.RunService#createAndStartContainer`

        if (mappedPorts.containsDynamicPorts()) {
            updateMappedPorts(id, mappedPorts);
        }

Changing it to

        if (mappedPorts.containsDynamicPorts() || mappedPorts.containsDynamicHostIps()) {
            updateMappedPorts(id, mappedPorts);
        }

seems to fix this bug, so I'll raise a PR for this.

caspark:
Fixes #323

Signed-off-by: Caspar Krieger &lt;email goes here&gt;

Replaces #324 .

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/service/RunService.java
+++ b/src/main/java/org/jolokia/docker/maven/service/RunService.java
@@ -96,7 +96,7 @@ public class RunService {
         String id = docker.createContainer(config, containerName);
         startContainer(imageConfig, id);
 
-        if (mappedPorts.containsDynamicPorts()) {
+        if (mappedPorts.containsDynamicPorts() || mappedPorts.containsDynamicHostIps()) {
             updateMappedPorts(id, mappedPorts);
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks a lot, looks good to me.

PR #333, opened 2015-11-17
resolves issue-87

jgangemi:
it's a glorious day! :)

couple of things...

1) i'm not sure how you intended for the `runId` to work, but i had to add an option to ignore it when calling `docker:stop` from a standalone context. i had already made a change (i forget when) that fixed this for containers that were started/stopped during the same build run. i didn't remove any of that code, i just added an option to ignore the run id.

2). i bumped things to `0.14-0-SNAPSHOT` b/c this required the docker api to be rev-ed. `v1.18` is still part of docker 1.6, so perhaps it can stay at `0.13.7-SNAPSHOT`.

3) there is still the option to make the plugin work like it did before using `docker.sledgehammer` (and i really hope we can keep that name)

4) i did not bother to account for containers that may have been created using an older version of the plugin and have upgraded to whatever version this goes out as. it didn't seem worth the extra work when the containers can be stopped using `-Ddocker.sledgehammer` and then re-created w/ `docker:start` and have the labels applied.

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/util/PomLabel.java
+++ b/src/main/java/org/jolokia/docker/maven/util/PomLabel.java
@@ -1,4 +1,5 @@
-package org.jolokia.docker.maven.util;/*
+package org.jolokia.docker.maven.util;
+/*
  * 
  * Copyright 2014 Roland Huss
  *
@@ -25,12 +26,11 @@ import java.util.UUID;
  */
 public class PomLabel {
 
-    private String mavenCoordinates;
-    private String runId;
-
-
     // Environment variable used to label containers
     public static final String CONTAINER_LABEL_NAME = "docker.maven.plugin.container";
+    
+    private String mavenCoordinates;
+    private String runId;
 
     /**
      * Construct from a given label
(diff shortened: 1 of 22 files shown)


jgangemi:
i'd also really appreciate it if you could prioritize looking this over so i can address any feedback, etc as it's crucial for some stuff going on at the day job.

rhuss:
I'll try to look at the asap but I'm afraid that won't be before mid of next week, because of release deadline **and** preparing a talk for a conference next week. After that I'll have much more time to spend on the plugin in order to push it forward. Sorry for that.

In the meantime, could you check this PR again ? Seems that quite some old commits were included (including mine). Doesn't feel right ...

jgangemi:
i'll take a look. i pulled in your branch and then re-based the latest integration branch on top of that, but maybe i did something wrong.

no worries about not being able to get to this. we can use a snaphot version of the plugin that i can push up.

src/main/java/org/jolokia/docker/maven/util/PomLabel.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param label label as stored with the container
+     */
+    public PomLabel(String label) {
+        String[] parts = label.split(":");
+        if (parts.length != 4 && parts.length != 3) {
+            throw new IllegalArgumentException("Label '" + label +
+                                               "' has not the format <group>:<artifact>:<version>[:<runId>]");
+        }
+        mavenCoordinates = parts[0] + ":" + parts[1] + ":" + parts[2];
+        runId = parts.length == 4 ? parts[3] : null;
```
rhuss: Is there a reason why runId is here null by default ? That's inconsistent to the constructor with no runId which initializes the run id to an UUID.

jgangemi: i'm going to have to defer that back you to you - looking at your branch, it would seem you wrote that code. :)

i had to alter `matches` to allow comparison against the `runId` to be optional b/c there's no way the plugin can know that value when it's called directly via `docker:stop`, so in order to stop all the containers, we can only match on the coordinates.

as an aside, i'm not sure the `runId` is needed on a certain level b/c the plugin already only stops/removes containers it started during integration tests and/or if there is some problem during initial start up. having said that, i still think it's a good idea to tag them.

rhuss: ;-P well, was quite some time ago (and I'm getting old :) ....

Thinking again, I also think that a `runId` is probably not needed for this use case but could be helpful to identify all containers which has been started together with a single 'mvn docker:start' call. But I wonder whether we shouldn't add it as an extra Label (if needed, so not now) ? I'll remove it for now to make things more straight forward.

PR #349, opened 2015-12-07
Add support for docker build nocache option via -Ddocker.nocache=true

stevemuk:
I was having some trouble with a corrupted docker cache so added this to the plugin.


rhuss:
Thanks for the PR. Since a `nocache` option has been added to the build config, too, I added this feature a little bit differently.

PR #366, opened 2016-01-26
Fixed documentation links

voltecrus:
Fixed one of the links in the global configuration md file, but there are still broken ones. Specifically the "Port Mapping" and "Log Configuration".


rhuss:
Thanks !

PR #367, opened 2016-01-26
Check for repository with 'https://' scheme

nakedpony:
Judging the [docs](https://docs.docker.com/engine/reference/commandline/tag/) the name of the image contains the registry host rather then the full url.
For ex.: my.super.registry.com:443/images

But in the config.json there is usually full address with the http scheme:
```json
{
  "auths": {
    "https://my.super.registry.com": {
      "auth": "my_auth",
      "email": "none"
    }
  }
}
```


rhuss:
Thanks a lot :)

Issue #374, opened 2016-02-04
Use baseDirectory from assembly descriptor if specified


kevinconaway:
When using an assembly descriptor, the docker-maven-plugin should use the `baseDirectory` element if it is specified.  Using an assembly xml file centralizes the configuration for copying files to the resulting image, it makes sense to me to be able to specify the baseDirectory in the same place. 

Additionally, this slightly cuts down on the amount of configuration needed when multiple images are using the same assembly.

rhuss:
Currently when both elements are used, the are concatenated to `${baseDirFromPluginConfig}/${baseDirFromAssembly}`, so using only `{$baseDirFromAssembly}` overwriting a given `${baseDirFromPluginConfig}` would break current configurations. 

I agree that it might have been better to use the baseDir from the assembly and set the plugin base dir to `/` (that would be a sane default). For the next major release, the default should be changed (but not for the next minor release coming shortly). 

Would this solve your use case, i.e. setting the default of `baseDir` in the plugin configuration to `/` (which btw you could set now explicitly to '/' in order to achieve the same behaviour) ?

kevinconaway:
Sorry for the delay in replying.

> Currently when both elements are used, the are concatenated to ${baseDirFromPluginConfig}/${baseDirFromAssembly}

I'm not seeing where that happens.  Looking in [DockerAssemblyManager](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyManager.java#L264), I only see the _baseDir_ from the image assembly config being used

rhuss:
One is the base directory within the assembly descriptor which is not explicitly used by d-m-p but by the assembly archiver when [creating the assembly](https://github.com/fabric8io/docker-maven-plugin/blob/8c3b34f5ef3393d42d3e185e2ac6a72a2710162b/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java#L304-L304). The `basedir` in the `Assembly` config used here is the one from the assembly descriptor and results in a top-level directory within the assembly archive. The `basedir` in the DockerAssemblyManager specifies the directory where this archive should be extracted, so effectively both directories are added (that's also true when not really an assembly is created but a directory moved around when using the `dir` mode).

kevinconaway:
[...]

PR #375, opened 2016-02-04
#374 Use assembly baseDirectory if it is provided

kevinconaway:
Fix for #374.  If a use specifies a `baseDirectory` in the assembly.xml, it should take precedence over the `baseDir` specified in the build image config

### DIFF
--- a/src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyManager.java
+++ b/src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyManager.java
@@ -71,9 +71,10 @@ public class DockerAssemblyManager {
 
         AssemblyConfiguration assemblyConfig = buildConfig.getAssemblyConfiguration();
         AssemblyMode assemblyMode = (assemblyConfig == null) ? AssemblyMode.dir : assemblyConfig.getMode();
+        Assembly assembly = null;
 
         if (hasAssemblyConfiguration(assemblyConfig)) {
-            createAssemblyArchive(assemblyConfig, params, buildDirs);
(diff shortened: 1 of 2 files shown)


rhuss:
Thanks a lot, I will have a look at it probably at the end of next week (will be off until then)

src/main/java/org/jolokia/docker/maven/assembly/DockerAssemblyManager.java
```diff
@@ -262,8 +263,13 @@ DockerFileBuilder createDockerFileBuilder(BuildImageConfiguration buildConfig, A
             builder.workdir(buildConfig.getWorkdir());
         }
         if (assemblyConfig != null) {
+            String baseDir = assemblyConfig.getBasedir();
+            if (assembly != null && assembly.getBaseDirectory() != null) {
+                baseDir = assembly.getBaseDirectory();
+            }
```
rhuss: The problem here is, if you set the *outer* basedir (the directory, that is used by d-m-p to place the assembly within a docker image) to the same as the *inner* basedir of the assembly descriptor you will get the directories, twice (e.g. if you set the assembly `baseDirectory` to `/app-dir` then your files will be placed into `/app-dir/app-dir`. That's because the `AssemblyArchiver` uses this directory, too create an extra level. 

In fact in this case the outer `baseDir` should be set to `/` to avoid this.

rhuss:
Coming back to one of d-m-p's oldest outstanding PR ;-) Are you still interested into this change and if so, it the compromise suggested is acceptable to you ?

rhuss:
Closing as beeing stale.

Issue #379, opened 2016-02-09
Enhancement: add log-driver capability to docker:run


rikcarve:
I could not find any support for using a different log-driver within docker:run.
For some containers we use the gelf log driver in docker:
`sudo docker run -d --log-driver=gelf --log-opt gelf-address=udp://localhost:12201 myimage`

Any plans to add this?

jgangemi:
at some point, yes. pull requests welcome...

rikcarve:
Ok, I've got it working now locally. After some cleanup I will make a pull request.

rhuss:
Cool. Looking forward to the PR ;-)

rikcarve:
Finally made a pull request: #387

rhuss:
Thanks ;-) 'will have a look at it ASAP

rhuss:
Merged, should be available in 0.14.0 (to be released in this week)

rikcarve:
Thanks, tested the latest release from maven central and it works as expected.

PR #387, opened 2016-02-16
379 logdriver

rikcarve:
added support for --log-driver and --log-opts in docker:start - enhancement  #379


rhuss:
Thanks a lot !

I changed the PR [a bit](https://github.com/fabric8io/docker-maven-plugin/commit/6f56395f5fb3c6771cda9cf69dc7e45af7aee84a), hope that's ok.

Will be available in 0.14.0, to be released this week.

rikcarve:
Yeah, you're right, the separate logConfig section was probably too confusing. Thanks for integrating it!

PR #390, opened 2016-02-24
fix link to the shooutout project

JohnnyQQQQ:


rhuss:
Thanks !

'knew that changing the refs to the new moved repo will screw up things a bit ;-)

PR #417, opened 2016-03-23
Creating of custom Networks during the start if they don't exist.

bohmber:
It would be nice to have some base support of creating user-defined bridge network (custom network) if they do not exist.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/StartMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StartMojo.java
@@ -46,6 +46,10 @@ public class StartMojo extends AbstractDockerMojo {
     @Parameter(property = "docker.pull.registry")
     private String pullRegistry;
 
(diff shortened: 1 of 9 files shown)


rhuss:
Sorry for being late, in fact I also considered to include the creation of a custom network. However I'm not sure whether its really the duty of the plugin. 

It's hard to keep things focussed, especially in gold rush times.

Maybe you could you elaborate a bit on your use case ? This might help to get a better feeling whether to include a (more operational) task to the plugin.

bohmber:
I was not completely happy with the code. I created a new push.
The usecase is; 
How to create the user defined network on jenkins?
Call a script (I don't like this)
Call a script inside maven build (I don't like this)
Should be automatic created by the docker-maven-plugin (I like this)

The simple bridge network could be creating during start. If you need to configure a more complex network there should be a own network goal. 

I don't like the name custom Network in the code. It's not custom it's now the recommend way instead of container linking.

PR #420, opened 2016-03-25
Creating of custom Networks during the start and delete during the stop

bohmber:
Now the user-defined networks (Docker Networking) are the precommend way instead of container linking.

### DIFF
--- a/samples/data-jolokia-demo/pom.xml
+++ b/samples/data-jolokia-demo/pom.xml
@@ -22,13 +22,13 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>dmp-sample-parent</artifactId>
-    <version>0.15.5</version>
+    <version>0.15-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
   <groupId>io.fabric8</groupId>
   <artifactId>docker-jolokia-demo</artifactId>
-  <version>0.15.5</version>
+  <version>0.15-SNAPSHOT</version>
   <!-- add custom lifecycle -->
   <packaging>docker</packaging>
 
--- a/samples/dockerignore/pom.xml
+++ b/samples/dockerignore/pom.xml
@@ -10,12 +10,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>dmp-sample-parent</artifactId>
-    <version>0.15.5</version>
+    <version>0.15-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
   <artifactId>dmp-sample-dockerignore</artifactId>
-  <version>0.15.5</version>
+  <version>0.15-SNAPSHOT</version>
   <packaging>docker-build</packaging>
 
   <build>
--- a/samples/net/pom.xml
+++ b/samples/net/pom.xml
@@ -16,12 +16,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>dmp-sample-parent</artifactId>
-    <version>0.15.5</version>
+    <version>0.15-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
   <artifactId>dmp-sample-net</artifactId>
-  <version>0.15.5</version>
+  <version>0.15-SNAPSHOT</version>
 
   <profiles>
     <profile>
--- a/samples/pom.xml
+++ b/samples/pom.xml
@@ -21,7 +21,7 @@
 
   <groupId>io.fabric8</groupId>
   <artifactId>dmp-sample-parent</artifactId>
-  <version>0.15.5</version>
+  <version>0.15-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <url>http://www.jolokia.org</url>
--- a/samples/properties/pom.xml
+++ b/samples/properties/pom.xml
@@ -10,12 +10,12 @@
   <parent>
     <groupId>io.fabric8</groupId>
     <artifactId>dmp-sample-parent</artifactId>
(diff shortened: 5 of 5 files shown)


rhuss:
Thanks, I will review it soon.

src/main/java/io/fabric8/maven/docker/access/NetworkCreateConfig.java
```diff
@@ -0,0 +1,26 @@
+package io.fabric8.maven.docker.access;
+
+import org.json.JSONObject;
+
+public class NetworkCreateConfig {
+    final JSONObject createConfig = new JSONObject();
+    final String name;
```
rhuss: wonder whether it would make sense to support more of the options as described in https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#create-a-network ?

src/main/java/io/fabric8/maven/docker/StopMojo.java
```diff
@@ -105,4 +112,27 @@ private boolean invokedTogetherWithDockerStart() {
...
+
+    private Set<Network> getUnusedCustomNetworks(QueryService queryService, PomLabel pomLabel) throws DockerAccessException {
+        if (!createCustomNetworks) {
+            return Collections.emptySet();
+        }
+        Set<Network> customNetworks = new HashSet<>();
+        for (ImageConfiguration image : getImages()) {
+            final NetworkingMode networkingMode = image.getRunConfiguration().getNetworkingMode();
+            if (networkingMode.isCustomNetwork()) {
+                Network network = queryService.getNetworkByName(networkingMode.getCustomNetwork());
```
rhuss: Probably a `Set<Network> customNetworks = queryService.getNetworksByName()` is better here since you then need only a single HTTP request to the Docker daemon instead of one for each image.

src/main/java/io/fabric8/maven/docker/StopMojo.java
```diff
@@ -105,4 +112,27 @@ private boolean invokedTogetherWithDockerStart() {
         Boolean startCalled = (Boolean) getPluginContext().get(CONTEXT_KEY_START_CALLED);
         return startCalled != null && startCalled;
     }
+
+    private Set<Network> getUnusedCustomNetworks(QueryService queryService, PomLabel pomLabel) throws DockerAccessException {
```
rhuss: Maybe `getNetworksToRemove()` is a better name here.

src/main/java/io/fabric8/maven/docker/StopMojo.java
```diff
@@ -105,4 +112,27 @@ private boolean invokedTogetherWithDockerStart() {
...
+                        if (!shouldStopContainer(container, pomLabel)) {
+                            // it's sill in use don't collect it
+                            customNetworks.remove(network);
+                        }
+                    }
+                }
+            }
+        }
+        return customNetworks;
+    }
```
rhuss: What's about the following situation:

* User creates custom network outside of Maven 
* User references this custom network in d-m-p image configs
* User calls 'mvn docker:stop' (without 'docker:start' in the same run but maybe in an extra call before)

At the end I guess the user probably expects the custom network to still exist. But if I understand the code correctly the network get removed nevertheless. 

For the container we had the same problem and could solve it by putting a d-m-p specific label on the container that were created by d-m-p itself. 

I wonder how we could solve the situation here. One could be to prefix all networks created by dmp with e.g. `dmp-` transparently so that it could be detected whether the network was created by the plugin or outside. 

wdyt ?

rhuss: Maybe one could use the `options` field in `NetworkCreateConfig` to pass in a tag for dmp. This would work if the network driver are supposed to ignore unknown options (but thats not documented and probably need to be tried out).

rhuss:
I know we don't have much integration tests (yet), however a simple example in `samples/custom-network` would help people understand the concepts.

what would be also awesome: a documentation update in `doc/manual/docker-start.md` and `doc/manual/docker-stop.md` ;-)

bohmber:
I was really busy. Hopefully i had tonight or tomorrow time we working on your comments.
Sorry

benjefferies:
This is an awesome addition, we have been looking around for something to do this and had to settle with making a http request to create a network via another plugin. It would be great if this could make the next release.

rhuss:
@bohmer ping ;-) Any news or comments on your PR ?

bohmber:
Sorry I was really busy. This weekend is the first weekend without any marriage, bachelor party or anniversary. I try to update the PR until friday. I hope I have some time at the weekend or in the evening during the week.

rhuss:
hey, no problem at all ;-) There are things which are much more important :) 

If its too much, please let me know, we could it pick up then. But please don't feel thronged.

src/main/java/io/fabric8/maven/docker/StopMojo.java
bohmber: ok

src/main/java/io/fabric8/maven/docker/StopMojo.java
bohmber: I think changing the default of creating customNetwork to false is the safest option

src/main/java/io/fabric8/maven/docker/access/NetworkCreateConfig.java
bohmber: At the moment i have no usecase for using options for networks. We could extend the NetworkCreateConfig if we support a network goal.

src/main/java/io/fabric8/maven/docker/StopMojo.java
bohmber: ok

Issue #264, opened 2015-08-18
Plans for docker-machine support?


fastnsilver:
Docker is deprecating `boot2docker`. See message when you `boot2docker up` with v1.8.0.

What plans do you have for supporting `docker-machine`?  See https://www.docker.com/docker-machine and https://docs.docker.com/machine/get-started/.

jgangemi:
i'm not sure the two are related here. you would use `docker-machine` to achieve the same thing that the boot2docker setup was doing for you, set up a `VirtualBox` instance that will run `docker-engine` that you can then deploy containers to.

you would then configure the `docker-maven-plugin` to point to that `VirtualBox` host, similar to how `boot2docker` works now, and run the plugin accordingly.

could you describe a use-case you see the plugin fulfilling for `docker-machine`?

fastnsilver:
Thanks for the quick reply. It's just the out of the box experience I'm bemoaning here.  If someone (like me ;) ) walks up to use this plugin, what's the minimum config necessary to achieve the same results with a `Dockerfile` and `docker build...` via `docker-machine`?  

My project does not employ a `Dockerfile` rather the config as per this plugin (though I know the former is an option).  I did NOT know I could point the `docker-maven-plugin` at my VBox image. (How and where would I do this? properties?)  

I'll attempt to illustrate what lead me to write-up this issue...

When I had uninstalled `boot2docker` and installed `docker-machine` the `docker-maven-plugin` `docker:build` goal complained about a missing `boot2docker`. With `boot2docker` installed this Maven plugin works without any further need of config (it knows where to source the VM hosting Docker).

rhuss:
By default the plugin uses the environment variable `DOCKER_HOST` which should be set externally like with `eval $(boot2docker shellinit)` or `$(docker-machine env default)` (when `default` it the docker machine's name). The plugin has been tested with a docker-machine setup and it works well.

So, the plugin needs the same setup like the docker CLI (regardless whether you are using docker-machine or boot2docker). BTW, there is no direct reference at all to boot2docker  from within the plugin, so I wonder what is complaining about boot2docker in your case. Could you lease post some logs ?

I agree with you that a 'self-contained' is a good goal, but even when you are using docker-machine you still need to install docker-machine itself and setup these environment variables for you with the commands given above.

And I also agree with @jgangemi that calling docker-machine from within the plugin is out of scope ;-)

jgangemi:
[...]

PR #424, opened 2016-04-04
#264, update references to boot2docker to include docker-machine

stagrlee:
The documentation referring to the deprecated boot2docker software makes the plugin sound dated even though it works fine with docker-machine.  I think this pull request addresses the intent of issue #264


rhuss:
Thanks a lot, make much sense !

PR #431, opened 2016-04-19
Add an env goal

chonton:
Use 'docker-machine env' to set the java process environment variables.  This allows less environmental setup for jenkins.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/MachineMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/MachineMojo.java
@@ -1,0 +1,174 @@
+package io.fabric8.maven.docker;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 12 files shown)


rhuss:
thanks will check this out for the next release (which will come sooner than the last one)

chonton:
We will also want to add to Docker lifecycle.  Probably bind to _validate_ or _initialize_ phase.

src/main/java/io/fabric8/maven/docker/MachineMojo.java
```diff
@@ -0,0 +1,174 @@
...
+        try {
+            outputStream.close();
+        } catch (IOException e) {
+            log.info("failed to close docker-machine output stream: " + e.getMessage());
+        }
+    }
+
+    private Process startDockerMachineProcess() throws MojoExecutionException {
+        try {
+            return Runtime.getRuntime().exec(new String[] { "docker-machine", "env", "--shell", "cmd" });
```
rhuss: This wont work if you have more than one docker machine installed.

rhuss:
I wonder whether we really need a new goal (which is always something 'heavy weight'), but maybe a proeprty instead which gets evaluated in the base Mojo for adding to the properties. Or even, as an automatic fallback when not DOCKER_HOST is set ? This would make things much more transparent for people.

rhuss:
Maybe we can add it the evaluation of docker-machine as fallback here --> https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java#L25-L33 (should be checked whether docker-machine is installed first, though)

src/main/java/io/fabric8/maven/docker/MachineMojo.java
chonton: Warn and choose one at random or fail?

chonton:
I like the idea of this being a fallback mechanism.  I'll look at this tonight or tomorrow.

chonton:
Updating the pull request per your comments.

PR #433, opened 2016-04-24
Docker lifecycle

chonton:
Add docker packaging.  This eliminates much boilerplate configuration.

### DIFF
--- a/src/main/resources/META-INF/plexus/components.xml
+++ b/src/main/resources/META-INF/plexus/components.xml
@@ -1,0 +1,29 @@
+<component-set>
+ <components>
+  <component>
+   <role>org.apache.maven.lifecycle.mapping.LifecycleMapping</role>
+   <role-hint>docker</role-hint>
+   <implementation>org.apache.maven.lifecycle.mapping.DefaultLifecycleMapping</implementation>
+   <configuration>
+    <lifecycles>
+     <lifecycle>
+      <id>default</id>
+      <phases>
+       <process-resources>org.apache.maven.plugins:maven-resources-plugin:2.7:resources</process-resources>
+       <compile>org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile</compile>
+       <process-test-resources>org.apache.maven.plugins:maven-resources-plugin:2.7:testResources</process-test-resources>
+       <test-compile>org.apache.maven.plugins:maven-compiler-plugin:3.5.1:testCompile</test-compile>
+       <test>org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test</test>
+       <package>org.apache.maven.plugins:maven-jar-plugin:2.6:jar,${project.groupId}:${project.artifactId}:${project.version}:build</package>
+       <pre-integration-test>${project.groupId}:${project.artifactId}:${project.version}:start</pre-integration-test>
+       <integration-test>org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:integration-test</integration-test>
+       <post-integration-test></post-integration-test>
+       <verify>${project.groupId}:${project.artifactId}:${project.version}:stop,org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:verify</verify>
+       <deploy>${project.groupId}:${project.artifactId}:${project.version}:push</deploy>
+      </phases>
+     </lifecycle>
+    </lifecycles>
+   </configuration>
+  </component>
+ </components>
+</component-set>
--- a/src/main/resources/META-INF/plexus/components.xml
+++ b/src/main/resources/META-INF/plexus/components.xml
@@ -8,7 +8,12 @@
     <lifecycles>
      <lifecycle>
       <id>default</id>
+       <!--
+       phase mappings extend from jar mappings
+       http://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Default_Lifecycle_Bindings_-_Packaging_ejb__ejb3__jar__par__rar__war
+       -->
       <phases>
+       <generate-source>${project.groupId}:${project.artifactId}:${project.version}:source</generate-source>
        <process-resources>org.apache.maven.plugins:maven-resources-plugin:2.7:resources</process-resources>
        <compile>org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile</compile>
        <process-test-resources>org.apache.maven.plugins:maven-resources-plugin:2.7:testResources</process-test-resources>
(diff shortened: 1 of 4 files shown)


chonton:
cargo-jolokia-demo does not work on my local host.  Master branch has same problem, so the cause is probably not this changeset.

rhuss:
Thanks a lot ! I will check it out next week

src/main/resources/META-INF/plexus/components.xml
```diff
@@ -17,6 +22,7 @@
        <package>org.apache.maven.plugins:maven-jar-plugin:2.6:jar,${project.groupId}:${project.artifactId}:${project.version}:build</package>
        <pre-integration-test>${project.groupId}:${project.artifactId}:${project.version}:start</pre-integration-test>
        <integration-test>org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:integration-test</integration-test>
+       <!-- post-integration-test is opportunity for code coverage, such as jacoco, to collect statistics from running container -->
        <post-integration-test></post-integration-test>
        <verify>${project.groupId}:${project.artifactId}:${project.version}:stop,org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:verify</verify>
```
rhuss: Any particular reason why "stop" is not in `<post-integration-test>` ?

I ask because if you only run `mvn integration-test` (without `verify`) a server wont be stopped.

src/main/resources/META-INF/plexus/components.xml
```diff
@@ -17,6 +22,7 @@
        <package>org.apache.maven.plugins:maven-jar-plugin:2.6:jar,${project.groupId}:${project.artifactId}:${project.version}:build</package>
        <pre-integration-test>${project.groupId}:${project.artifactId}:${project.version}:start</pre-integration-test>
        <integration-test>org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:integration-test</integration-test>
+       <!-- post-integration-test is opportunity for code coverage, such as jacoco, to collect statistics from running container -->
```
rhuss: Ok, got your intention to leave the hook free for others to kick in. However, putting it into `verify` has the nasty side effect that `mvn integration-test` wont call this phase so the container keeps running. It also *feels* more natural to put it into `pre` and `post` goals. Couldn't a user still bind to the `post-integration-test` phase so that both goals are called ?

src/main/resources/META-INF/plexus/components.xml
rhuss: Ok, just checked it: any extra post-integration binding goes after the default one (strange behaviour btw, would expect for `pre` that custom executions goes after the default and for `post` before).

But if you need a customization you can still bind e.g. jacoco to the `integration-test` phase, which then will called *after* the integration test but before the `post-integration-phase`.

I adapted the lifecycle accordingly and also removed `docker:source` since this is duplicated in `docker:build`, too (adding to the build time)

chonton: It's documented that order of plugins is [those specified by packaging](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Plugins), followed by [order declared in pom](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#A_Build_Phase_is_Made_Up_of_Plugin_Goals).

[Maven Failsafe](https://maven.apache.org/surefire/maven-failsafe-plugin/), usually binds with _integration-test_ and _verify_.  [Jacoco report-integration-mojo](http://eclemma.org/jacoco/trunk/doc/report-integration-mojo.html) likewise binds to _verify_.  So, current practice is that running and actually verifying integration tests requires using _verify_ phase.

I was guided in my choices by what other container technologies use; [Cargo](https://codehaus-cargo.github.io/cargo/Maven2+Plugin+Tips.html) recommends associating start goal with _pre-integration-test_ and stop goal with _post-integration-test_. [Tomcat maven plugin](http://tomcat.apache.org/maven-plugin-2.2/run-mojo-features.html) recommends associating run-war-only goal with _pre-integration-test_ and shutdown goal with _post-integration-test_.

I would prefer not to bind [jacoco:dump](http://eclemma.org/jacoco/trunk/doc/dump-mojo.html) with integration-test.  This goal must be called after failsafe runs the tests and before the container is stopped.  It's often difficult to ensure the ordering of goals within a phase, particularly when parent poms and profiles are involved.  You already understand the problem with binding to post-integration-test.

rhuss: > I was guided in my choices by what other container technologies use; Cargo recommends associating start goal with pre-integration-test and stop goal with post-integration-test. Tomcat maven plugin recommends associating run-war-only goal with pre-integration-test and shutdown goal with post-integration-test.

That is exactly now what I suggest, too.`docker:start` in *pre-integration-test* and `docker:stop` in *post-integration-test*. My problem to moving `docker:stop` to *verify* is that you can easily call `mvn integration-test` which won't run *verify* and hence doesnt stop the container.

I know there wont be a perfect solution for the lifecycle choice, but as an alternative you could use a [preStop](http://fabric8io.github.io/docker-maven-plugin/docker-start.html#wait-during-startup-and-shutdown) command to get data out of the container before it stops. Or you mount the the directory where jacoco dumps its data as a volume dir within `target`. 

wdyt, would this be ok ?

Btw, I introduced two other packagings, too: 'docker-build' for only building images and 'docker-tar' for creating a docker tar as artifact (Dockerfile + support files).

Thanks btw for the nice PR, makes things quite simpler :)

chonton: Even after moving `docker:stop` to _post-integration-test_, calling `mvn integration-test` won't run _post-integration-test_ and therefore not stop the container.

Unfortunately, mounting the directory is a non-starter in our build chain. Let me look at preStop tonight.

rhuss: > Even after moving docker:stop to post-integration-test, calling mvn integration-test won't run post-integration-test and therefore not stop the container.

Yes, you are right, one would have to call `mvn post-integration-test` (totally intuitive ;-). Just out of curiosity : How do extract the jacoco data from a running container ?

chonton: We use [jacoco:dump](http://eclemma.org/jacoco/trunk/doc/dump-mojo.html) to connect to port 6300.

Reviewing my personal usage of phases: mostly it's _mvn clean install_. For jenkins it's _mvn clean deploy_.  If I want to build without releasing to my local respository it's _mvn clean verify_.   I use _verify_ because I want to look at the jacoco and failsafe reports generated from the integration tests.  Occasionally while debugging an aspect of the build I might use _mvn package_.

I almost never use any of the really long named phases;  they take too long to type and I usually mis-remember the actual name.  I suspect that these phases are more about sequencing than about the expectation of use from the command line.  In the [Maven Introduction to the Build Lifecycle](http://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Build_Lifecycle_Basics), you don't see any of the _pre-*_, _post-*_, or _process-*_ phases mentioned until the reference section.

rhuss: ok, you convinced me ;-). I will put `docker:stop` to *verify* phase. We can still tune this if we get more feedback later on (but I want to make a release today).

The last question is whether `docker:stop` should be before `failsafe:verify` or after ? I guess before, because `failsafe:verify` probably stops the build in case of an error so that `docker:stop` wouldnt run anymore ....

Will revert to you original suggetion.

thanks :)

chonton: Thanks for working through this with me.  I appreciate the time you've taken.  I'm now working with maven team to [update lifecycle documentation](https://github.com/apache/maven-site/pull/4).

PR #440, opened 2016-05-10
Adding general user command

stevenbower:
Just allowing for changing user during assembly creation does not allow for commands to have executed prior to using the "USER" tag.. In my use case I was calling "adduser <user>" in a RUN command but since the RUN commands don't execute until after the assembly is unpacked if you need to have files owned by a specific user you'd need to have mutliple containers and layer this in..

this patch adds a generalized "user" tag which goes at the end of the Dockerfile.


rhuss:
Integrated the patch into branch `integration`. Thx!

stevenbower:
thanks!

PR #443, opened 2016-05-11
Select always the last volume part after splitting at the colon.

nicolaimainiero:
On windows you may end with the following volume declaration:
c:\this\is\my\path:/data

When splitted on : and the second part is selected we end with \this\is\my\path which won't work.
By selecting the last part, we get the correct /data


rhuss:
Thanks ! I merged it to the integration branch will be available in the next release.

Issue #413, opened 2016-03-21
apply 'docker.skipTags' to docker:push


jgangemi:
i'd like the option of being able to not push tags to the registry when doing a `docker:push`. i'd like to just change the current `push` goal to support the `docker.skipTags` property and if that's enabled, it won't push any images.

thoughts on just re-using that property vs adding something completely new?

rhuss:
I think reusing the tag makes sense. Since its currently part of the `BuildMojo` the property should be moved up one level so that it can be used by the `PushMojo`, too.

jgangemi:
cool - i'll get a pr put together for this over the weekend.

rhuss:
Would be good if we would have some semantics to enable also the possibility to skip certain tags.

See https://github.com/fabric8io/docker-maven-plugin/issues/449#issuecomment-219802485 for a suggestion.

rhuss:
@jgangemi short heads up: I just renamed 'docker.skipTags` to `docker.skip.tag` in order to align with `docker.skip.build`, `docker.skip.run` and `docker.skip.push` which I just have introduced ;-)

PR #450, opened 2016-05-17
Fixes #449,

### LINKED ISSUE #449: Make it possible to skip push of 'latest' image.
Our registry-implementation does not allow 'moving tags' like 'latest'. So every build with configured push will fail with the current implementation of 'docker-maven-plugin'. Introducing a boolean property 'docker.push.skipLatestTag' should do the trick.

I will provide an PR.

Thanks

jbellmann:
... makes it possible to skip push of 'latest' tag to avoid 'moving-tags'.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -137,6 +137,11 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
     @Parameter
     private List<ImageConfiguration> images;
 
+    // to skip pushing images with tag 'latest', if only explicit tagged images
+    // are allowed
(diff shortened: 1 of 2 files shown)


src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -137,6 +137,11 @@
     @Parameter
     private List<ImageConfiguration> images;
 
+    // to skip pushing images with tag 'latest', if only explicit tagged images
+    // are allowed
+    @Parameter(property = "docker.skipPushLatestTag", defaultValue = "false")
```
jgangemi: this should use `docker.skipTags` property which already exists, see #413

rhuss:
We need to revise this and implement it differently for #413. So I'll close this PR.

PR #456, opened 2016-05-21
Add log at http wait beginning

reda-alaoui:


rhuss:
Meged into branch `integration`. Thanks !

PR #464, opened 2016-06-02
Add ShmSize support for "docker run" command

obigard:
Need to add "--shm-size" property for "docker run" command


rhuss:
awesome !

Two things:

* Could you please add to the documentation in https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/manual/docker-start.md (look for 'Memory') ?
* Could you please add the support also to the property configuration handler like in https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/config/handler/property/PropertyConfigHandler.java#L104 ?

would be awesome if you could add this, too. btw, we are planning a new release of d-m-p tomorrow or so.

obigard:
I'll do it...

obigard:
I've got a question: when implementing this "shm-size" enhancement, I checked the JSon data that should be sent to Docker Remote API. And it seems that the "Memory" and "MemorySwap" should be placed under "HostConfig" tag, which doesn't seem to be the case in your "ContainerCreateConfig"...
Am I wrong? Shouldn't tags "memory" and "memorySwap" be placed in "ContainerHostConfig" class instead of "ContainerCreateConfig"?

Thanks

rhuss:
damned, you are right. wonder why nobody noticed that yet. Guess the tests could be improved. 

thanks for the heads up I will fix this while I'm merging in your PR.

Thanks a lot !

obigard:
Do you know why the Jenkins check is failing???

rhuss:
looks like some internal issue with the git merge plugin. will check that later. but 2 out of 3 are good enough ;)

obigard:
ah, ok :-)

PR #472, opened 2016-06-06
fixes #471

### LINKED ISSUE #471: add support for .maven-dockerincludes file to act as the reverse of .maven-dockerignore

jstrachan:


rhuss:
Let me have a look, maybe we could even integrated this also into the XML config. At least we should add `.mvn-dockerexclude` as synonym for `.mvn-dockerignore` (as physicist I'm always after symmetry :)

Or we add a general '.mvn-docker' which holds include *and* exclude patterns.

What is your use case for this ? At the end we could always use a dedicated directory which is safe to use as the root of a docker.tar without includes or excludes.

rhuss:
Ah, no. Again the bot which merges on each comment. 'have to fix the regexp ...

jstrachan:
@rhuss agree on the includes/excludes names btw. 

My main use case was for the nodejs example:
https://github.com/fabric8-quickstarts/funktion-nodejs-example

its much easier and safer to just list what to include; rather than guessing all the stuff to exclude. e.g. for nodejs its much easier.  The `.maven-dockerinclude` is:
```
Dockerfile
**/*.js
**/*.json
```
Whereas IDEs, operating systems and build systems tend to generate all kinds of stuff in the file system that you'd normally have to list and exclude.

Issue #412, opened 2016-03-21
New feature request: Add support for a containerId property


niriak:
Please add a configuration option for the goal _docker:start_ to provide a maven project property name the containerId is stored into. I'd like to run some integration tests using docker-java client with the container setup done by the docker-maven-plugin.

E.g. in the configurations's run section a new tag _containerIdProperty_:
```
<run>
   <containerIdProperty>containerId</containerIdProperty>
</run>
```

The maven project property _containerId_ will contain the id of the new generated container. This property can now be used in the maven-failsafe-plugin to pass the containerId to Java.

rhuss:
IMO that makes sense, maybe we can make it a bit more generic, too (don't like long property names so much ;-)

My suggestion would be 

```xml
<run>
   <defineProperties>
       <containerId>myPropertyName</containerId>
       <!-- Maybe later other meta data like the "name" or "startDate", stuff which 
             is available from "docker inspect" -->
   </defineProperties>
</run>
```

We could also implicitly set properties after some conventions like `docker.container.<alias | image name>.id` automatically. I'm not so much a big fan of this because it couples the alias/image name to the code where it is used implicitly, so changing an name/alias would need to update the code, too but there is no direct reference whether the implicitly used property is ever used.

wdyt ?

niriak:
The variant with the defineProperties tag sounds good to me.

niriak:
Please add a property "containderId" to the list of specific properties which are exposed as Maven properties. E.g.: "docker.container.tomcat.containerId".

See also https://fabric8io.github.io/docker-maven-plugin/docker-start.html

rhuss:
Sorry, forgot this issue. will be available in the next release.

niriak:
Ups, got your answer to late. Already send a pull request.
Sorry, I'm a newbie - how to proceed now?

rhuss:
No problem. The release is underway, you will be able to use 0.15.5 in about half an hour.

niriak:
Thanks a lot!

PR #474, opened 2016-06-07
Adds containerId property

niriak:
#412 Change-Id: I9a490127734bd01128eeb3e9c2ea0980f1669c7f


rhuss:
Already added otherwise. Thanks a lot, though !

PR #482, opened 2016-06-13
Creating of custom Networks during the start and delete during the stop

bohmber:
Default of createCustomNetworks is now false. No issues with existing networks.
Apply your small suggestions.

I need some advise how to create some unit tests


rhuss:
thanks a lot ! I'm at conferences this week, but will have a look at the PR soon when I'm back (or maybe if I have some spare time then even this week ;-)

rhuss:
Thanks for the PR. I polished it a bit and merged it to the `integration` branch so that it will end up in the next release.

/wrt unit test: They are quite a bit hard to write, so I suggest that you create a small sample project in `samples/` which show the usage of a custom network. Maybe a simple java service calling a mysql db or sth. like that. That would be awesome. Please open a new PR for this since I'm going to close this.

Thank a lot !

PR #483, opened 2016-06-14
Skip push

bohmber:
Allow skip push via global parameter like skipBuild and skipTags and skipRun. The docker.skip parameter doesn't really fits into my use case without some configuration nightmare.


rhuss:
Thanks ! I'm at a conference this week, but will have a look early next week.

rhuss:
Thanks, I added this and also skip hooks for build & run. The properties are now called

* docker.skip.build
* docker.skip.run
* docker.skip.push

PR #487, opened 2016-06-17
docker container fetch limit is now configurable

greggulrajani:
Hello,

We are using your excellent plugin to deploy pushed code to a docker server. But when we have > 100 containers running we have been having issues restarting containers that are older than the 99st container.
I am happy to add any unit tests or make any changes ...

Thanks again for the great plugin...

### DIFF
--- a/doc/manual/external-configuration.md
+++ b/doc/manual/external-configuration.md
@@ -36,8 +36,8 @@ Example:
 </image>
 ```
 
-Given this example configuration a single image configuration is build
-up from the following properties, which correspond to corresponding
+Given this example configuration a single image configuration is built
+up from the following properties, which correspond to the corresponding
 values in the `<build>` and `<run>` sections.
 
 * **docker.alias** Alias name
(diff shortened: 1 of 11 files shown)


greggulrajani:
Another option which IMO seems preferable  is to use
    GET /containers/json?all=1HTTP/1.1

which will return all running containers with no upper limit.

rhuss:
Looks good to me. Some remarks:

* Could you please add the new configuration to the documentation, too (its within `doc/manual`) ? 
* We also have an  [PropertyConfigHandler](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/config/handler/property/PropertyConfigHandler.java) where this option would need to be added, too (as well in the doc for *external configuration*)
* Allowing for now limit is also a good idea. Maybe we can use a value of `0` to specify no limit ?

Thanks a lot !

greggulrajani:
Hello Rhuss,

I have created another pull request that is on the 'integration' branch.
Please refer to that for the changes you requested.

-best
-greg

PR #490, opened 2016-06-21
lets add a warning if the watch goal finds no files to watch; it could be the user didn't invoke `package`

jstrachan:
lets add a warning if the watch goal finds no files to watch; it could be the user didn't invoke `package`

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -237,6 +237,18 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
         }
     }
 
+    protected void clearResolvedImages() {
+        Map ctx = getPluginContext();
+        ctx.remove(CONTEXT_KEY_RESOLVED_IMAGES);
+    }
+
+    protected List<ImageConfiguration> recreateResolvedImages() {
+        clearResolvedImages();
+        initImageConfiguration();
+        return getResolvedImages();
+    }
+
+
     private DockerAccess createDockerAccess(String minimalVersion) throws MojoExecutionException, MojoFailureException {
         DockerAccess access = null;
         if (isDockerAccessRequired()) {
--- a/src/main/java/io/fabric8/maven/docker/WatchMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/WatchMojo.java
@@ -228,7 +228,7 @@ public class WatchMojo extends AbstractBuildSupportMojo {
         };
     }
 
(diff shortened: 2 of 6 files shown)


src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -237,6 +237,18 @@ private void storeInPluginContext(List<ImageConfiguration> resolvedImages, Strin
         }
     }
 
+    protected void clearResolvedImages() {
+        Map ctx = getPluginContext();
+        ctx.remove(CONTEXT_KEY_RESOLVED_IMAGES);
+    }
+
+    protected List<ImageConfiguration> recreateResolvedImages() {
```
rhuss: This method is currently not used. What for do you need it ?

jstrachan: its used in fabric8:watch ;)

it will be used in docker:watch if we decide to push up the changes there too to deal with timestamps

src/main/java/io/fabric8/maven/docker/WatchMojo.java
```diff
@@ -176,6 +176,9 @@ private Runnable createBuildWatchTask(final ServiceHub hub, final ImageWatcher w
             throws MojoExecutionException {
         final ImageConfiguration imageConfig = watcher.getImageConfiguration();
         final AssemblyFiles files = hub.getArchiveService().getAssemblyFiles(imageConfig, mojoParameters);
+        if (files.isEmpty()) {
+            log.error("No assembly files for " + imageConfig + ". Are you sure you invoked this goal after the `package` goal?");
+        }
```
rhuss: @jstrachan what do you think instead of adding a warning, to introduce to fork the Maven lifecycle for this goal so that package is automatically included and introduce a second `build-nofork` goal which can be used for binding in `<executions>` (where lifecycle forking is evil) ?

That the same way how e.g. the maven-source-plugin is designed.

rhuss: Problem would be that this change is not backwards compatible for those already binding `docker:build` to a phase. So I probably postpone this until the 0.16.x release and just leave the error here.

Do you think we should even throw an exception here because its an error ? If its only a warning, then maybe we should also use only a warning message.

rhuss: Oops, missed that it was on the WatchMojo (not BuildMojo). So maybe we could simply use a forked lifecycle for watch ? So that we can simply call `docker:watch` instead of `package docker:build docker:watch` ? (package and build would be called implicitely). 

Since `docker:watch` is not supposed to be bound to an execution phase there should be no problem with forking.

wdyt ?

Issue #448, opened 2016-05-17
docker-maven-plugin re-launch other plugin


burtsevyg:
Please see source pom.xml https://github.com/burtsevyg/test/blob/master/pom.xml for example

in log I see:
First time:

```
...
[INFO] --- maven-antrun-plugin:1.8:run (ls-command) @ docker-maven-plugin-test ---
[INFO] Executing tasks
...

```

Second time:

```
...
[INFO] >>> docker-maven-plugin:0.15.1:source (start) @ docker-maven-plugin-test >>>
[INFO]
...
[INFO] --- maven-antrun-plugin:1.8:run (ls-command) @ docker-maven-plugin-test ---
[INFO] Executing tasks
...
[INFO] <<< docker-maven-plugin:0.15.1:source (start) @ docker-maven-plugin-test <<<
```

rhuss:
Yes, that's true and because of that `docker:source` forks the Maven lifecycle in order to create the artefacts (jars, copied resources) to include in the `docker.tar`

Hence it is probably not well suited to add as phase to the lifecycle where the the `package` phase is already called (hence it will run twice through the package phase, one because of the norma lifecycle and one for the callout by `docker:source`).

There is no easy solution for this if one want to guarantee that the *package* phase is called when running `mvn docker:start` directly. 

Maybe we should use the same 'trick' like the [maven-source-plugin](http://svn.apache.org/viewvc/maven/plugins/tags/maven-source-plugin-3.0.0/src/main/java/org/apache/maven/plugins/source/), e.g. having two goals `source:jar` and `source:jar-no-fork`, one for when called directly and the other when bound to an execution.

I will look into this and probably mimic this behaviour if I cant find a more elegant solution.

PR #505, opened 2016-07-06
- invoke the 'initialize' phase before goals (resolves #315)

### LINKED ISSUE #315: invoking lifecycles before direct goals
i know this has been a discussion before...did we ever come to any resolution on this, i can't remember.

if we didn't, guess what i just found/realized! the `source:jar` goal invokes the `generate-sources` lifecycle phase before it execute itself, which is what i want the `docker:*` commands to do. now i can go digging through that code to see how it's done.

if we did figure something out, `sheepish grin` for not remembering and having to ask what we came up w/. :)

jgangemi:
adds `@Execute(phase = LifecyclePhase.INITIALIZE)` to all the docker goals w/ the exception of `source`, which gets this behavior by default b/c it calls `LifecyclePhase.PACKAGE`).

it also looks like the `changelog` has a bunch of whitespace errors that the `any edit tools` plugin for eclipse has fixed (it runs automatically on editor save). if that annoys you, let me know and i'll roll it back.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,15 +2,16 @@
 
 * **0.15-SNAPSHOT**
   - Dont do redirect when waiting on an HTTP port (#499)
+  - Invoke the `initialize` phase before docker goals (#315)
 
 * **0.15.9** (2016-06-28)
   - Fixed issue when target directory does not exist yet ([#497](https://github.com/fabric8io/docker-maven-plugin/issues/497))
-  
+
 * **0.15.8** (2016-06-27)
   - Removed image configuration caching ([#495](https://github.com/fabric8io/docker-maven-plugin/issues/495))
   - Fix for tcp wait when used with Docker for Mac ([#430](https://github.com/fabric8io/docker-maven-plugin/issues/430))
   - Add warning when assembly is empty when watching a Docker image ([#490](https://github.com/fabric8io/docker-maven-plugin/issues/490))
-  - Add `docker.skip.build`, `docker.skip.run`, `docker.skip.push` properties and 
+  - Add `docker.skip.build`, `docker.skip.run`, `docker.skip.push` properties and
     renamed `docker.skipTags` to `docker.skip.tag` ([#483](https://github.com/fabric8io/docker-maven-plugin/issues/483))
   - Reverted jansi back to version 1.11 because of [this issue](https://github.com/fusesource/jansi/issues/58)
   - Add new assembly config options `permissions` for fine tuning permissions in the docker.tar ([#477](https://github.com/fabric8io/docker-maven-plugin/issues/477)). Deprecated `ignorePermissions`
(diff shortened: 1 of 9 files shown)


rhuss:
Could you please elaborate on why you need to fork the lifecycle for every goal to call this phase ? What's your use case ? Remember, that the `validate` and `initialize` phase will be called for *every* Mojo Goal, even when called because its bound to an exectuion phase, so that this phase happens to be called multiple times in this case. E.g. when you bind `docker:build` and `docker:push` to the lifecycle viae executions, then a 'mvn deploy' will call both goals, so you end up calling the `validate` and `initialize` phase 3 times (one for `deploy` itself, and twice for the docker goals).

Also, by default `validate` and `initialize` phase have no goals bound to them in the default lifecycle, so this would be a noop.

`docker:source` does a fork of the lifecycle because it is not supposed to be used in executions. If we want to have this goal to be used in executions we could introduce a `docker:source-nofork`. This is the strategy e.g. the maven-source-plugin uses, too.

Forking the lifecycle is a double-edged sword, because of causing trouble with goals running multiple times (e.g. when calling a goal multiple times on the CLI like in `mvn clean package docker:build` or when binding to an execution phase as described above).

If you want certain phases to be called you can be the docker goals you need to a phase, too, and then also call this phase.

Another trick I used was to use profiles with defaults goals. See [here](https://github.com/fabric8io/ipaas-quickstarts/blob/master/quickstart/cdi/camel/pom.xml#L307-L333) for an example. This also guarantees that all goals are called in order.

rhuss:
See also See https://github.com/fabric8io/docker-maven-plugin/issues/448 for a real world issue of this.

jgangemi:
i thought my use case was pretty clear in #315. 

the primary reason is i want to use the `properties` plugin to put properties into an external file and then be able to call the individual `docker:*` goals w/o the need to either use a profile or include `initialize` on the command line.

both are slightly cumbersome and require you to remember that you need to do it, which really isn't what you want when you're trying to roll this out to a team of developers, many of whom are junior and/or afraid of maven for whatever reason. 

if `initialize` is a no-op in the first place, then this should be completely safe to do and it should be on me or whomever else wants to leverage this functionality to know to not do anything dumb.

jgangemi:
also just realized this is against the wrong branch...let me know what we're doing about this PR in general and then i can resubmit or it can get merged another way.

rhuss:
@jgangemi agree, when we fork only standard no-ops, so think it safe to add them. I'm still thinking about goals 'build' and 'build-nofork' (same for 'source' and 'source-nofork') to allow a `mvn docker:build` to rebuild the application before (currently one are forced to run `mvn packag docker:build` when using an assembly descriptor because otherwise artefacts are not added to the reactor). I hesistate a bit, because if people have bound `docker:build` as executions they would need change that to `docker:build-nofork`, otherwise e.g. `package` would run twice. 

BTW, for you its not a problem that when you bind the d-m-p goals to executions, that your actions bound to `initialize` are run twice then ? 

If not, I'm ok with this PR as it is as you said a no-op for most of us ;-)

jgangemi:
i'm fine w/ things running twice - my primary use case for wanting this functionality is so i can externalize things into property files and it won't make any difference if those get re-loaded.

do you want me to re-open this against the `integration` branch?

rhuss:
Will integrate this tomorrow ...

rhuss:
Merged into `integration`

PR #509, opened 2016-07-08
minor doc update

jgangemi:
removed duplicate parameters i noticed while looking for something else.


jgangemi:
looks like this got fixed already

PR #512, opened 2016-07-11
- add support for pulling an image once per (reactor) build

### LINKED ISSUE #504: add option so reactor builds only pull an image once
we have a rather large reactor build at the day job that uses the same container(s) over and over for integration testing. we have `autoPull` set to `always` to ensure we always get the latest container when the build starts but then each module in the build pulls the container over and over again which is somewhat needless and increases an already lengthy build time.

an additional option should be added (i'm open to naming) that will allow reactor projects to skip pulling images if it has been pulled by a previous module.

overloading `autoPull` w/ some additional values is also another option although i feel it could be confusing. i am happy to go w/ whatever the consensus is.

jgangemi:
resolves #504 and allows images to be auto-pulled once for an entire build.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -3,15 +3,17 @@
 * **0.15-SNAPSHOT**
   - Dont do redirect when waiting on an HTTP port (#499)
   - Removed the container fetch limit of 100 and optimized getting containers by name and image (#513)
-  
+  - Allow images to only be pulled once per build (useful for reactor projects) ([#504](https://github.com/fabric8io/docker-maven-plugin/issues/504))
+
+>>>>>>> Stashed changes
 * **0.15.9** (2016-06-28)
   - Fixed issue when target directory does not exist yet ([#497](https://github.com/fabric8io/docker-maven-plugin/issues/497))
-  
+
 * **0.15.8** (2016-06-27)
   - Removed image configuration caching ([#495](https://github.com/fabric8io/docker-maven-plugin/issues/495))
   - Fix for tcp wait when used with Docker for Mac ([#430](https://github.com/fabric8io/docker-maven-plugin/issues/430))
   - Add warning when assembly is empty when watching a Docker image ([#490](https://github.com/fabric8io/docker-maven-plugin/issues/490))
-  - Add `docker.skip.build`, `docker.skip.run`, `docker.skip.push` properties and 
+  - Add `docker.skip.build`, `docker.skip.run`, `docker.skip.push` properties and
     renamed `docker.skipTags` to `docker.skip.tag` ([#483](https://github.com/fabric8io/docker-maven-plugin/issues/483))
   - Reverted jansi back to version 1.11 because of [this issue](https://github.com/fusesource/jansi/issues/58)
   - Add new assembly config options `permissions` for fine tuning permissions in the docker.tar ([#477](https://github.com/fabric8io/docker-maven-plugin/issues/477)). Deprecated `ignorePermissions`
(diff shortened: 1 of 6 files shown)


rhuss:
Thanks a lot, will have look soon.

jgangemi:
any updates on this?

rhuss:
Uff, everyone is on PTO except for me. Tons of stuff todo, but yeah, you are right, this should be a bit upper in the queue ;-) will try to do it this evening. If not, please keep nagging me ;-)

jgangemi:
cool - i'm going to try and get the push retry finished off today (i just need to do the tests) as well.

doc/changelog.md
```diff
@@ -3,15 +3,17 @@
 * **0.15-SNAPSHOT**
   - Dont do redirect when waiting on an HTTP port (#499)
   - Removed the container fetch limit of 100 and optimized getting containers by name and image (#513)
-  
+  - Allow images to only be pulled once per build (useful for reactor projects) ([#504](https://github.com/fabric8io/docker-maven-plugin/issues/504))
+
+>>>>>>> Stashed changes
```
rhuss: oops ;-)

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -425,6 +430,19 @@ protected void checkImageWithAutoPull(ServiceHub hub, String image, String regis
         }
     }
 
+    private synchronized Set<String> getPreviouslyPulledImageCache() {
+        @SuppressWarnings({ "rawtypes", "unchecked" })
+        Set<String> cache = (Set) session.getUserProperties().get(CONTEXT_KEY_PREVIOUSLY_PULLED);
+        if (cache == null) {
+            System.out.println("creating new context");
+
+            cache = Sets.newConcurrentHashSet();
```
rhuss: Any reason why a simple HashSet() doesn't work (or Hashtable() if it needs to be synchronized) ? Just curious whether the google collections provides some additional benefits. not a principal problems, since we already have this dependency ....

src/main/java/io/fabric8/maven/docker/service/QueryService.java
```diff
@@ -181,30 +181,42 @@ public boolean hasImage(String name) throws DockerAccessException {
...
+    }
+
+    private boolean imageRequiresPull(AutoPullMode autoPullMode, String imageName, boolean always, Set<String> previouslyPulled)
+            throws DockerAccessException {
+
+        if (autoPullMode == AutoPullMode.ONCE && previouslyPulled.contains(imageName)) {
+            return false;
+        }
+
+        previouslyPulled.add(imageName);
```
rhuss: I would update the set here as a side effect. Instead its better to do this when the auto pull actually happens.

rhuss:
Thanks a lot, LGTM. If you don't mind (and agree to my nit-picking ;-) I'd fix the minor gotchas myself, in order to speed things up.

rhuss:
And yes, thanks to get into jmockit. Not easy in the beginning, I know, but quite powerful (especially when it comes to partial mocks).

Also kudos for providing unit tests and docs from the very beginning.  Not the usual case.

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
jgangemi: lol - so you already merged this before i got to comment but the reason i did this was for when you run `-T 2C`. given we need to `synchronize` other methods, this seemed to be the safe option.

src/main/java/io/fabric8/maven/docker/service/QueryService.java
jgangemi: i only did this b/c if the image couldn't be pulled, the build would fail regardless but i'm ok w/ whatever change here.

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
rhuss: Sorry, I'm a bit in a hurry ;-). It perfectly fine for me ;-)

rhuss: Funnily this google code is not much more than a 

```
Collections.newSetFromMap(new ConcurrentHashMap<String,Boolean>())
```

@jgangemi At the moment when testing this I have problems with the FailSafePlugin which tries to write the `seesion.getUserProperties()` to disk because of a String class cast exception. And as Properties are supposed to carry String values this is not so a big surprise.

What do you think, should we serialize the map to JSON string, or do you can think about a different sharing mechanism among multiple reactor projects ?

jgangemi: i'm not sure what other sharing mechanisms we have and i'm going to guess that the plugin writes `user` properties and `system` properties right?

i don't want to make this overly complicated...what about a comma delimited list?

how do i trigger the integration tests locally?

rhuss: I could trigger an exception by calling the sample in data-jolokia. Comma
separated list sounds fine.

Jae Gangemi <notifications@github.com> schrieb am Di., 19. Juli 2016, 22:53:

rhuss: I fixed it by introducing an `ImagePullCache` which is serializable (via JSONObject). Seems to work now with the session userProperties (it would already work with the executionProperties() but that's deprecated and putting sth else than a String as a Property value feels wrong anyway.

PR #513, opened 2016-07-12
docker container fetch limit is now configurable

greggulrajani:
Signed-off-by: Gregory Gulrajani <gregory.gulrajani@nucleusfinancial.com>

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -120,6 +120,10 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
     @Parameter(property = "docker.skip", defaultValue = "false")
     private boolean skip;
 
+    // Max number of containers to fetch when stopping/removing
+    @Parameter(property = "docker.fetchLimit", defaultValue = "100")
+    protected int fetchLimit;
+
     /**
      * Whether to restrict operation to a single image. This can be either
      * the image or an alias name. It can also be comma separated list.
@@ -139,7 +143,7 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
     // property file to write out with port mappings
     @Parameter
     protected String portPropertyFile;
-    
+
     // Authentication information
     @Parameter
     Map authConfig;
@@ -178,7 +182,9 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
             DockerAccess access = null;
             try {
                 access = createDockerAccess(minimalApiVersion);
-                ServiceHub serviceHub = serviceHubFactory.createServiceHub(project, session, access, log, logSpecFactory);
+                fetchLimit = resolvedImages.get(0).getRunConfiguration().getFetchLimit() != null
+                        ? resolvedImages.get(0).getRunConfiguration().getFetchLimit() : fetchLimit;
+                ServiceHub serviceHub = serviceHubFactory.createServiceHub(project, session, access, log, logSpecFactory, fetchLimit);
                 executeInternal(serviceHub);
             } catch (DockerAccessException exp) {
                 log.error("%s", exp.getMessage());
@@ -225,6 +231,7 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
 
     // Customization hook for subclasses to influence the final configuration. This method is called
     // before initialization and validation of the configuration.
+    @Override
     public List<ImageConfiguration> customizeConfig(List<ImageConfiguration> imageConfigs) {
         return imageConfigs;
     }
--- a/src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
+++ b/src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
@@ -53,7 +53,7 @@ public final class UrlBuilder {
                 .p("follow", follow)
                 .build();
     }
-    
+
     public String createContainer(String name) {
         return u("containers/create")
                 .p("name", name)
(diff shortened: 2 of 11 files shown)


rhuss:
LGTM. I'm going to merge it and resolve the conflicts which happens because I moved the documentation to asciidoc.

src/main/java/io/fabric8/maven/docker/service/QueryService.java
```diff
@@ -52,7 +53,7 @@ public Container getContainer(String containerId) throws DockerAccessException {
      * @throws DockerAccessException in case of an remote error
      */
     public Container getContainerByName(final String containerName) throws DockerAccessException {
-        for (Container el : docker.listContainers(CONTAINER_LIMIT)) {
+        for (Container el : docker.listContainers(containerLimit)) {
```
rhuss: Having limit here is probably quite dangerous because one can easily miss the container with the given name returning a false result. The iteration is not optimal anyway, wonder whether there's an easier way to perform a 'fetchByName' ?

src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
```diff
@@ -70,13 +70,19 @@ public String inspectContainer(String containerId) {
...
+
     public String listContainers(int limit) {
         return u("containers/json")
                 .p("limit", limit)
                 .build();
     }
 
+    public String listAllContainers() {
+        return u("containers/json")
+                .p("all", 0)
```
rhuss: Shouldnt this be "1" ?

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -178,7 +182,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             DockerAccess access = null;
             try {
                 access = createDockerAccess(minimalApiVersion);
-                ServiceHub serviceHub = serviceHubFactory.createServiceHub(project, session, access, log, logSpecFactory);
+                fetchLimit = resolvedImages.get(0).getRunConfiguration().getFetchLimit() != null
```
greggulrajani: Well spotted. I meant to ask you about fetching the resolvedImage. I was not happy with the get(0) it's a bit dangerous ..

rhuss: Yeah, that was a misunderstanding on my side. A global parameter is good enough.

However, our problem indeed is how we can make the lookup of a container by its name and by its image more effective. Because then we could omit the call to list() altogether. And it would be much more effective, too, especially for a large set of running containers. At the time this was implemented, there was no way with the API todo this direct lookup, but it could be that there is now an alternative API call. 

I will go hunting a bit ... ;-)

rhuss: Just saw that there are new filters for the list call:
````
exited=<int>; -- containers with exit code of <int> ;
status=(created|restarting|running|paused|exited|dead)
label=key or label="key=value" of a container label
isolation=(default|process|hyperv) (Windows daemon only)
ancestor=(<image-name>[:<tag>], <image id> or <image@digest>)
before=(<container id> or <container name>)
since=(<container id> or <container name>)
volume=(<volume name> or <mount point destination>)
network=(<network id> or <network name>)
````

So maybe `since` and `ancestors` could be a match for our use case ....

rhuss: E.g. 

* for fetching by name: since=<container name>&limit=1
* for fetching by images: ancestor=<image name>&limit=1

Hopefully the filter is inclusive. 'will try this out now ...

src/main/java/io/fabric8/maven/docker/service/QueryService.java
greggulrajani: there is a a 'filter' argument that accepts key=label but oddly not 'Names'

src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
greggulrajani: Yep -- misread the documentation.
Will change.

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
greggulrajani: gah -- we are on 1.11 -- I will upgrade Nucleus's docker tomorrow to 1.24..

src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
rhuss: no worries, I already did some refactorings, will push my changes in an own branch soon.

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
rhuss: I'm also only on 1.21, however there is `since` query param (not as part of the filter), but it seems to eclusive, not inclusive (so the entry with the name itself is not included). 

For the name lookup I have to check why I haven't use the `/inspect` call. I remember darkly, there was a reason ;-)

'have to run now, see you tomorrow ...

rhuss: Cool, filtering with 'ancestor=<image name>' works. Unfortunately only available since 1.23

But with this two features we should get rid of the local iteration of image names (we leave it in nevertheless so that the plugin will work with older clients, too).

src/main/java/io/fabric8/maven/docker/service/QueryService.java
rhuss: I changed now `getContainerByName()` and `getContainerForImage()` to not use `listContainers()`. containerByName should work always, getImageForContainer only for API > 1.23 (Docker 1.11). As a fallback for getContainerForImage (which actually is only used by docker:stop and docker:watch afais) the only list() mechanism is used.

The limit was there to prevent unexpected runtime behaviour if you have thousands of containers running. And since the list of containers is sorted according to creation time, I thought that it would be a good idea to impose a limit here (assuming the d-m-p is used for tests and development only with no long running containers started).

But you are right, after the change a limit is even less important, so, if you don't mind I'm going to remove it altogether.

PR #516, opened 2016-07-19
allow retries of a failed docker:push (resolves #508)

### LINKED ISSUE #508: add retry option for docker:push
we occasionally get `500` errors when pushing to our docker registry (which probably has something to do w/ it being backed by s3).

the plugin should have an option that allows the push to be retried in this case.

jgangemi:
i went to update the docs to add the `retryCount` parameter and now i can't find them...

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
+++ b/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
@@ -24,6 +24,7 @@ import io.fabric8.maven.docker.model.NetworksListElement;
 import io.fabric8.maven.docker.util.*;
 import io.fabric8.maven.docker.access.hc.unix.UnixSocketClientBuilder;
 import org.apache.http.HttpResponse;
+import org.apache.http.client.HttpResponseException;
 import org.apache.http.client.ResponseHandler;
 import io.fabric8.maven.docker.access.hc.http.HttpClientBuilder;
 import org.json.JSONArray;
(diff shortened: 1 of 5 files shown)


jgangemi:
would also be great to get a release after this lands.

rhuss:
np, since I'm using now the fabric8 CD I can do releases by the push of a button. Hope I can finish tomorrow ....

jgangemi:
don't worry about killing yourself over this, the next day or two is fine w/ me.

src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -461,6 +465,26 @@ private HcChunkedResponseHandlerWrapper createPullOrPushResponseHandler() {
...
+    private boolean isRetryableErrorCode(int errorCode) {
+        // there eventually could be more then one of this
+        return errorCode == HTTP_INTERNAL_ERROR;
+    }
+
+    private void pushImage(String url, Map<String, String> header, HcChunkedResponseHandlerWrapper handler, int status,
+            int retries) throws IOException {
+        for (int i = 0; i <= retries; i++) {
+            try {
+                delegate.post(url, null, header, handler, HTTP_OK);
```
rhuss: Shouldn't there be a `return` after a successful post ?

src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -461,6 +465,26 @@ private HcChunkedResponseHandlerWrapper createPullOrPushResponseHandler() {
...
+            } catch (HttpResponseException e) {
+                if (isRetryableErrorCode(e.getStatusCode()) && retries > 0) {
+                    log.warn("failed to push image to [{}], retrying...", url);
+                } else {
+                    throw e;
+                }
+            }
+        }
+    }
```
rhuss: Also, when every retry failed, no error is thrown but silently returned at the end of the method.

src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -461,6 +465,26 @@ private HcChunkedResponseHandlerWrapper createPullOrPushResponseHandler() {
...
+        return errorCode == HTTP_INTERNAL_ERROR;
+    }
+
+    private void pushImage(String url, Map<String, String> header, HcChunkedResponseHandlerWrapper handler, int status,
+            int retries) throws IOException {
+        for (int i = 0; i <= retries; i++) {
+            try {
+                delegate.post(url, null, header, handler, HTTP_OK);
+            } catch (HttpResponseException e) {
+                if (isRetryableErrorCode(e.getStatusCode()) && retries > 0) {
```
rhuss: would put `&& i != retries` instead of `&& retries > 0`

PR #520, opened 2016-07-22
- use Futures instead of sleeping, request for feedback 

### LINKED ISSUE #518: replace `WaitUtil.sleep` with Futures
right now if you specify a `wait` time, the plugin just basically sleeps which means if the container exits sooner you still have to wait for the sleep to complete.

what (hopefully) would be better would be to do is use a `Future` what would allow the process to continue as soon as it completes and timeout if it takes longer then the allotted `wait` time.

thoughts?

jgangemi:
this will resolve #518 and i'm looking for some feedback on this approach before i finish it off w/ some tests etc.

the approach w/ the current implementation is if the container shuts down before the maximum grace period value, the plugin will still wait the entire time. for example, if you use a value of `30` and the container shuts down in `5` seconds, you're sitting around for `25` more doing nothing which doesn't make for fast builds.

the introduction of the `Future` lets us short circuit the extra wait time.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/WaitUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/WaitUtil.java
@@ -7,7 +7,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -24,6 +24,8 @@ import org.apache.http.impl.client.HttpClientBuilder;
  * @since 18.10.14
  */
 public class WaitUtil {
+
+    private static final int DEFAULT_DOCKER_TIMEOUT = 10;
 
     // how long to wait at max when doing a http ping
     private static final long DEFAULT_MAX_WAIT = 10 * 1000;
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/util/WaitUtil.java
```diff
@@ -48,6 +48,26 @@
...
+             */
+            task.get(((wait == 0) ? 10 : wait) + 5, TimeUnit.SECONDS);
+        } catch (@SuppressWarnings("unused") InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (@SuppressWarnings("unused") TimeoutException e) {
+            /*
+             * this should not happen b/c the wait time has been padded such that if the docker daemon fails
+             * to kill the container in the alloted time, it will be booted out via the ExecutionException
+             */
+        }
```
jgangemi: this is the bit i am most interested in feedback on. the idea here is the call to the `Future` itself should never time out b/c we should be relying on the docker daemon to throw the error in the event the container doesn't shut down in time.

the question here is what to do in the event that actually fails, which is theoretically possible. throw an `IllegalStateException`?

rhuss: Goof question ;-) From my feeling we should simply continue as we have waited the configured time (and even longer). As you said, its not likely to happen. Maybe we can output a warning so that we at least we get noticed in that case.

wdyt ?

jgangemi: i'm ok w/ a log statement but i still need to figure out the best way to deal w/ that given this class is static and i don't really want to pass it in as an argument. 

maybe i'll just throw the exception and have the caller deal w/ it b/c the logger will be available there.

jgangemi:
closing...

Issue #524, opened 2016-07-28
Add support for passing security options to docker


gschulze:
It would be nice if security options (--security-opt) were supported for running an image. In particular, this would allow to select a Seccomp security profile (https://docs.docker.com/engine/security/seccomp/), which is required for running systemd inside a container.

PR #525, opened 2016-07-28
#524 Add support for passing security options to docker

gschulze:
Implemented support for adding security options via <securityOpt> tag.

### DIFF
--- a/src/main/asciidoc/inc/start/_configuration.adoc
+++ b/src/main/asciidoc/inc/start/_configuration.adoc
@@ -83,6 +83,9 @@ a| Network mode for the container:
 | *restartPolicy*
 | <<start-restart,Restart Policy>>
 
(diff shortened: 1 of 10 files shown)


rhuss:
Thanks, will review it soon. Would you mind to add some documentations, too ? You find them in src/main/asciidoc and you can recreate the documentation with `mvn -Pdoc-html` from the root dir.

Thanks !

gschulze:
Sure!

rhuss:
Thanks, looks good to me ! 

I changed the property to `securityOpts` (plural) as it is a list. I know, Docker used `SecurityOpt` for this list, but this is imo an error we should not replicate.

I'll do a release today, including your change.

Thanks again !

gschulze:
That was fast! Thanks!

PR #527, opened 2016-07-29
Fix DockerAccessWHC.removeNetwork exception msg

danielwegener:
Should be network, not? :)


rhuss:
Good find ;-). Thanks !

PR #528, opened 2016-07-30
- run 'stopContainer' in a Future (resolves #518)

### LINKED ISSUE #518: replace `WaitUtil.sleep` with Futures

jgangemi:
resolves #518, now the plugin will continue as soon as the container stops, hurray faster builds!

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/RunService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/RunService.java
@@ -18,6 +18,8 @@ package io.fabric8.maven.docker.service;
  */
 
 import java.util.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 
 import io.fabric8.maven.docker.model.Container;
 import io.fabric8.maven.docker.log.LogOutputSpecFactory;
@@ -25,6 +27,7 @@ import io.fabric8.maven.docker.access.*;
 import io.fabric8.maven.docker.config.*;
 import io.fabric8.maven.docker.model.Network;
 import io.fabric8.maven.docker.util.*;
+import io.fabric8.maven.docker.util.WaitUtil.WaitTimeoutException;
 
 
 /**
@@ -116,13 +119,13 @@ public class RunService {
      * @param removeVolumes whether to remove volumes after stopping
      */
     public void stopContainer(String containerId,
-                              ImageConfiguration imageConfig,
-                              boolean keepContainer,
-                              boolean removeVolumes)
-            throws DockerAccessException {
+            ImageConfiguration imageConfig,
+            boolean keepContainer,
+            boolean removeVolumes)
+        throws DockerAccessException {
         ContainerTracker.ContainerShutdownDescriptor descriptor =
-                new ContainerTracker.ContainerShutdownDescriptor(imageConfig,containerId);
-        shutdown(docker, descriptor, log, keepContainer, removeVolumes);
+                new ContainerTracker.ContainerShutdownDescriptor(imageConfig, containerId);
+        shutdown(descriptor, keepContainer, removeVolumes);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


jgangemi:
@rhuss i'd appreciate consideration (and a release) at your earliest convenience. this shaved 10 minutes off our build.

rhuss:
ok, will have a look tomorrow. The only possible problem I see here is, that I already have concurrency issues with Docker daemon running over unix socket (when waiting to a log output in a different thread), so I'm afraid that for this use case there might even more serious issues. However, for TCP based connection everything should be fine.

jgangemi:
yeah - unix socket handling is crappy in general. i haven't looked at any of the libraries i saw to see if they were better then what we are using.

thinking about it now, it might not even need to run in a `Future` and we could just rely on the docker daemon to time itself out, but ultimately this is probably safer in that regard.

maybe we could add a `caveats` section that for the unix socket.

rhuss:
oops, I mixed this up with #531 (where I see more serious concurrency issue coming). Guess the future is more or less sage ;-). to be continued tomorrow ...

src/main/java/io/fabric8/maven/docker/util/WaitUtil.java
```diff
@@ -25,6 +25,8 @@
  */
 public class WaitUtil {
 
+    private static final int DEFAULT_DOCKER_TIMEOUT = 10;
```
rhuss: 10s maybe a bit high as default, which is even used when no killGracePeriod is specified. In fact I wouldn't wait this no killGracePeriod is specified so simply skipping any wait when `int wait` is null.

Any particular reason why you always want a wait when a shutdown happens ?

src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -412,32 +413,20 @@ private void shutdown(DockerAccess access,
...
         }
-        // Stop the container
-        int killGracePeriod = descriptor.getKillGracePeriod();
-        int killGracePeriodInSeconds = (killGracePeriod + 500) / 1000;
-        if (killGracePeriod != 0 && killGracePeriodInSeconds == 0) {
-            log.warn("A kill grace period of %d ms leads to no wait at all since its rounded to seconds. " +
-                     "Please use at least 500 as value for wait.kill",killGracePeriod);
-        }
-        access.stopContainer(containerId, killGracePeriodInSeconds);
-        if (killGracePeriod > 0) {
```
rhuss: There will be only a wait if killGracePeriod is > 0. I think we should do the same with the new logic, which means when no wait specified we shouldn't wait at all.

rhuss:
LGTM except for the default value.

If don't mind I'll fix that right now and merge to integration. We can still discuss this till tomorrow. You can expect a release tomorrow or the day after.

src/main/java/io/fabric8/maven/docker/util/WaitUtil.java
jgangemi: i don't always want to wait - the docker container takes up to 10 seconds which is why i used that default. if it exited before then, the future would complete, so i don't think we were waiting any extra time w/ this.

but whatever you think is best here.

jgangemi: yeah - looking at your change, i'm not 100% sure that's correct either in terms of returning `wait = 0` b/c you are always waiting some time for the daemon to complete, so i actually think the original implementation was correct.

```
get(long timeout, TimeUnit unit)
Waits if necessary for at most the given time for the computation to complete, and then retrieves its result, if available.```

if anything maybe it should be `if (wait == 0) { wait = DEFAULT }` (i can't remember what i originally had and i'm about to walk out the door for something).

PR #530, opened 2016-07-30
- don't pass build args that are empty (resolves #529)

### LINKED ISSUE #529: unable to override 'buildArg' properties
i have `docker.buildArg.version` set in order to set the version as a label on the container inside a `Dockerfile` (i thought i could use `$project.version` but it was never substituted and i didn't have time to track it down since i could make the build arg work) however there are times i would like to `null` that value out in a child pom. 

currently the properties are still added to the map as an empty value leading to the docker daemon giving an error about an unused argument.

or something a bit more concise...we shouldn't pass build args that have an empty value as a result of doing `<docker.buildArg.version />`

jgangemi:
resolves #529

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,7 +2,8 @@
 
 * **0.15.14** (2016-07-29)
   - Pattern match fix for multiline log output. Related to ([#259](https://github.com/fabric8io/docker-maven-plugin/issues/259))
-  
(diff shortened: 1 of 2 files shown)


rhuss:
thanks, lgtm

rhuss:
merged to `integration`

Issue #259, opened 2015-08-13
Invalid log format for ...  Failed to' (expected: "<timestamp> <txt>")


mason-uptake:
I'm sure you guys hate me by now :)

This is related to: https://github.com/rhuss/docker-maven-plugin/issues/242 

I saw it happen in 0.13.3-SNAPSHOT and also 0.13.4-SNAPSHOT. I posted about this in another pull request that got closed(so maybe you guys didn't see it?) but it probably makes sense to make another issue. I was able to replicate the issue here: https://github.com/mason/mvn-docker-plugin-bug

just do `mvn clean install` and you should see. attached is a screenshot.
<img width="1013" alt="screen shot 2015-08-12 at 6 07 10 pm" src="https://cloud.githubusercontent.com/assets/9662835/9252215/e1a8fc5e-419c-11e5-95e8-7f2a3d43bd10.png">

rhuss:
No, no, we **love** bug reports ;-) (because it manifests that ppl are using the software :)

Seems to be a bug when parsing the streaming answers. Could you provide a little bit more context, i.e. the point in the stack trace where the plugin kicks in ?

mason-uptake:
So I'm not sure if the stack trace in the image is related to the datetime bugs. If I run the same `mvn clean install` twice, I will always get a stacktrace due to failed tests, but the timestamp error doesn't always appear.

I ran the clean install a few times and I was able to generate a stacktrace from within the plugin.

Exception in thread "Thread-5" java.lang.IllegalArgumentException: Invalid format: "��2015-08-13T15:24..."
	at org.joda.time.format.DateTimeFormatter.parseDateTime(DateTimeFormatter.java:873)
	at org.jolokia.docker.maven.util.Timestamp.<init>(Timestamp.java:53)
	at org.jolokia.docker.maven.access.log.LogRequestor.callLogCallback(LogRequestor.java:142)
	at org.jolokia.docker.maven.access.log.LogRequestor.parseResponse(LogRequestor.java:121)
	at org.jolokia.docker.maven.access.log.LogRequestor.run(LogRequestor.java:97)

mason-uptake:
I've been doing some debugging and I've narrowed it down to:

 `private void callLogCallback(int type, String txt)` in `LogRequestor` class. I put a println in the plugin.

```
private void callLogCallback(int type, String txt) throws LogCallback.DoneException {
        Matcher matcher = LOG_LINE.matcher(txt);
[...]

PR #531, opened 2016-07-31
- parallel StartMojo

danielwegener:
that respects dependencies and starts all containers in parallel as soon as all dependencies are resolved. This can greatly improve startup time if the waits that you use are mostly io- or sleep related.

My current use-case: Starting 5-8 different applications on tomcats and wait until each of them is fully operational and ready for e2e tests (including hazelcast cluster convergence and JPA h2 initialization).

This feature reduced my regular startup-time from ~5:00 minutes to 1:23.

This is a POC that adds an `ExecutionService` and `ExecutorCompletionService` to the `StartMojo`. The nonParallel configuration simply falls back to a guava SameThreadExecutionService.

The Parallel start is configurable (`docker.startParallel`, disabled by default).

This PR adds another imageRunConfiguration property `dependsOn` (naming from docker-compose) which is the link-equivalent to express inter container dependencies in docker custom networks (it delays the start of a container until all of its dependent containers are started (inclusive wait)). 

There might be some concurrency bugs lurking around (I saw that most of the DockerAccess state is synchronized). Especially the ordered shutdown is a bit tricky (especially it would be great if we could shutdown the executor by not starting new tasks but also not stopping tasks that are running).

I am open for feedback :)

### DIFF
--- a/samples/custom-net/pom.xml
+++ b/samples/custom-net/pom.xml
@@ -31,6 +31,7 @@
           <logDate>default</logDate>
           <verbose>true</verbose>
           <autoPull>always</autoPull>
+          <startParallel>true</startParallel>
           <images>
             <image>
               <alias>box1</alias>
@@ -62,6 +63,9 @@
                   <name>test-network</name>
                   <alias>box2</alias>
                 </network>
+                <dependsOn>
+                  <dependsOn>box1</dependsOn>
+                </dependsOn>
                 <namingStrategy>none</namingStrategy>
                 <cmd>
                   <exec>
--- a/src/main/asciidoc/inc/_docker-start.adoc
+++ b/src/main/asciidoc/inc/_docker-start.adoc
@@ -29,6 +29,11 @@ include::start/_links.adoc[]
 
 include::start/_network.adoc[]
 
+[[start-depends-on]]
+=== Depends-On
+
+include::start/_depends-on.adoc[]
+
 [[start-restart]]
 === Restart Policy
 
--- a/src/main/asciidoc/inc/start/_configuration.adoc
+++ b/src/main/asciidoc/inc/start/_configuration.adoc
@@ -1,4 +1,20 @@
+In addition to the <<global-configuration>>, this goal supports the following global configuration options.
 
+.Start options
+[cols="1,5,1"]
+|===
+| Element | Description | Property
+
+| *showLogs*
+| In order to switch on globally the logs *showLogs* can be used as global configuration (i.e. outside of `<images>`). If set it will print out all standard
+ output and standard error messages for all containers started. As value the images for which logs should be shown can be given as a comma separated list.
+ This is probably most useful when used from the command line as system property `docker.showLogs`.
+| `docker.showLogs`
+| *startParallel*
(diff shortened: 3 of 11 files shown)


rhuss:
Thanks a lot ! I will have a look soon. I'm only afraid that when using the Unix socket connection to Docker that we will soon run in the concurrency issues we already have when waiting on logoutput. See #259 for details. If we would find either a unix socket Java lib which is thread safe or we could serialize the access to the socket, this would be perfect.

I had no chance yet to look into the PR, but hope I can do it this week (sorry, still quite busy with other stuff). Maybe you could make some tests with using the Unix socket in the meantime, too ?

danielwegener:
@rhuss Unfortunately I do not have a native docker host setup ready (usually using docker for mac native/machine), but I'll try to set it up later.

rhuss:
I just had the chance for a quick review, looks great ! And even if there are issues with the Unix socket (and soon to come, Windos named pipes, too), we could document that parallel execution is only supported for TCP connections.

Some things still which I'd need:

* Could you please sign-off your commit as described in https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/contributing.md . Feel free to rebase or squash the commits, too.
* Documentation would be awesome, too. This would go into https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/asciidoc/inc/start/_configuration.adoc for `dependsOn` and `https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/asciidoc/inc/start/_overview.adoc` for `docker.startParallel`
* I know, its hard, but tests would be awesome ! Maybe one could extract the starting functionality into an external class which can be easier unit tested than the whole StartMojo. Also, it would make the Mojo smaller as these beasts tend to grow all the time ;-)

BTW, I wonder whether this would make sense for building images, too ? well, probably not so much as one does not has many images to build anyway.

Thanks again, will add your contribution in the overnext release (since I have to cut a release right now).

danielwegener:
Thanks for the feedback. All right, gonna try to come up with some tests and docs.

I think I'll put the `executorService.awaitTermination` timeout to infinity such that, given enough time, all starting containers will eventually start and, given one of the containers fail to start, all started containers will be shut down properly (this assumes that each single start job has its own timeout and will terminate eventually). WDYT?

Concerning image builds: It definitely depends on the build job if the parallel image builds bring real benefits (for my use case its not such a big deal because we just stack some local files onto an already existing base image) - but images with a lot of latency bound io (lots of small network connections, `apt-get update` for example) could benefit. But I think the effect will not be as strong as waiting for polling start wait conditions.

rhuss:
Just want to let you know that I will soon be off for PTO in the next three weeks and wont be able to continue here. but after this we should integrated this nice PR.

danielwegener:
Sorry, that took a while (I was also on PTO :)). I tuned the custom-net example to use dependsOn and parallelStart and added some bits of documentation (required some restructuring).

The refactoring/testing of the parallel start is not so easy without either introducing methods which basically "give me the whole world"-parameter lists or by working around the mojo-cdi pattern. We could simply rip out some parts of the `startInternal` method via extract method - but I am not sure if that would make it much clearer (some parts like the start order resolver could be better be tested in isolation though).

danielwegener:
Cool - thanks alot :)

PR #533, opened 2016-08-02
Add custom network IP address property mappings

jdcasey:
This adds property mappings for custom networks, using the format: 
```
docker.container.<alias>.net.<name>.ip
```
It seems important when the build is running on a container, and tries to start other containers in order to use them as remote services for integration testing. In this case, the only way I've found success is by using a custom network created using:

```
    $ docker network create -d bridge ci-network
```

My Jenkins instance is running in its own docker container, so the bridge aspect allows me access to Jenkins. All the docker containers started by Jenkins will also need to use this custom network, I've found, and when the build needs to access them it can use the container's IP address on the custom network along with the ports from the EXPOSE line in the Dockerfile (not mapped ports).

This works for me, but I'm not sure I understand how to write tests for the patch I'm submitting. I've tried to match code formatting, but ended up just guessing at the rules. Please let me know if I've missed something that needs attention.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/StartMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StartMojo.java
@@ -8,12 +8,6 @@ package io.fabric8.maven.docker;
  * the License.
  */
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Properties;
-import java.util.regex.Pattern;
-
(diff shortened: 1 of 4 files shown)


rhuss:
LGTM, thanks !

I'll add some documentation and try to add a unit test for the IP extraction part.

Code formatting looks fine, its not so strict anyway ;-)

rhuss:
fyi, 0.15.16 is released .....

PR #536, opened 2016-08-09
Fixes for stop mojo

transamericamoon:
Fixed: stop containers based on naming strategy (should stop all containers based on alias if naming strategy is used)

Fixed: catch exception if container alias does not exists (previously it would throw exception now it just displays a warning)

Fixed: docker.allContainers, docker.sledgeHammer can be set from maven properties now (before only from system properties).

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/StopMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StopMojo.java
@@ -95,15 +95,15 @@ public class StopMojo extends AbstractDockerMojo {
         //Do not fail if container is not found
         try {
             
-	        if (strategy == RunImageConfiguration.NamingStrategy.alias) {
+            if (strategy == RunImageConfiguration.NamingStrategy.alias) {
                 Container container = queryService.getContainer(image.getAlias());
-		        if (container !=  null) {
-	                containers = Collections.singletonList(container);
-		        }
-	        } else {
-	            containers = queryService.getContainersForImage(image.getName());
-	        }
-	        
+                if (container !=  null) {
+                    containers = Collections.singletonList(container);
+                }
+            } else {
+                containers = queryService.getContainersForImage(image.getName());
+            }
+            
             if(containers == null) {
                 containers = Collections.emptyList();
             }
--- a/src/main/java/io/fabric8/maven/docker/StopMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StopMojo.java
@@ -47,6 +47,13 @@ public class StopMojo extends AbstractDockerMojo {
      */
     @Parameter(property = "docker.autoCreateCustomNetworks", defaultValue = "false")
     protected boolean autoCreateCustomNetworks;
+    
+    @Parameter( property = "docker.allContainers", defaultValue = "false" )
+    private boolean allContainers;
+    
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks ! I will review it ASAP, however I'm soon off for 3 weeks, so it might take a bit

transamericamoon:
no worries, have a good vacation.

src/main/java/io/fabric8/maven/docker/StopMojo.java
```diff
@@ -82,40 +89,56 @@ private void stopContainers(QueryService queryService, RunService runService, Po
...
             } else {
+                containers = queryService.getContainersForImage(image.getName());
+            }
+            
+            if(containers == null) {
                 containers = Collections.emptyList();
             }
-        } else {
-            containers = queryService.getContainersForImage(image.getName());
+        } catch (Exception ex) {
```
rhuss: Shouldn't this be more specific and e.g. around `queryService.getContainer()` ? (since the other method wont throw an exception in case of an unknown name)

src/main/java/io/fabric8/maven/docker/StopMojo.java
```diff
@@ -82,40 +89,56 @@ private void stopContainers(QueryService queryService, RunService runService, Po
...
+            if (strategy == RunImageConfiguration.NamingStrategy.alias) {
+                Container container = queryService.getContainer(image.getAlias());
+                if (container !=  null) {
+                    containers = Collections.singletonList(container);
+                }
             } else {
+                containers = queryService.getContainersForImage(image.getName());
+            }
+            
+            if(containers == null) {
```
rhuss: At that point, containers will never be null since queryService.getContainersForImage() will always return a list

src/main/java/io/fabric8/maven/docker/StopMojo.java
```diff
@@ -82,40 +89,56 @@ private void stopContainers(QueryService queryService, RunService runService, Po
...
         }
+        
+        RunImageConfiguration.NamingStrategy strategy = image.getRunConfiguration().getNamingStrategy();
+        
+        if (RunImageConfiguration.NamingStrategy.alias.equals(strategy)) {
+            if(container.getName().equals(image.getAlias())) {
+                return true;
+            }
+            return false;
+
```
rhuss: can be simplified to 

```
return container.getName().equals(image.getAlias());
```

src/main/java/io/fabric8/maven/docker/StopMojo.java
rhuss: Hmm, still wondering where an exception could occur in case of an non existing container:

* getContainersForImage() returns an empty list if no container is found
* getContainer() return null if none is found

So if an exception is occurs is must be sth more serious which I wouldn't catch.

rhuss: But I probably overlooked sth here. Could you give an example where an exception happened for you ?

rhuss: Yeah, but then the exception when I try to stop it. This try block is only about querying for containers / images, so I think it safe to remove the 'catch' (and I did it for the latest release).

PR #565, opened 2016-09-25
Avoid auto-pulling 'scratch' image

Pac72:
Since version 1.5.0, Docker interprets the Dockerfile instruction `FROM scratch` as a special case indicating a no-base image.
This implies that there is no more an image named 'scratch'.
Building an image from a Dockerfile with `FROM scratch` currently
causes the error:
`Unable to pull 'scratch:latest': 'scratch' is a reserved name (Internal Server Error: 500)`
Avoid auto-pulling 'scratch' image.


rhuss:
Thanks ! will add this to the next release.

PR #569, opened 2016-09-27
Adding property for plugin configurations 

nirkoren:
Hi,

We need to add this alias so we can use it in the plugin configuration and not only via command line: 
Something like that. 
`<removeAll>true</removeAll>`

Is it fine from your side?

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/RemoveMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/RemoveMojo.java
@@ -44,7 +44,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 public class RemoveMojo extends AbstractDockerMojo {
 
     // Should all configured images should be removed?
-    @Parameter(property = "docker.removeAll", defaultValue = "false")
+    @Parameter(property = "docker.removeAll", defaultValue = "false", alias = "removeAll")
     private boolean removeAll;
 
     @Override


src/main/java/io/fabric8/maven/docker/RemoveMojo.java
```diff
@@ -44,7 +44,7 @@
 public class RemoveMojo extends AbstractDockerMojo {
 
     // Should all configured images should be removed?
-    @Parameter(property = "docker.removeAll", defaultValue = "false")
+    @Parameter(property = "docker.removeAll", defaultValue = "false", alias = "removeAll")
```
rhuss: From the [Maven documentation](http://books.sonatype.com/mvnref-book/reference/writing-plugins-sect-mojo-params.html#writing-plugins-sect-param-annot) : 

>  The alias provides the name of the parameter. If alias is omitted, Maven will use the name of the variable as the parameter name. 

So this change is a noop since the field is called also `removeAll`.

In fact, you can already use `removeAll` as a plugin configuration parameter:

Either in a `<properties>` sections:

```xml
<properties>
   <docker.removeAll>true</docker.removeAll>
</properties>
```

or direct in the plugin configuration:

```xml
<plugin>
  ...
  <configuration>
    <removeAll>true</removeAll>
    ....
  </configuration>
</plugin>
```

Please let me know whether this work for you. If not, its a bug.

nirkoren:
Right! somehow it didnt work for me but now it does.... I am closeing this PR

PR #570, opened 2016-09-27
refactorings to make it easier to implement fabric8:watch

jstrachan:


rhuss:
Already pushed to `integration` will be available very soon in 0.16.5

PR #571, opened 2016-09-27
OSFUSE-283: Support using a more complex dom object for the for confi…

chirino:
…guration.


src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -34,7 +36,7 @@
     /**
      * @parameter
      */
-    private String from;
+    private Xpp3Dom from;
```
rhuss: Does Maven parameter injection with a string value still work ?

rhuss:
The CI test error is my fault, will fix that.

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
chirino: just tested.. and no. :(  Let me think about this a bit more!

PR #576, opened 2016-10-06
Do not thow exception for remove failures

transamericamoon:
if the image we are removing does not exists should we throw a failure?  Our current maven workflow is:

```
          <goals>
                <goal>stop</goal>
                <goal>remove</goal>
                <goal>build</goal>
                <goal>start</goal>
            </goals>
```

First time we run the image doesn't exists, and since remove throws an exception we cannot proceed without changing the pom.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
+++ b/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
@@ -413,7 +413,8 @@ public class DockerAccessWithHcClient implements DockerAccess {
 
             return response.getStatusCode() == HTTP_OK;
         } catch (IOException e) {
-            throw new DockerAccessException(e, "Unable to remove image [%s]", image);
+            log.warn( "Unable to remove image [%s]", image);
+            return false;
         }
     }
 


src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -413,7 +413,8 @@ public boolean removeImage(String image, boolean... forceOpt) throws DockerAcces
 
             return response.getStatusCode() == HTTP_OK;
         } catch (IOException e) {
-            throw new DockerAccessException(e, "Unable to remove image [%s]", image);
+            log.warn( "Unable to remove image [%s]", image);
```
rhuss: Question here is, why an `IOException` is thrown, since when an Image is not found, it shouldn't throw an exception but return a 404 (which is allowed in line 409 without throwing an exception). 

In any other case, i.e. when the call itself went wrong and an IOException occurs, I probably wouldn't want to continue as usual.

Do you have a specific scenario and the specific IOException that occurs for you ?

rhuss:
Do you have still interest in this PR and could please describe your use case ? If so please let me know until mid next week.

transamericamoon:
I'll just close it for now, i put the exception handling in my own build to get our stuff working.  I'll revisit it when we update the plugin.  Thanks.

PR #579, opened 2016-10-07
Removed restriction that port mappings are integers to support UDP

tmckeag:
I have a need to expose UDP ports by adding a /udp at the end of the port mappings.  Making this a string allows the same capability one would normally put in a docker file.  

I don't know if we'd want to validate the port mapping form at build time or not... if so, I'd suggest reusing the logic in io.fabric8.maven.docker.access.PortMapping, as it is a more complete representation of what one can do.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
@@ -48,7 +48,7 @@ public class DockerFileBuilder {
     private List<AddEntry> addEntries = new ArrayList<>();
 
     // list of ports to expose and environments to use
-    private List<Integer> ports = new ArrayList<>();
+    private List<String> ports = new ArrayList<>();
 
     // list of RUN Commands to run along with image build see issue #191 on github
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks a lot for the PR and sorry for the delay.

Looks good but as you mentioned we should probably validate the ports already during buildtime (_fail fast_). Although `PortMapping`  has the appropriate logic and pattern, is to deeply integrated, so that I wouldn't mind to put the (simple) validation logic also within the `DockerfileBuilder`.

Could you update your PR for this ? Also I've seen that the Git email address  doesn't match your Github account (seen as a '?' after your user name in the commit). Would be awesome if you could fix this, too and sign-off your commit a described [here](https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/contributing.md)

tmckeag:
Will do.  I have to travel over the next week for work so it might not get done for another week or two.  Thanks for the doc, this is actually my first time sending a PR on GitHub.  I'll correct it.

rhuss:
hey @tmckeag , thanks a lot and congratulations to your first PR. 'hope for many many more to come ;-)

PR #580, opened 2016-10-08
multiple links in single property, closes #558

### LINKED ISSUE #558: multiple links in single property
Our corporate parent pom configures docker plugin and sub-projects use the docker packaging.  We'd like to have each sub-project have different <links>.  Unfortunately, it's rather difficult to provide just the <links> configuration in the sub-project.  I cannot figure any combination of combine.children and combine.self that doesn't require re-configuring large sections of the image/run element.  What I have done currently is set a single link to a property which is defined in the sub-project.

I now need multiple links.  I'd like to either add a new image/run element called linksList the takes a comma/space separated list of link declarations OR modify <link> to support multiple declarations.  Or, do you have other ideas how to add multiple links in a template fashion? 

I'm will work on pull request based upon your suggested direction.

chonton:


src/main/java/io/fabric8/maven/docker/config/handler/property/PropertyConfigHandler.java
```diff
@@ -155,21 +160,23 @@ private String extractName(String prefix, Properties properties) throws IllegalA
...
         if (ports == null) {
             return null;
         }
-        List<String[]> parsedPorts = EnvUtil.splitOnLastColon(ports);
-        for (String[] port : parsedPorts) {
-            ret.add(port[1]);
-        }
-        return ret;
+        Iterable<String[]> parsedPorts = EnvUtil.splitOnLastColon(ports);
+        return Lists.newArrayList(Iterables.transform(parsedPorts, EXTRACT_SECOND));
```
rhuss: All in all this is not much smaller than the first, more direct version for extracting the port ;-) As much as I appreciate higher abstractions were necessary, in this case I believe the original version is more readable (and does not require knowledge of guava)

src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -78,34 +81,54 @@ public static boolean greaterOrEqualsVersion(String versionA, String versionB) {
...
     }
 
+    private static final Function<String, String[]> SPLIT_ON_LAST_COLON = new Function<String, String[]>() {
+        @Override
+        public String[] apply(String element) {
+          int colon = element.lastIndexOf(':');
+          if (colon < 0) {
+              return new String[] {element, element};
+          } else {
+              return new String[] {element.substring(0, colon), element.substring(colon + 1)};
```
rhuss: what if colon is the last character in `element` ? wouldn't then `element.substring(colon+1)` throw an error here ?

src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -78,34 +81,54 @@ public static boolean greaterOrEqualsVersion(String versionA, String versionB) {
...
-
-            for (String element : listToSplit) {
-                String[] p = element.split(":");
-                String rightValue = p[p.length - 1];
-                String[] nameParts = Arrays.copyOfRange(p, 0, p.length - 1);
-                String leftValue = StringUtils.join(nameParts, ":");
-                if (leftValue.length() == 0) {
-                    leftValue = rightValue;
-                }
-                ret.add(new String[]{leftValue, rightValue});
```
rhuss: Oops, that's really a unneeded complicated algo. good that its simplified now :)

src/main/java/io/fabric8/maven/docker/config/RunImageConfiguration.java
```diff
@@ -263,18 +264,14 @@ public VolumeConfiguration getVolumeConfiguration() {
...
-        return links;
+    public Iterable<String> getLinks() {
+        return EnvUtil.splitAtCommasAndTrim(links);
     }
 
     public List<UlimitConfig> getUlimits() {
         return ulimits;
     }
 
-    public void setUlimits(List<UlimitConfig> pUlimits) {
```
rhuss: thanks for cleaning up :)

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
```diff
@@ -108,7 +108,7 @@ private void addVolumes(RunImageConfiguration runConfig, List<String> ret) {
 
     private void addLinks(RunImageConfiguration runConfig, List<String> ret) {
         // Custom networks can have circular links, no need to be considered for the starting order.
-        if (runConfig.getLinks() != null && !runConfig.getNetworkingConfig().isCustomNetwork()) {
+        if (!runConfig.getNetworkingConfig().isCustomNetwork()) {
```
rhuss: Could we leave this check in ? I know its not needed since `EnvUtil.splitOnLastColon()` always return a list but it makes it easier to understand what the conditions are for links to add.

src/main/java/io/fabric8/maven/docker/config/RunImageConfiguration.java
```diff
@@ -263,18 +264,14 @@ public VolumeConfiguration getVolumeConfiguration() {
         return volumes;
     }
 
-    public List<String> getLinks() {
-        return links;
+    public Iterable<String> getLinks() {
```
rhuss: Can we leave `List<String>` as return type ? All the other methods in RunImageConfiguration return a `List` for collections, and also it is not the most generic type which can be used here, I would like to keep it uniform.

rhuss:
Thanks a lot ! Merged in you changes to branch `integrtation` and will be part of the next release.

PR #589, opened 2016-10-15
Clarify log message when removing old image

hwellmann:
When building an image, the log messages are slightly confusing:

> [INFO] DOCKER> [mydata] "data": Built image sha256:04512
> [INFO] DOCKER> [mydata] "data": Removed image 

This sounds like the image gets deleted immediately after being built. In fact, it's only the previous version that is deleted.

With this change, the message is now

> [INFO] DOCKER> [mydata] "data": Removed old image sha256:e0bea


rhuss:
Yeah, thats indeed quite confusing. Thanks for the heads up.

Issue #595, opened 2016-10-19
Sometimes don't match log output when wait on log


smoczarski:
When configured in run section to wait until a defined log message happens sometimes this log output won't be recognized. 
Config:
```
<run>
...                                        
   <wait>
      <log>Setup finished</log>
      <time>30000</time>
   </wait>
</run>
```
Sometime the build breaks because of this:
```
setup> Setup finished
...
[ERROR] DOCKER> [tns-ticket/setup] "setup": Timeout after 30076 ms while waiting on log out 'Setup finished'
[ERROR] DOCKER> Error occurred during container startup, shutting down...
```
Version: 0.16.8

smoczarski:
Please see pull request https://github.com/fabric8io/docker-maven-plugin/pull/596

rhuss:
Thanks a lot for issue and especially for the analysis and fix as well. Your PR has been applied to branch `integration`, will be available soon in the next release.

smoczarski:
Thanks a lot for applying the patch!

seanyinx:
@rhuss I am still having this issue with version 0.20.0.
```
crossapp-server> 2017-05-23 13:41:25,068 [INFO] Register microservice instance success. microserviceId=2 instanceId=2 endpoints=[rest://localhost:8080] lease 120s 
[ERROR] DOCKER> [crossapp-server:0.1.0-SNAPSHOT] "crossapp-server": Timeout after 120029 ms while waiting on log out 'Register microservice instance success'
```
Running 5 integration tests using docker maven plugin. All servers were started using docker and 4 out of 5 were successful. Sometimes all 5 tests can pass.

rhuss:
@seanyinx Are you using unix socket or a TC connection to the Docker daemon ?

seanyinx:
@rhuss I run the tests on Travis.
you can take a look at the build logs at https://travis-ci.org/WillemJiang/java-chassis/builds.
The recent failed builds are mostly related to this issue.

mdsardar:
Anyone still facing this issue ? I'm getting the same when trying to initiate a docker image to run some funtional-tests. 
Any insights ?

[INFO] --- docker-maven-plugin:0.33.0:start (start-containers) @ function-tests ---
[INFO] DOCKER> [maven.com:5001/you/dev/couchbase:enterprise-6.5.0-2020] "couchbase": Start container 5a4af7309b2b
[...]

PR #596, opened 2016-10-19
#595 - Fix concurrency issue on log wait.

smoczarski:
Removed the synchronized because the `check()` method will always performed by the same thread. The member `detected` was only partly covered by the original synchronized because it was written in an anonymous inner class (`LogCallback`) which was not part of the synchronized block.
This fix mark the `detected` flag as `volatile` to ensure that a write by an other thread calling the log callback will be propagate to the thread who calls the `check()` method.


rhuss:
Thanks a lot. I'm currently on the road but will be back soon. Will check out the PR ASAP.

rhuss:
You are absolutely correct, `detected` needs to be volatile. Thanks for spotting this, the PR will be included in the next release.

rhuss:
0.16.9 is out with fix applied

smoczarski:
Fantastic - thanks a lot!

Issue #594, opened 2016-10-19
Adding support for Docker 1.12 HEALTHCHECK instruction


nicolaferraro:
I can work on this

rhuss:
awesome, go ahead ...

nicolaferraro:
@rhuss do you think we should add something like this:

```xml
<image>
  <alias>box1</alias>
  <name>busybox</name>
  <build>
    <healthCheck>--interval=5s --timeout=3s CMD echo "good"</healthCheck>
  </build>
</image>
```

Or it's better the `<healthCheck>` to be a complex type, like:

```xml
<image>
  <alias>box1</alias>
  <name>busybox</name>
  <build>
    <healthCheck>
      <interval>5s</interval>
      <timeout>3s</timeout>
      <command>echo "good"</command>
    </healthCheck>
  </build>
</image>
```

I think I know the answer...

rhuss:
LOL ;-P Didn't know that I'm so predictable :)

Of course I'd prefer the second solution, also because we could reuse [Arguments](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/config/Arguments.java) for the actual healthcheck (wonder though, whether Docker supports this). Its also documented in https://dmp.fabric8.io/#misc-startup which could be reused, too.

There is another thing, which should be implemented, too. In [initAndValidate](https://github.com/fabric8io/docker-maven-plugin/blob/5b1006d669c080ff87e6680b6790b2934fad7957/src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java#L155-L155) the minimal Docker API version is calculated based on the features used. So, when a health check is used, then this minimal API version should be set to 1.24 (which is the API version of Docker 1.12)

rhuss:
0.16.9 with healthceck support is out.

danielwegener:
Hi guys. Just curious, is this healthcheck also used during docker:start to ensure a proper start order? I am thinking about migrating wait statements to HEALTHCHECKs but need too keep the _dependencies are started completely_ semantic.

PR #597, opened 2016-10-21
Adding support for health checks #594

nicolaferraro:
This should add support for declaring new Docker health checks in containers.


rhuss:
Thanks a lot, really like it ;-) I looked hard, but couldn't find something to nit pick about ;-P

I will add some docs and it will be available in the next release (probably tomorrow or so).

BTW, your choice of prefixing a `#` to your branch name triggers some bug in GitHub, e.g. when copying the manual checkout instruction, they are not escaping it with a `\` so that it can be directly used via C&P ;-) no problem at all, just saying ...

samples/healthcheck/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+              <build>
+                <healthCheck>
+                  <mode>cmd</mode>
+                  <interval>5m</interval>
+                  <cmd>
+                    <exec>
+                      <args>curl</args>
+                      <args>-f</args>
+                      <args>http://localhost/</args>
+                      <args>||</args>
```
rhuss: Short question: Does this really work as I thought that `||` is a shell feature and you are using a direct "exec" call here.

Issue #567, opened 2016-09-26
Plugin triggers maven-checkstyle-plugin during build


kuhnroyal:
Somehow the plugin triggers the maven-checkstyle-plugin which is bound to the `validate` phase during a build. Checkstyle is run twice and potentially fails the build during the second execution, since it now scans generated sources etc.
I can't track it down, if I remove the docker build, checkstyle is not triggered twice.

```
...
[INFO] >>> docker-maven-plugin:0.16.3:build (build) > initialize @ foo-bar >>>
[INFO] 
[INFO] --- maven-checkstyle-plugin:2.17:check (validate) @ foo-bar ---
[INFO] Starting audit...
Audit done.
[INFO] 
[INFO] <<< docker-maven-plugin:0.16.3:build (build) < initialize @ foo-bar <<<
[INFO] 
[INFO] --- docker-maven-plugin:0.16.3:build (build) @ foo-bar ---
...
```

```
                <plugin>
                    <groupId>io.fabric8</groupId>
                    <artifactId>docker-maven-plugin</artifactId>
                    <version>0.16.3</version>
                    <executions>
                        <execution>
                            <id>build</id>
                            <phase>package</phase>
                            <goals>
                                <goal>build</goal>
                            </goals>
                        </execution>
                        <execution>
                            <id>push</id>
                            <phase>deploy</phase>
                            <goals>
                                <goal>push</goal>
                            </goals>
                        </execution>
                    </executions>
                </plugin>
                <plugin>
                    <artifactId>maven-checkstyle-plugin</artifactId>
                    <executions>
                        <execution>
                            <id>validate</id>
                            <phase>validate</phase>
                            <goals>
                                <goal>check</goal>
                            </goals>
                        </execution>
                    </executions>
                </plugin>
```

rhuss:
The problem is that the `docker:build` forks the lifecycle up to "initialize" which is one after "validate". These two phases are not bound to by default and hence are no-ops by default. @jgangemi and others had a use case for a pre-docker build hook like copying some resources, so we though it is a good idea to use this very early, unbound phase as such a hook.

However this means that for every d-m-p goal `validate` and `initialize` is called (that the drawback of forking the lifecycle). 

[...]

PR #599, opened 2016-10-24
Introduce -fork versions of mojos

jbarnettwomply:
Specifically revert the behavior change from commit 62d7ee3aee5a5df90b6e3c631a484d0dce3a29a7
and introduce -fork versions of those mojos that allow the lifecycle to be forked

see the discussion in issue #567 for context


rhuss:
I changed it now as discussed in #567 

Thanks for taking care.

jbarnettwomply:
Great, thanks!

PR #610, opened 2016-11-03
Removed restriction that port mappings are integers to support UDP

tmckeag:
Hi @rhuss, sorry for the confusion, I've been out awhile with pneumonia.  Finally got back to this, and my branch was far enough behind (and I apologize I'm kind of a Git newb) I thought it would be easier to apply to a fresh fork.  

Your original link for committing guidelines doesn't work anymore, so I don't think I got everything right, other than correcting the author on the commit.  Let me know if I need to do anything else.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
@@ -3,6 +3,8 @@ package io.fabric8.maven.docker.assembly;
 import java.io.File;
 import java.io.IOException;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.google.common.base.Joiner;
 import io.fabric8.maven.docker.config.Arguments;
@@ -52,7 +54,7 @@ public class DockerFileBuilder {
     private List<AddEntry> addEntries = new ArrayList<>();
 
     // list of ports to expose and environments to use
-    private List<Integer> ports = new ArrayList<>();
+    private List<String> ports = new ArrayList<>();
 
     // list of RUN Commands to run along with image build see issue #191 on github
     private List<String> runCmds = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


rhuss:
No problem, hope you are fine again !

Thanks a lot for your revised PR, looks good to me. The link for the contribution guidelines is here https://github.com/fabric8io/docker-maven-plugin/blob/master/CONTRIBUTING.md

Actually the only thing you would need to do is a 'git commit -s --amend' on your master tree (if this is still your last commit) and then a 'git push --force' to update the PR.

After this we can merge it into trunk.

src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
```diff
@@ -235,14 +237,48 @@ private String quote(String value) {
...
+        try {
+            Matcher matcher = PROTOCOL_SPLIT_PATTERN.matcher(input);
+            // Matches always.  If there is a tcp/udp protocol, should end up in the second group
+            // and get factored out.  If it's something invalid, it should get stuck to the first group.  
+            matcher.matches();
+            
+            // First group is the port mapping.  
+            // It may be split with a : in a variety of ways.  
+            String mapping = matcher.group(1);
+            String[] mappingParts = mapping.split(":", 3);
```
rhuss: Just re-looked at this PR and I think that is not correct here as we are talking about the ports to expose in a Dockerfile and not about mappings. So there must not be a `:` inlcude in the specificiation and only `8080` and `8080/tcp` should be allowed.

src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
```diff
@@ -235,14 +237,48 @@ private String quote(String value) {
...
+            String[] portsS = ports.toArray(new String[]{});
+            for(String port : portsS) {
+            	validatePortMapping(port);
             }
             DockerFileKeyword.EXPOSE.addTo(b, portsS);
         }
     }
+    
+    // Pattern for splitting of the protocol part of the port.  
+    private static final Pattern PROTOCOL_SPLIT_PATTERN = Pattern.compile("(.*?)(?:/(tcp|udp))?$");
```
rhuss: Guess we can match case insensitive here, and do a lowercase when adding them to the port list. makes it a bit more flexible for users.

src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
rhuss: If you don't mind I'd fix this now, since I have to make a release right now.

PR #613, opened 2016-11-07
Fixes bug where status was not being set on HttpConfiguration

nayyara-samuel:
The status parameter in the `HttpConfiguration` constructor is ignored and results in a null status. This is not a problem when using an XML config such as below, because the fields are injected differently. 
```xml
<wait>
    <http>
        <url>http://localhost/status</url>
        <method>GET</method>
        <status>200</status>
    </http>
</wait>
```
Found this out when writing an `ExternalConfigHandler` which uses code to construct the `WaitConfiguration` object.


rhuss:
Good find, thanks !

PR #618, opened 2016-11-15
Making the assembly 'targetDir' parameter usable.

Tzrlk:
This is somewhat of a blind fix, so apologies for the waste of time if it's not actually useable.

I found that when executing the plugin, it would fail when attempting to set the `targetDir` variable, but not basedir. I'm assuming that's because targetDir isn't actually annotated as a parameter. I could be wrong.

I'm just working around the issue in my project at the moment, but I reckon if this is a quick fix for an oversight, I might as well submit it for inspection at the very least.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/AssemblyConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/AssemblyConfiguration.java
@@ -12,6 +12,7 @@ public class AssemblyConfiguration implements Serializable {
 
     /**
      * @parameter
+     * @deprecated Use {@link #targetDir} instead.
      */
     @Deprecated
     private String basedir;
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks for the PR !

Sorry for the confusion, the documentation didn't fit really the latest release (normally I used a dedicated branch `integration` for doing documentation updates, however in this case this slipped through). I will make a proper 0.18.1 release right now.

Your fix doesn't really change anything since `@Parameter` is not use by Maven within inner configuration objects but only on top-level config objects within a Mojo. In this case it uses only pure reflection to inject the data, so all the annotation and doc comments are only serve as some kind of documentation.

rhuss:
With the release coming out in an hour, targetDir should work as expected.

PR #624, opened 2016-11-25
Handle properly when dockerFile and dockerFileDir also set

lmesz:
When both parameter set the dockerFileDir was ignored because of the logic how to find the given dockerfile.
After the pull request when both parameter set docker-maven-plugin determine properly the location of the Dockerfile.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -457,7 +457,14 @@ public class BuildImageConfiguration implements Serializable {
 
     // Initialize the dockerfile location and the build mode
     private void initDockerFileFile(Logger log) {
-        if (dockerFile != null) {
+        if (dockerFile != null && dockerFileDir != null) {
+            dockerFileFile = new File(dockerFile);
+            if (dockerFileFile.isAbsolute()) {
+                throw new IllegalArgumentException("<dockerFile> can not be absolute path if <dockerFileDir> also set.");
+            }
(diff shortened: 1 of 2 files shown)


src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -457,7 +457,10 @@ public String initAndValidate(Logger log) throws IllegalArgumentException {
 
     // Initialize the dockerfile location and the build mode
     private void initDockerFileFile(Logger log) {
-        if (dockerFile != null) {
+        if (dockerFile != null && dockerFileDir != null) {
+            dockerFileFile = new File(dockerFileDir, dockerFile);
```
rhuss: Could you please add a check, that `dockerFile` is not an absolute path and throw an error if it is ?

lmesz:
Added check for absolutepath that you assked :)

lmesz:
May I ask if it will be merged or not?

rhuss:
@lmesz sorry, I'm bit work overflowed, but I will add it asap when I have the next free cycles. I've one conference next week, but after this things should calm down considerably.

rhuss:
@lmesz next week I have dedicated some cycles to docker-maven-plugin and will integrate you PR then. Thanks for your patience ...

lmesz:
Sure

rhuss:
@lmesz I just integrated PR #645 which conflicts with this PR, so that I will resolve the conflict (and will update the documentation on the go.

fyi, the documentation can be found in `src/main/asciidoc` and uses Asciidoc to generate the user manual.

PR #630, opened 2016-12-01
Keep envs

mfra:
make overwriting behavior in `ContainerCreateConfig` of envs property file optional by setting a `keepEnvs` property

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
+++ b/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
@@ -1,60 +1,62 @@
 package io.fabric8.maven.docker.access;
 
-import java.io.*;
-import java.util.*;
-
+import io.fabric8.maven.docker.config.Arguments;
 import io.fabric8.maven.docker.util.EnvUtil;
 import org.apache.commons.lang3.text.StrSubstitutor;
-import io.fabric8.maven.docker.config.Arguments;
 import org.json.JSONArray;
 import org.json.JSONObject;
+
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.*;
 
 public class ContainerCreateConfig {
 
     private final JSONObject createConfig = new JSONObject();
     private final String imageName;
 
-    public ContainerCreateConfig(String imageName) {
+    public ContainerCreateConfig(final String imageName) {
         this.imageName = imageName;
         createConfig.put("Image", imageName);
     }
 
-    public ContainerCreateConfig binds(List<String> volumes) {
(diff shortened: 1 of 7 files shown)


rhuss:
Thanks a lot and sorry for the delay ;-( I will have a review next week.

rhuss:
Just a quick question: What exactly is your use case ? i.e. why do you need to override the environment from the properties ? Maybe we should make this the default behaviour if it makes sense ? (of course with a minor version upgrade)

rhuss:
Maybe we can introduce something like:

```xml
<envExt>
  <file>....</file>
  <priority>FILE|CONFIG</priority>
  <vars>
     <....>...</...>
  </vars>
</envExt>
```

with keeping `<env>` as a simplified version and deprectating `<envPropertyFile>`. This would keep the configuration concise without adding to much top-level configs in the run configuration which is already quite large. That way we could also add a `<priority>` enum of `ENV` which takes it from an environment variable defined outside of the build (but needs to be declared in the file of `<vars>` with a dummy value to be picked up)

(`<envExt>` is derived from `<fromExt>` which we already use to an extended from definition)

mfra:
Hi,

cause of that CR is, that apparently the docker API doesn't interpolate given env variables with host's values, if there is just a variable set. We currently use that configuration heavily on user specific values, like a username or machine's name.
https://docs.docker.com/compose/environment-variables/#/passing-environment-variables-through-to-containers

In order to use maven's interpolation mechanics like `<ORACLE_JDBC_USER>${env.ORACLE_JDBC_USER}</ORACLE_JDBC_USER>`, I think it is a proper way for users to define which source of envs to use.

I like your idea of refactoring/introducing the `envExt` section.

rhuss:
I think I understand your use case: You want to include properties from a file by default, but want also to partially overwrite the values therein by declaring them in the `pom.xml` for letting a user overwrite them for her environment. Right ? 

Of course one could also add the feature, that properties with the same name as enviroment variables could be overwritten directly by the environment variable (so that you don't need to add them again in the pom.xml), e.g. like in

```xml
<envExt>
  <file>....</file>
  <priority>FILE|CONFIG|HOST_ENV</priority>
  <vars>
     ....
   </vars>
</envExt>
```

so when `HOST_ENV` is given, then this direct host environment would kick in (and you don't need to use the `<vars> ... </vars>` section.

If you like, you could work on this, but I could take this over, too (however since I'm usually quite busy, I can't give any ETA on this feature).

mfra:
Hi,
feel free to take over. I think it'll be easer for you to change the configuration structure.
Don't hurry, will wait for your changes anyway.

rhuss:
@mfra I opened an issue #681 for this new feature to track it. So I will close this PR for now. Thanks for raising this issue.

PR #632, opened 2016-12-01
[resolves #631] Allow commands to run as root

### LINKED ISSUE #631: Allow commands to run as root
We have a user case where the wildfly base image does not run on openshift because the user 'jboss' is not in group 'root'

The generated Dockerfile should perhaps place the USER directive at the end and run entries and RUN commands as root.

I'll would provide a PR

CrossRef: https://github.com/wildfly-extras/wildfly-camel/issues/1501

tdiesler:


src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
```diff
@@ -189,10 +189,10 @@ private void addEntries(StringBuilder b) {
             String[] userParts = StringUtils.split(assemblyUser, ":");
             String userArg = userParts.length > 1 ? userParts[0] + ":" + userParts[1] : userParts[0];
             String chmod = "chown -R " + userArg + " " + tmpDir + " && cp -rp " + tmpDir + "/* / && rm -rf " + tmpDir;
-            if (userParts.length > 2) {
+            if (userParts.length > 2 || user != null) {
```
rhuss: Its not clear that the user always want to run the commands which follow under root and not under the user of the base image. 

Also this implementations relies on that `addEntries()` is called before `addRun()` which is not really mandatory and might change. Its probably better that the user detection happens outside (and is feed in addEntries() and addRun() as arguments, or is replicated in in `addRun()` where the run user could be extracted from the assemblyUser's third part. The problem I have here is that a lower level configuration or unrelated configuration from the `<assembly>` influences the behaviour of another config option `<run>`. IMO it would be much simpler and easier to understand if the run user can be configured independently. 

E.g. 

```
<runCmds>
  <user>root</user>
  <run>....</run>
  <run>....</run>
<runCmds>
```

a heterogenous configuration like this is a bit tricky in Maven, but can be achieved a described in http://blog.sonatype.com/2011/03/configuring-plugin-goals-in-maven-3/ (look for 'Inlined collections').

tdiesler:
Our of Date

Issue #637, opened 2016-12-09
AWS ECR login credential exchange


chonton:
[AWS ECR](http://docs.aws.amazon.com/AmazonECR/latest/userguide/Registries.html) requires a [cli login](http://docs.aws.amazon.com/cli/latest/reference/ecr/get-login.html) before any docker image push or pull.  I currently have my jenkins job running the `aws ecr get-login` command.  However, I prefer not install the amazon toolkit on all my build infrastructure and require all our developers to also install the amazon toolkit.

I would like to submit a PR that would invoke the [GetAuthorizationToken api](http://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetAuthorizationToken.html) to exchange AWS IAM credentials for temporary ECR credentials.  This PR would augment the AuthConfigFactory.  Any login credentials specified for an ECR repository would be considered IAM credentials needing exchange.

An ECR repository has a URL of form **https://**_awsAccountId_**.dkr.ecr.**_awsRegion_**.amazonaws.com**/_repositoryName_

1. Should credentials explicitly specified with docker.* properties be exchanged?
2. Should credentials from ~/.config/kube be exchanged?
3. Should credentials provided in plugin configuration be exchanged?
4. Should credentials from .m2/settings be exchanged?
5. Should credentials from ~/.docker/config.json be exchanged?
6. Should there be a flag to prevent the exchange?

I'm thinking that credentials 1, 2, 3, 4 should be exchanged.  ~/.docker/config.json likely contains already exchanged credentials. ~/.config/kube likely contains IAM credentials, as kubernetes ["will fetch and periodically refresh ECR credentials"](http://kubernetes.io/docs/user-guide/images/#using-aws-ec2-container-registry) .

rhuss:
Thats a good idea.

If I understand you right, d-m-p would do an exchange to obtain an AWS token when it detects that in [pulls](https://github.com/fabric8io/docker-maven-plugin/blob/0912d1ffa3861ba8ca8c7b8e17a999a724487752/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java#L360-L360) or [pushes](https://github.com/fabric8io/docker-maven-plugin/blob/0912d1ffa3861ba8ca8c7b8e17a999a724487752/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java#L374-L374) ECR is used. 

This token is then used as password, right ?

[...]

PR #638, opened 2016-12-09
#637 - outline of credentials exchange

chonton:
Here's an outline of the changes to the mojos.  Todo is the actual exchange.
I'd appreciate your feedback if this is the right behavior, best property names, etc.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -140,6 +140,12 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
     @Parameter(property = "docker.registry")
     protected String registry;
 
+    /**
+     * Skip exchanging local iam credentials for ecr credentials
+     */
+    @Parameter(property = "docker.exchange.skip")
+    protected boolean skipExchange;
+
     // maximum connection to use in parallel for connecting the docker host
     @Parameter(property = "docker.maxConnections", defaultValue = "100")
     private int maxConnections;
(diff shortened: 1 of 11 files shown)


src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -140,6 +140,12 @@
     @Parameter(property = "docker.registry")
     protected String registry;
 
+    /**
+     * Skip exchanging local iam credentials for ecr credentials
+     */
+    @Parameter(property = "docker.exchange.skip")
```
rhuss: I would call it something related to ECR as this feature is connected to the way how AWS does authentication (or is this useful in another context, too ? Mayeb GCR does something similar, but haven't checked).

Also the common schem is `docker.skip.something`, so for the property name I'd suggest

```java
@Parameter(property = docker.skip.ecrAuth)
protected boolean skipEcrAuth;
```

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -385,12 +391,12 @@ protected PomLabel getPomLabel() {
     }
 
 
-    protected AuthConfig prepareAuthConfig(ImageName image, String configuredRegistry, boolean isPush)
+    protected AuthConfig prepareAuthConfig(ImageName image, String configuredRegistry, boolean isPush, boolean skipExchange)
```
rhuss: can we use `skipEchange` directly from member var ? This wouldn't pollute the the argument list that much (which always tend to grow unbounded ;)

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -65,6 +65,62 @@ public AuthConfigFactory(PlexusContainer container) {
...
+     * @param authConfigMap String-String Map holding configuration info from the plugin's configuration. Can be <code>null</code> in
+     *                   which case the settings are consulted.
+     * @param settings the global Maven settings object
+     * @param user user to check for
+     * @param registry registry to use, might be null in which case a default registry is checked,
+     * @return the authentication configuration or <code>null</code> if none could be found
+     *
+     * @throws MojoFailureException
+     */
+    public AuthConfig createAuthConfig(boolean isPush, boolean skipExchange, Map authConfig, Settings settings, String user, String registry)
```
rhuss: as mentioned would avoid 'exchange' for the wording as its really a technical. Its ECR which is important here.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -65,6 +65,62 @@ public AuthConfigFactory(PlexusContainer container) {
...
+            throws MojoExecutionException {
+
+        AuthConfig ret = createAuthConfig(isPush, authConfig, settings, user, registry);
+        if (ret != null) {
+            if (!skipExchange) {
+                return exhangeCredentials(registry, ret);
```
rhuss: typo

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
rhuss: so, maybe `skipExtendedAuth` is even better ....

PR #640, opened 2016-12-11
Clarify log message when waiting for URL

hwellmann:
The log message when waiting on a URL with default settings is a bit confusing:

```
Waiting on url http://localhost:33068/foo/ with method null for status null
```

WIth this change, the message now includes the default HTTP method and the default
status code range:

```
Waiting on url http://localhost:33068/foo/ with method HEAD for status 200..399
```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/WaitConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/WaitConfiguration.java
@@ -11,6 +11,15 @@ import org.apache.maven.plugins.annotations.Parameter;
  */
 public class WaitConfiguration implements Serializable {
 
+    // Default HTTP Method to use
+    public static final String DEFAULT_HTTP_METHOD = "HEAD";
+
+    // Default status codes
+    public static final int DEFAULT_MIN_STATUS = 200;
+    public static final int DEFAULT_MAX_STATUS = 399;
+
+    public static final String DEFAULT_STATUS_RANGE = String.format("%d..%d", DEFAULT_MIN_STATUS, DEFAULT_MAX_STATUS);
+
     @Parameter
     private int time;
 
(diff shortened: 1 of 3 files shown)


davsclaus:
@rhuss lgtm

davsclaus:
ok to test

rhuss:
thanks all, lgtm too

davsclaus:
okay lets merge it then

[merge]

rhuss:
oops, forgot the merge ;-) thanks @davsclaus ....

davsclaus:
Oh @rhuss just noticed it was merged into the `integration` branch. I would have thought this should go to `master`. And btw what is the `integration` branch used for?

rhuss:
The `integration` branch is used to integrate PRs anf is merged into master just before a release. This is still the 'traditional' d-m-p workflow, because the documentation is generated directly from master on every push and hence could describe features which are not yet released. This caused some confusion in the past.

So by default I merge every PR into the `integration` branch when approved as I mostly have to update the changelog.md, too.

However I'm open to align the workflow to the other fabric8 projects and maybe create a dedicated 'documentation' branch for the reference manual updates. This branch would then be rebased before eacht release.

Or, as soon @kameshsampath contribution to our pipeline-library for re-creating the manual we could switch to that to only create documentation during a release build.

rhuss:
Btw, this worklfow is also explained in https://github.com/fabric8io/docker-maven-plugin/blob/master/CONTRIBUTING.md:

> When sending pull request we prefer that to be a single commit. So please squash your commits with an interactive rebase before sending the pull request. Also, your pull request should be rebased to the current branch _integration_. All this is nicely explained [here](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request).

kameshsampath:
@rhuss the pipeline doc is merged into lib, not able to test as our CD server is down

rhuss:
awesome, will switch as soon as things stabilise .....

kameshsampath:
let us check once our CD server is up Monday ;) ??

rhuss:
hope so, but a lot of folks are already off for christmas holidays .... well, let' see ;-)

PR #645, opened 2016-12-13
docker load support.

Katsuya-Tomioka:
Signed-off-by: Katsuya Tomioka <katsuya@basistech.com>


rhuss:
Thanks a lot ! I will have review next week.

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
```diff
@@ -33,6 +33,9 @@
     @Parameter
     private String registry;
 
+    @Parameter
+    private String file;
```
rhuss: Actually I would move this to the `<build>` configuration as it is about *creating images*. There is should be probably called something like `dockerArchive` on the same footprint as `dockerFile` or `dockerFileDir`. Also there should be a check, that only one of `dockerArchive` and `dockerFile` is given.

rhuss:
Ah, yes, and when adding a new build option don't forget to update `PropertyConfigHandler`, too (along with a minor doc update)

Katsuya-Tomioka:
I moved dockerFile part to the build side. Also added a bunch of tests to pass the coverage check. Hope you don't mind. If you could review/edit the doc, that'd be great.

rhuss:
[merge]

PR #648, opened 2016-12-15
Corrected reference to global option 'filter'

tamslinn:
The global option 'image' was renamed to 'filter' but the push section referenced the old name.
Signed-off-by: Tamsin Slinn <tam.slinn@gmail.com>


rhuss:
Thanks !

Issue #660, opened 2016-12-17
Socket connection error


hwellmann:
### Description

With a local build of 0.18-SNAPSHOT from the integration branch, I'm getting the following errors:

```
[INFO] --- docker-maven-plugin:0.18-SNAPSHOT:build (default-build) @ foo-docker ---
[INFO] I/O exception (java.net.SocketException) caught when processing request to {}->unix://127.0.0.1:1: already connected
[INFO] Retrying request to {}->unix://127.0.0.1:1
[INFO] I/O exception (java.net.SocketException) caught when processing request to {}->unix://127.0.0.1:1: already connected
[INFO] Retrying request to {}->unix://127.0.0.1:1
[INFO] I/O exception (java.net.SocketException) caught when processing request to {}->unix://127.0.0.1:1: already connected
[INFO] Retrying request to {}->unix://127.0.0.1:1
[ERROR] DOCKER> Cannot create docker access object
```

This is a regression from 0.18.1.

### Info

* d-m-p version : 1.18-SNAPSHOT 8545c3d94723fbbe6fa8fba8ad5f70a6d60d7114
* Maven version (`mvn -v`) : 3.3.9
* Docker version : 1.12.3

chonton:
I have potentially similar problem.  On a macbook using master branch commit [f7008c3](https://github.com/fabric8io/docker-maven-plugin/commit/f7008c38a9ab08709e1ff51a37569a7931e9152c), I see the following stacktrace when running the samples:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:docker-maven-plugin:0.18-SNAPSHOT:start (start) on project dmp-sample-net: Cannot create docker access object
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
        at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
[...]

PR #661, opened 2016-12-17
Create at most one PrintStream per log file.

### LINKED ISSUE #652: Garbled container logs
### Description

I'm running a container with the following configuration:

```
<run>
    <wait>
        <http>
            <url>http://localhost:${tomcat.http.port}/foo/</url>
        </http>
        <time>600000</time>
        <exec>
            <postStart>initialize.sh</postStart>
        </exec>
    </wait>
    <log>
        <file>${basedir}/backend.log</file>
    </log>
</run>
```

The problem is that the resulting logfile `backend.log` is garbled. It contains the output of my `postStart` command (which I'm not really interested in), followed by a large block of null bytes, followed by some but not all standard output and standard error messages from the container. 

A large chunk of container messages is missing (possibly overwritten by the null bytes).

Using `docker.keepContainer` and

`docker logs mycontainer > out.txt 2> err.txt` 

I can see that the Docker logs are complete, so I believe d-m-p is to blame. 

Is it intended that `postStart` messages automatically go to the log file as soon as there is a `<log>` element? Is there a way to route `postStart`, `preStop` and normal container messages to different sinks (i.e. files or the console)?

In any case, log messages should not be corrupted.

Looking at the source code, it seems that `DefaultLogCallback` is lacking proper synchronization.

### Info

* d-m-p version : 0.18.1
* Maven version (`mvn -v`) : 3.3.9
* Docker version : 1.12.3

hwellmann:
This avoids corrupted log files.

Fixes #652.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
+++ b/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
@@ -154,6 +154,7 @@ public class DockerAccessWithHcClient implements DockerAccess {
                     LineNumberReader reader = new LineNumberReader(new InputStreamReader(stream));
                     String line;
                     try {
+                        callback.open();
                         while ( (line = reader.readLine()) != null) {
                             callback.log(1, new Timestamp(), line);
                         }
--- a/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
+++ b/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
@@ -19,7 +19,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-import io.fabric8.maven.docker.access.hc.util.ClientBuilder;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpResponseException;
 import org.apache.http.client.ResponseHandler;
@@ -40,6 +39,7 @@ import io.fabric8.maven.docker.access.hc.ApacheHttpClientDelegate.BodyAndStatusR
 import io.fabric8.maven.docker.access.hc.ApacheHttpClientDelegate.HttpBodyAndStatus;
 import io.fabric8.maven.docker.access.hc.http.HttpClientBuilder;
 import io.fabric8.maven.docker.access.hc.unix.UnixSocketClientBuilder;
+import io.fabric8.maven.docker.access.hc.util.ClientBuilder;
 import io.fabric8.maven.docker.access.hc.win.NamedPipeClientBuilder;
 import io.fabric8.maven.docker.access.log.LogCallback;
 import io.fabric8.maven.docker.access.log.LogGetHandle;
@@ -159,6 +159,8 @@ public class DockerAccessWithHcClient implements DockerAccess {
                         }
                     } catch (LogCallback.DoneException e) {
                         // Ok, we stop here ...
+                    } finally {
+                        callback.close();
                     }
                 }
                 return null;
--- a/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
+++ b/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
@@ -32,6 +32,7 @@ import io.fabric8.maven.docker.access.DockerAccess;
 import io.fabric8.maven.docker.access.DockerAccessException;
 import io.fabric8.maven.docker.access.NetworkCreateConfig;
 import io.fabric8.maven.docker.access.UrlBuilder;
+import io.fabric8.maven.docker.access.VolumeCreateConfig;
 import io.fabric8.maven.docker.access.chunked.BuildJsonResponseHandler;
 import io.fabric8.maven.docker.access.chunked.EntityStreamReaderUtil;
 import io.fabric8.maven.docker.access.chunked.PullOrPushResponseJsonHandler;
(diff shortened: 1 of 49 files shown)


src/main/java/io/fabric8/maven/docker/log/LogCallbackFactory.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+        PrintStream ps = printStreamMap.get(file);
+        if (ps == null) {
+            ps = new PrintStream(new FileOutputStream(file), true);
+            printStreamMap.put(file, ps);
+        }
+        return ps;
+    }
+
+    public static void closeLogs() {
```
rhuss: This method should be `synchronized`, too

src/main/java/io/fabric8/maven/docker/log/LogDispatcher.java
```diff
@@ -37,15 +40,16 @@ public LogDispatcher(DockerAccess dockerAccess) {
     }
 
     public synchronized void trackContainerLog(String containerId, LogOutputSpec spec) throws FileNotFoundException {
-        LogGetHandle handle = dockerAccess.getLogAsync(containerId, new DefaultLogCallback(spec));
+        LogGetHandle handle = dockerAccess.getLogAsync(containerId, createLogCallback(spec));
```
rhuss: I'm not a a fan of static imports as they dilute the real locations and I always search first in the current class when doing review. Could we refer to the static method via its class (`LogCallbackFactory.createLogBack()`) ? Even saves some lines.

src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -145,7 +146,7 @@ public void startExecContainer(String containerId, LogOutputSpec outputSpec) thr
     }
 
     private ResponseHandler<Object> createExecResponseHandler(LogOutputSpec outputSpec) throws FileNotFoundException {
-        final LogCallback callback = new DefaultLogCallback(outputSpec);
+        final LogCallback callback = createLogCallback(outputSpec);
```
rhuss: Who is closing this DefaultLogCallback() ?

src/main/java/io/fabric8/maven/docker/log/LogDispatcher.java
```diff
@@ -37,15 +40,16 @@ public LogDispatcher(DockerAccess dockerAccess) {
...
         logHandles.put(containerId, handle);
     }
 
     public synchronized void fetchContainerLog(String containerId, LogOutputSpec spec) throws FileNotFoundException {
-        dockerAccess.getLogSync(containerId, new DefaultLogCallback(spec));
+        dockerAccess.getLogSync(containerId, createLogCallback(spec));
     }
 
     public synchronized void untrackAllContainerLogs() {
+        closeLogs();
```
rhuss: This will close all streams from all LogDispatcher instances running. Not sure why it could be more than one, but at least we should avoid the mismatch with the LogCallbackFactory storing the stream in a static Map vs. using object instances for the LogDispatcher here. I would prefer to have LogCallbackFactory to be an object, too.

What's about the following idea: We simply allow `DefaultLogCallback` to reuse an existing printstream (i.e. we do the caching on a static field in DefaultLogCallback along with a usage count) and add a `close()` method to the callback so that the user can signal that he is done with logging ? In `close()` an the associated printstream's usage count will be decreased and if 0 then closed.

I think this kind of resource management makes more sense, also for potential future use cases. If you don't have the cycles to implement this, no problem, I could take over (as this is indeed an important fix to have). Also, a unit tests for `DefaultLogCallback` would be awesome then.

hwellmann:
Thanks for your feedback. I'll get back to this tomorrow...

rhuss:
Awesome,thx unfortunately I just catched an ugly flu, and staying in bed
currently. I will continue here asap. Sorry ...

Harald Wellmann <notifications@github.com> schrieb am Di., 20. Dez. 2016,
13:44:

hwellmann:
Reworked `DefaultLogCallback` with reference counting and added a test class. The `LogCallbackFactory` is gone.

Note: To test this in the context of my real-life project, I had to revert 7afec26aaaa9ebe39065a8daa8a96db42945a22e, see #660. 

Master appears to be broken, I'm seeing the same issue both on MacOS and Linux.

src/main/java/io/fabric8/maven/docker/log/SharedPrintStream.java
```diff
@@ -0,0 +1,35 @@
...
+    public SharedPrintStream(PrintStream ps) {
+        this.printStream = ps;
+        this.numUsers = 1;
+    }
+
+    public PrintStream getPrintStream() {
+        return printStream;
+    }
+
+    public void setPrintStream(PrintStream printStream) {
```
rhuss: Doesnt look like that this method is ever used. Also, removing this set-method makes the class immutable with respect to the printstream which is a good thing :)

src/main/java/io/fabric8/maven/docker/access/log/LogRequestor.java
```diff
@@ -91,6 +91,8 @@ public void fetchLogs() {
             parseResponse(resp);
         } catch (IOException exp) {
             callback.error(exp.getMessage());
+        } finally {
+            callback.close();
```
rhuss: We should ensure, that, when the callback is closed that this any call to the `LogRequestors` method is invalid (by throwing an exception).

Its also not so nice that the responsibilities for creating the callback (== caller of constructor of `LogRequestor`) and the one closing the callback (== the `LogRequestor`) are different objects. It would be much better when the `LogRequestor` has the complete control over allocation and closing of the print stream.

I suggest therefor to introduce a an `LogCallback open()` which allocates the stream instead of automatically allocating it in the constructor. Each method which can close a stream then would have to open it before.

src/main/java/io/fabric8/maven/docker/access/log/LogRequestor.java
```diff
@@ -224,6 +227,7 @@ public void finish() {
                 }
             }
         }
+        callback.close();
```
rhuss: finish() is called in case of an exception but the `callback.close()` is also included in the finally clauses of the top level methods. This leads to closing a callback() twice which is probably not intended.

src/test/java/io/fabric8/maven/docker/log/DefaultLogCallbackTest.java
```diff
@@ -0,0 +1,173 @@
...
+        DefaultLogCallback callback2 = new DefaultLogCallback(spec);
+
+        callback.log(1, ts, "line 1");
+        callback2.log(1, ts, "line 2");
+        callback.log(1, ts, "line 3");
+        callback.close();
+        callback2.log(1, ts, "line 4");
+        callback2.close();
+
+        List<String> lines = Files.readAllLines(path);
```
rhuss: Unfortunately we can't use Java 8 and are still stuck to Java 7 for now (this might change in the future). So we need some other way to read all lines here.

rhuss:
@hwellmann I made some changes and would like to push it to this PR. Could you please grant me push permissions ? This is described [here](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

rhuss:
Sorry, my faul. Mispelled the target branch for the push.

rhuss:
[merge]

PR #654, opened 2016-12-17
Add support for Creating Volumes

raystorm:
This is a new feature to add named volume support.

`volume-create` and `volume-remove`  goals
example config in the samples directory.
simple unit tests added.
Non-automatic, both goals have to be explicitly called.

PR #664, opened 2016-12-17
Add Support for Creating Volumes Take #2

raystorm:
This is a new feature to add named volume support.

added `volume-create` and `volume-remove` goals
example config in the samples directory.
simple unit tests added.
Non-automatic, both goals have to be explicitly called.

PR to Master per @rhuss ask on Pull Request https://github.com/fabric8io/docker-maven-plugin/pull/654

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/ImageVolumeConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/ImageVolumeConfiguration.java
@@ -1,0 +1,95 @@
+package io.fabric8.maven.docker.config;/*
+ *
+ * Copyright 2014 Roland Huss
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.Serializable;
+import java.util.*;
+
+import org.apache.maven.plugins.annotations.Parameter;
+
+/**
+ * Run configuration for volumes.
+ *
+ * @author roland
+ * @since 08/12/14
+ */
+public class ImageVolumeConfiguration implements Serializable {
+
+    /**
+     * List of images names from where volumes are mounted
+     */
+    @Parameter
+    private List<String> from;
+
+    /**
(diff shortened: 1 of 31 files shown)


rhuss:
Thanks !

raystorm:
@rhuss I noticed after I created the pull requests where the documentation was stored. So I updated it. Thank you for your patience.

src/main/java/io/fabric8/maven/docker/config/ImageVolumeConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+public class ImageVolumeConfiguration implements Serializable {
+
+    /**
+     * List of images names from where volumes are mounted
+     */
+    @Parameter
+    private List<String> from;
+
+    /**
+     * List of bind parameters for binding/mounting host directories
```
raystorm: This file is just a rename.  I don't think I actually changed anything in it.

raystorm:
@rhuss,  Documentation / code formatting updated.

rhuss:
Awesome,thx unfortunately I just catched an ugly flu, and staying in bed
currently. I will continue here asap. Sorry ...

Tom Burton <notifications@github.com> schrieb am Di., 20. Dez. 2016, 16:34:

rhuss:
Thanks a lot, I will integrate your PR soon to master.

While thinking about it and comparing it to the way how we deal with custom networks, I think it might be as well a good addition to *implicitly* create (during `docker:start`) and delete (during `docker:stop`) volumes on the fly when referenced in a custom configuration and when they not already exists. If there would be a configuration, it would be picked up, otherwise volumes with default driver and options would be created. This would allow an even better UX since you don't have to run `volume-create` before `docker:start` when calling the goals manually.

Also, since we have [Docker compose integration](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeConfigHandler.java) it would be awesome, to pickup the  `volumes` section from a given docker-compose file as configuration. 

Finally an integration in the [external property handler](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/config/handler/property/PropertyConfigHandler.java) would be nice, too.

I suggest, that I will open issues for these and we can work on this after the next release. If you feel fancy, you could work on this, too. (I would give you the relevant pointers then). would be awesome ;-)

raystorm:
I was doing some further work and testing with my current implementation of the volume support.
Im not sure if its me or docker, but it's not properly passing driverOptions to docker to set permissions even though it works when I directly use the options on the command line to docker.

Would you like to merge and follow up with an issue or hold off merging first?

rhuss:
Thanks for the info, will try to check this. But in this case (also because time is running away ;-) I suggest to merge and fix it it later.

rhuss:
merged updated version #669 into master

PR #672, opened 2016-12-23
Revert "Upgrade to jnr-unixsocket 0.15"

rhuss:
7afec26aaaa9ebe39065a8daa8a96db42945a22e

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,8 +2,7 @@
 
 * **0.18.2** 
   - Better log message when waiting for URL (#640)
-  - Update to jnr-unixsocket 0.15
-  - Add two new goals: "volume-create" and "volume-remove" for volume handling independent of images. 
+- Add two new goals: "volume-create" and "volume-remove" for volume handling independent of images. 
 
 * **0.18.1** (2016-11-17)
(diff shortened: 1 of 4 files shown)


rhuss:
This reverts the upgrade to jnr-unixsocket 0.15 which seems to break d-m-p. See #660.

Unfortunately I don't have time yet to investigate this issue more, so let's go back to the old version which works quite stable most of the time.

Fixes #660.

rhuss:
[merge]

jgangemi:
fyi, i'm going to spend some time on this starting in january and perhaps investigate another library entirely. i keep hitting conditions where the `stop` for a single container hangs the build which is no bueno. 😄

jdve:
@jgangemi have you had any success?  i'm also facing this hanging-on-stop problem regularly.  i'd love to hear if you've had any leads/breakthroughs!

jgangemi:
unfortunately, no. `jnr-unixsocket` is the only thing that's even remotely maintained so it's going to be the best bet and i haven't had enough time to deep dive into the library.

a co-worker reminded me that i could get the ip address of an aws instance through the metadata service, so i've worked around the issue by opening the tcp socket and going going that route (i'm running builds inside kubernetes but still want maven to control spinning up containers).

Issue #205, opened 2015-06-26
Enhance direct Dockerfile support


rhuss:
The following features should be added for a better experience when using Dockerfiles directly:

* It should be possible to support multiple Dockerfiles (as Docker 1.6 does). The idea is to use `dockerFile` instead of `dockerFileDir` and the base directory is then set to the directory of this Dockerfile (if `dockerFileDir` is used, then the name of the dockerfile is supposed to be `Dockerfile`)
* Allow variable substitution within Dockerfiles (as for other Maven resources).
* Allows special variables for:
  - `ADD_ASSEMBLY` will add the proper ADD statements for the assmbly
* Allow the configuration of the assembly directory used during the build. See also #203 
* Optional: Allow optimizations like compressing all subsequent RUN statements into a single RUN statement or so.

rhuss:
Yet another idea:

After the image has been build one could compress it into a single layer by:

* Starting a container with `/bin/true`
* Using `export` to get a tar-stream and ...
* ... pipe that directly to `create` with query option `fromSrc=-`
* Remove the initial image
* Name the new image like the old image

Quite involved, but at the end we would have a nicely squeezed image which is especially useful if this image would be used as a based image. 

Whether this is worth to implement is discussable, though ;-)

twz123:
Do you know https://github.com/jwilder/docker-squash ?

rhuss:
Yeah, but never tried it (and forgot about it again ;-). Thanks for the heads up. 

Might be an even more advance algorithm one could look at. However, with the plugin we are restricted for things which can be done via the Docker API (don't know whether docker-squash can do it)

vberruchon:
Hi Roland, I can see here in the envisaged features :
<i>"Allow variable substitution within Dockerfiles (as for other Maven resources)."</i>

I'd like to use variables defined in my pom.xml in my Dockerfile?

After reading the documentation I believed that the <env> defined vars would have been available for the docker build? Seems it's not true?
Variable substitution would be the only way to have this?
Or would you have an idea to get this.

For example I'd like to use the ${project.version} value in the build process based on my Dockerfile...

Possibility to use "dockerFile" instead of dockerFileDir would be great too.

Thank you
Vincent

rhuss:
Hi Vincent, 

yes, exactly this feature is planned, but not yet implemented. You should be able to use Maven properties in a Dockerfile. This will be in one of the next releases, shouldn't be to hard of course. 

[...]

PR #686, opened 2017-01-06
Dockerfile template issue 205

pgier:
Adds a new parameter to the "build" goal called "dockerFileTemplate".  This can be used to allow variable substitution into an existing Dockerfile.  Variable names are delimited by a "%%" at the start and end, and the plugin will attempt to find a matching property.
Related to issue #205, but only implements the variable substitution part.

### DIFF
--- a/src/main/asciidoc/inc/build/_configuration.adoc
+++ b/src/main/asciidoc/inc/build/_configuration.adoc
@@ -1,7 +1,7 @@
 
 All build relevant configuration is contained in the `<build>` section
-of an image configuration. In addition to `<dockerFileDir>` and
-`<dockerFile>` the following configuration options are available:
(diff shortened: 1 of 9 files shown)


rhuss:
Thanks a lot ! I will have a look soon, however a bit busy these weeks (including this ;-)

mseashor:
+1

rhuss:
Thanks a lot for the PR. However, we decided to select a different path for including filtering for external Dockerfiles. This is now enabled by default and you can tune the delimiters. Please check https://dmp.fabric8.io/#build-filtering and #777 for details.

Its included in 0.21.0 which just has been released.

Thanks for your patience, please let me know whether this works for you. In the meantime I'll close this PR.

pgier:
Ok, thanks for the update, I think the new filtering option works for me.

PR #687, opened 2017-01-08
docker-save support 

Katsuya-Tomioka:
I wrote to specify which one (and where) to save rather than go over build configs as in 'source'. Also not sure about where to autopull. Seemed too much in this context. Let me know.


rhuss:
Thanks for the PR and sorry for the delay ! I hope I will have a chance to look at it soon.

rhuss:
Sorry again for the enhanced delay, but should be added soon ;-)

Your PR looks good, I tuned it a bit. Especially its allowed now to run `mvn docker:save` with the need to provide any additional arguments as the plugin select some sane defaults:

* If no `name` or `alias` is given, then the configuration is examined if there is only a single image with a build configuration. If this is the case, this image's name is used.
* If no `file` is given, then the filename is calculated based on the alias or name. Currently the alias has prio, but maybe its better to use always the name. wdyt ?

What is still missing is documentation. As soon as we have this, we can merge it. When you feel fancy, feel free to submit some documentation. But I can update it, too.

Also some extra unit tests would be awesome (e.g. for `ArchiveCompression` which already was there but which I enhanced to cover the use cases introduced by SaveMojo).

Katsuya-Tomioka:
I'd probably prefer `name` over `alias` as default. $name-$alias.tar.gz if alias is available?

rhuss:
Thanks a lot, I will have a look at the weekend and will integrate the PR.

rhuss:
> I'd probably prefer name over alias as default. $name-$alias.tar.gz if alias is available?

In any other usage scenarios `alias` typically takes precedence, so we should probably use this here, too.

Also, we need to sanitise the name as it can contain `/`, too (names with more than two segments are allowed, too).

src/main/java/io/fabric8/maven/docker/SaveMojo.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	private void attachSaveArchive() {
+		File fileObj = new File(file);
+		if (fileObj.exists()) {
+            String type = FilenameUtils.getExtension(file);
+            if (classifier != null) {
+                projectHelper.attachArtifact(project, type, classifier, fileObj);
+            } else {
+                projectHelper.attachArtifact(project, type, fileObj);
```
rhuss: We should not do this here, as it will _replace_ the main artefact of the build when no classifier is used.

From the JavaDoc:

> Add or replace an artifact to the current project.

and with classifier == null it will replace the main artefact.

I suggest to remove this feature for now and, if needed for a specific use case, introduce it again.

src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
```diff
@@ -65,6 +65,11 @@ public String deleteImage(String name, boolean force) {
                 .build();
     }
 
+    public String getImage(ImageName name) {
+        return u("images/%s/get", name.getNameWithoutTag())
```
rhuss: I don't think we should use the name without tag because according to the Docker reference manual in that case all image of all tags are returned. Actually we should use the full image name here.

rhuss:
I updated the PR according to my comments, if you don't mind.

Going to merge this into master now, we can still tune it before doing the next release.

Thanks a lot for you contribution ! 'hope my updates are fine for you.

rhuss:
[merge]

PR #701, opened 2017-02-08
Refactoring

nicolaferraro:
@rhuss this is a refactoring to allow moving the watch code in f-m-p out of the mojo ([issue 799](https://github.com/fabric8io/fabric8-maven-plugin/issues/799)).

I moved the watch code into a `WatchService`. Since the watch code used to issue builds at some time, I had to move some other code into the existing `BuildService`. The `AuthService` helped to put some shared code out of the mojo.

The `DockerAccessFactory` object is an attempt to allow creating a docker access object without the need to inherit from the Mojo. It's hard to break the inheritance of 'f-m-p' mojos from 'd-m-p' mojos, but this abstraction eventually can help during tests.

I have a working version of 'f-m-p' that is compatible with this refactoring here: https://github.com/nicolaferraro/fabric8-maven-plugin/tree/watchers (`Watcher` abstraction already created and code moved to a specific watcher, need to refactor the 'spring-boot' one, but I have to fix it before moving the code).


rhuss:
Thanks a lot ! I will have a review ASAP, hopefully this afternoon.

nicolaferraro:
@rhuss perfect, thanks!

src/main/java/io/fabric8/maven/docker/service/DockerAccessFactory.java
```diff
@@ -0,0 +1,227 @@
...
+/**
+ *
+ */
+@Component(role = DockerAccessFactory.class, instantiationStrategy = "singleton")
+public class DockerAccessFactory {
+
+    // Minimal API version, independent of any feature used
+    public static final String API_VERSION = "1.18";
+
+    public DockerAccess createDockerAccess(DockerAccessContext dockerAccessContext, Logger log) throws MojoExecutionException, MojoFailureException {
```
rhuss: Guess this is what you need for f-m-p to create an own Access object, right ?

Makes much sense to me. Wonder whether we should call `DockerAccessContext` better `DockerAccessConfig` (as it is what it really is, a read-only configuration object without logic). But I'm fine with, context, too. Maybe we could even stuff the Logger into it then.

src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -291,9 +316,11 @@ public StartedContainer call() throws Exception {
             //String imageName = new ImageName(imageConfig.getName()).getFullNameWithTag(registry);
 
             String imageName = imageConfig.getName();
-            checkImageWithAutoPull(hub, imageName,
+
+            BuildService.BuildContext buildContext = getBuildContext();
```
rhuss: I see, that's the reason why `getBuildContext()` is in the AbstractDockerMojo.

However, I'm not sure if we would better move `checkImageWithAutoPull()` out of the BuildService. Doesn't sound very "build-y" and there is probably a better place for it, so that the `StartMojo` and the `BuildMojo` could both use it ?

rhuss:
@nicolaferraro hope the comments are ok so far. I will look into some other PRs for d-m-p now and plan a release tomorrow or over the weekend.

nicolaferraro:
@rhuss I was working on refactoring `f-m-p` and missed the review. I'm going to look at your comments and apply changes, thanks.

rhuss:
np :) happens to me all the time :) (thats why the extra ping). Also seems like sometimes GitHub mail notifications never make it to my mail box, so don't hesitate to bug me when I don't act :)

nicolaferraro:
Thinking about the `authService` that shouldn't exist and the `pull` stuff that didn't fit in the `Build` service, I realized that we needed just a `RegistryService` that handles the auth against the registry (now a private method) and allow pushing/pulling images (I've changed also the `PushMojo`, now it's a 1-liner :D). The RegistryConfig is completely defined in the `AbstractDockerMojo`, and the `BuildContext` now is in the build support mojo.

Hope it's ok. I'll be away tomorrow but I'll connect later and in the weekend (I'll open the related PR in f-m-p tomorrow morning). 

Thanks for your precious comments :)

rhuss:
On a brief look it looks fine, going to merge it tomorrow. 

I wonder whether we should consider to move the refactored part out into a dedicated project, so that both d-m-p and f-m-p can rely on it and we can loosen the coupling between d-m-p and f-m-p further. Not sure whether we can get rid of inheritance completely, but that would be awesome.

src/main/java/io/fabric8/maven/docker/BuildMojoNoFork.java
```diff
@@ -51,8 +52,13 @@ protected void executeInternal(ServiceHub hub) throws DockerAccessException, Moj
...
     protected void buildAndTag(ServiceHub hub, ImageConfiguration imageConfig)
-        throws MojoExecutionException, DockerAccessException {
-        buildImage(hub, imageConfig);
+            throws MojoExecutionException, DockerAccessException {
+
+        EnvUtil.storeTimestamp(getBuildTimestampFile(), getBuildTimestamp());
+
+        BuildService.BuildContext buildContext = getBuildContext();
+
+        hub.getBuildService().pullAndBuildImage(imageConfig, buildContext);
```
rhuss: would call it only 'buildImage' as the potential autopull of the base image is more an implementation detail.

src/main/java/io/fabric8/maven/docker/util/Task.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+
+/**
+ * Represents a generic task to be executed on a object.
+ */
+public interface Task<T> {
+
+    void execute(T object) throws DockerAccessException, MojoExecutionException, MojoFailureException;
```
rhuss: Would not throw Mojo exceptions here. Actually the more we can separate maven related and non-maven related code already, the easier will be the next step to separate those code (whether we will actually do it is another question ;-). But I don't think these exceptions are needed here (at least not in the single implementation in d-m-p. Maybe some f-m-p implementations ?)

The ideal goal would be to have all services independent of Maven and the parts which are needed from Maven then hidden behind interfaces (where we then implement those as adapter in the "maven" layer). But its more a long term goal (still not sure whether it's worth the effort)

src/test/java/io/fabric8/maven/docker/service/LoadImageTest.java
```diff
@@ -43,6 +48,9 @@
     @Injectable
     private ArchiveService archiveService;
 
+    @Injectable
+    private RegistryService authService;
```
rhuss: would be better called `registryService` or removed it if not needed.

nicolaferraro:
Agree with the comments.. going to implement changes.

rhuss:
[merge]

rhuss:
[merge]

PR #713, opened 2017-02-15
Removed build-nofork and source-nofork

rhuss:
Instead the missing artifact file (which is set by the 'package' packager and hence must run before docker:build) is calculated by the DockerAssemblyManager itself.

This is not bullet proof (e.g. doesnt work when using classifier for the jar) but should work in the majority of use cases.

The error message has been improved for the case when this still not works. As before one can avoid the problem by calling 'mvn package docker:build'.

BTW, the reason for this (non-changeable) Maven behaviour is described in https://issues.apache.org/jira/browse/MASSEMBLY-94


rhuss:
retest this please

rhuss:
@nicolaferraro time for a quick review ;-) ? if not no problem, would then merge it then.

I wanna add this to this d-m-p release because this issue is constantly bugging us (also for f-m-p).

nicolaferraro:
Sure!

rhuss:
@nicolaferraro Thanks ;-) the most interesting part is this method which tries to calculate the Artifact.file when not present --> https://github.com/fabric8io/docker-maven-plugin/pull/713/files#diff-dc400306e02b5f3fd827516ead371e09R384

This is normally the job of the packaging plugin (jar, war, ear, ...) which runs before build (if called or bound to an execution).

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -346,19 +349,71 @@ private void createAssemblyArchive(AssemblyConfiguration assemblyConfig, MojoPar
...
                                             + e.getMessage());
+        } finally {
+            setArtifactFile(params.getProject(), originalArtifactFile);
+        }
+    }
+
+    // Set an artifact file if it is missing. This workaround the issues
+    // mentioned first in https://issues.apache.org/jira/browse/MASSEMBLY-94 which requires the package
+    // phase to run so set the ArtifactFile. There is no good solution, so we are trying
+    // to be very passive and add a workaround for some situation which won't work for every occasion.
```
rhuss: "defensive" is the better word

doc/changelog.md
```diff
@@ -1,10 +1,13 @@
 # ChangeLog
 
-* **0.19.2** 
+* **0.20.0** 
 
   - Fix handling of `run` commands from properties (#684)
   - Fix empty `<link>` causing `NullPointerException` (#693)
-  
+  - Removed `build-nofork` and `source-nofork` in favor for a more direct solution which prevents forking of the lifecycle. Please refer the documentation, chapter "Assembly" for more information about this.
```
rhuss: yeah, in fact. As soon as we have released we should update f-m-p, too

src/main/asciidoc/inc/_docker-build.adoc
```diff
@@ -3,7 +3,7 @@
 == *docker:build*
 
 This goal will build all images which have a `<build>` configuration section, or, if the global configuration variable `filter` (property:
-`docker.filter`) is set, only the images contained in this variable (comma separated) will be built.
+`docker.filter`) is set, only the images contained in this variable (comma separated) will be built. `docker:build` is meant to be run from the command line and will automatically compile and package your application.
```
rhuss: gotcha, good point. Left over from a previous attempt ;). will clean that up.

rhuss:
@nicolaferraro thanks for the review. Actually its not about multi module projects, the problem occurs also for single modules when you use an assembly which involves the build artifact.

When you run 'mvn package docker:build', then the packager runs (whatever your packaging, e.g. jar) and sets on the current `MavenProject`'s `Artifact` the file path to the generated artifact (Maven is quite messy where it sets state). The `docker:build` use an assembly manager which requires this path to include the artifact. If its not set, then it first prints a warning and if the artifact was the only thing to include into the assembly then it fails with an error.

When you only run `mvn docker:build` no one sets this file, leading to the error condition above.

When you fork docker:build to the package phase then:

* Maven deeply clones the `MavenProjects` and runs the lifecycle up to `package`. `package` set the file in this cloned object.
* Then it continues with `docker:build` but with the original MavenProject where the file is not set.

So forking has no effect at all here (my faults, haven't test this properly ;-(

That the reason why I try to set this file on my own when it is not set.

nicolaferraro:
Ok, understood. I think that in this case it's better to have independent maven goals that can be combined without forking new lifecycles.

:+1:

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -346,19 +349,71 @@ private void createAssemblyArchive(AssemblyConfiguration assemblyConfig, MojoPar
...
+        String finalName = build.getFinalName();
+        String target = build.getDirectory();
+        if (finalName == null || target == null) {
+            return null;
+        }
+        File artifactFile = new File(target, finalName + "." + project.getPackaging());
+        if (artifactFile.exists() && artifactFile.isFile()) {
+            // TODO: Maybe check also that the artifact file is 'young' ? (so that it could be assumed that
+            //       it has been created recently by a 'package' in the same run.)
+            setArtifactFile(project, artifactFile);
```
rhuss: actual not, because the return value is used as the old value, to reset the file again. Since its was already null before, null is the proper return value.

nicolaferraro: right :)

rhuss:
[merge]

PR #719, opened 2017-02-17
Add healthy wait condition that waits until a container becomes healthy

### LINKED ISSUE #709: Add Wait-Support for docker 2.12 HEALTHCHECK
* d-m-p version : 0.19.0
* Maven version (`mvn -v`) : 3.3.9 (any)
* Docker version : 1.13.0

* If it's a feature request, what is your use case :
d-m-p introduces support for the HEALTHCHECK (see #594) in d-m-p build but has no effect in d-m-p run. It would be great to have a similar mechanism like docker compose `depends_on` `condition: service_healthy` (see https://docs.docker.com/compose/compose-file/#/dependson) to delegate the actual healthy checking to the docker engine.

One option would be a "wait" command "healthy" that waits until the started container becomes healthy according to the docker host status.

danielwegener:
according to its configured HEALTHCHECK (fixes #709).

Signed-off-by: Daniel Wegener <daniel@wegener.me>


danielwegener:
The actual healtcheck test is extracted from the docker inspect container configuration and displayed in the log file.

Example output:

```
[INFO] Building Fubibabi 7.0.0-SNAPSHOT
[INFO] 
[INFO] DOCKER> [my-docker-snapshot.com/core:7.0.0-SNAPSHOT] "cep-core": Start container fbcbb1469a6b
[INFO] DOCKER> [my-docker-snapshot.com/core:7.0.0-SNAPSHOT] "cep-core": Waiting on healthcheck: '"CMD-SHELL", "curl -sf http://localhost:8080/core/status || exit 1"'
[INFO] DOCKER> [my-docker-snapshot.com/core:7.0.0-SNAPSHOT] "cep-core": Waited on healthcheck '"CMD-SHELL", "curl -sf http://localhost:8080/core/status || exit 1"' 30337 ms
[INFO] BUILD SUCCESS
[INFO] Total time: 41.720 s
[INFO] Finished at: 2017-02-17T11:39:36+01:00
[INFO] Final Memory: 43M/495M
```

danielwegener:
@rhuss WDYT? Too verbose? should we really log the internally configured (and possibly longish) Healthcheck command?

rhuss:
thanks for the PR ! let me check this later, still busy a bit ...

rhuss:
the PR is not forgotten, plan some integration session on the weekend ;-) ....

src/main/asciidoc/inc/start/_wait.adoc
```diff
@@ -41,6 +41,11 @@ a| TCP port check which periodically polls given tcp ports. It knows the followi
 * *mode* can be either `mapped` which uses the mapped ports or `direct` in which case the container ports are addressed directly. In the later case the host field should be left empty in order to select the container ip (which must be routed which is only the case when running on the Docker daemon's host directly). Default is `direct` when host is _localhost_, `mapped` otherwise. The direct mode might help when a so called _user-proxy_ is enabled on the Docker daemon which makes the mapped ports directly available even when the container is not ready yet.
 * *host* is the hostname or the IP address. It defaults to `${docker.host.address}` for a mapped mode and the container ip address for the direct mode.
 * *ports* is a list of TCP ports to check. These are supposed to be the container internal ports.
+
+| *healthy*
+a| Check that waits until the container health state becomes `healthy`. A container is considered healthy when its <<build-healthcheck,  configured healtcheck>> succeeds.
+
+This behaviour mirros the docker compose dependsOn `condition: service_healthy`.
```
danielwegener: ... or maybe better: _mimics_

rhuss: yeah, sounds better ;-)

samples/healthcheck/pom.xml
```diff
@@ -39,6 +39,11 @@
                   <cmd>curl -f http://localhost/ || exit 1</cmd>
                 </healthCheck>
               </build>
+              <run>
+                <wait>
+                  <healthy>true</healthy>
```
rhuss: I wonder what `<healthy>false</healthy>` would imply ;-) 

I know, Maven doesn't allow empty tags, so that's probably fine.

src/main/java/io/fabric8/maven/docker/model/InspectedContainer.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Interface representing an inspected container
+ *
+ * @author daniel
+ * @since 17/02/15
+ */
+public interface InspectedContainer extends Container {
```
rhuss: What is the benefit to extend `Container` over simplify adding the method directly to `Container` ?

Please add also some javadocs to public interface methods.

src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
```diff
@@ -121,6 +126,21 @@ public boolean isRunning() {
...
 
+    @Override
+    public boolean isHealthy() {
+        final JSONObject state = json.getJSONObject(STATE);
+        // always indicate healthy for docker hosts that do not support health checks.
+        return !state.has(HEALTH) || HEALTH_STATUS_HEALTHY.equals(state.getJSONObject(HEALTH).getString(STATUS));
+    }
+
+    @Override
+    public String healthcheckTests() {
```
rhuss: please use standard getter naming convention (`getHealthcheck()`)

src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
```diff
@@ -121,6 +126,21 @@ public boolean isRunning() {
...
+        if (!json.getJSONObject(CONFIG).has(HEALTHCHECK)) {
+            return null;
+        }
+        return json.getJSONObject(CONFIG).getJSONObject(HEALTHCHECK).getJSONArray(TEST).join(", ");
```
rhuss: is CONFIG and TEST always set or might it make sense to add null checks here, too ?

Actually while looking at the code I wonder about the value to have constants and not using it literally (because they are probably only once anyway). I know, there are already constants, but that's maybe a dead end ? wdyt ?

src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -483,6 +488,45 @@ private String getTcpHost(WaitConfiguration.TcpConfiguration tcpConfig, Properti
         return host;
     }
 
+    private WaitUtil.WaitChecker getHealthyWaitChecker(final String imageConfigDesc, final DockerAccess docker, final String containerId, final List<String> logOut) {
```
rhuss: More a note to myself: We should extract functionality out of `StartMojo` as it already grows quite a bit. The wait stuff could go into an extra package (without normal instead of inner classes).

Also easier to test then.

src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -483,6 +488,45 @@ private String getTcpHost(WaitConfiguration.TcpConfiguration tcpConfig, Properti
...
+            @Override
+            public boolean check() {
+                try {
+                    final InspectedContainer container = docker.getContainer(containerId);
+                    if (container == null) {
+                        log.debug("HealthyWaitChecker:  Container %s not found");
+                        return false;
+                    }
+
+                    if (first) {
```
rhuss: Maybe its better to log out the health check itself in the constructor of the checker, so we don't need this boolean flag dance.

src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -483,6 +488,45 @@ private String getTcpHost(WaitConfiguration.TcpConfiguration tcpConfig, Properti
...
+                        final String healthcheckTests = container.healthcheckTests();
+                        if (healthcheckTests == null) {
+                            throw new IllegalArgumentException("Can not wait for healthy state of "+imageConfigDesc+". No HEALTHCHECK configured.");
+                        }
+                        log.info("%s: Waiting on healthcheck: '%s'", imageConfigDesc, healthcheckTests);
```
rhuss: We could use `log.verbose()` for the full health check test adn `log.info()` for a summary line (and keep it short)

src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
rhuss: Is there a way to detect whether the Docke daemon is too old (<1.12) so that we could prepare a proper error message ?

samples/healthcheck/pom.xml
danielwegener: To be honest, I have no clue what particularly maven does to merge those plugin configs into mojo parameters :) I am open to any other suggestion (maybe as attribute on wait?)

rhuss: It's more a complaint against Maven ;-). No, attributes are not possible at all for Maven configurations, no empty tags, too. 

So its fine for me as it is now, except when we would find something 'useful' to put as value (like a timeout, but thats also available otherwise). But no worries here ...

src/main/java/io/fabric8/maven/docker/model/InspectedContainer.java
danielwegener: `ContainersListElement` can not provide these properties since they are only available on the docker inspect endpoint (single container) and not the list container endpoint. To be honest, right now i can not even find the healthy field documented  it in the official docker api docs, I took the path directly from the docker source. I'll try to provide the original reference asap.

src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
danielwegener: Good catch, docker api might leave the test field undefined if the array is empty - however an empty test does not really make sense. I'll add a precondition.

We could ask `DockerAccess.getServerApiVersion` before each `getHealthyWaitChecker` and ensure a minimal api version (happens on other places too). I wasn't sure if it would be worth the overhead since we are backward-compatible anyway (by not waiting if there is no health status).

src/main/java/io/fabric8/maven/docker/StartMojo.java
danielwegener: Can we be absolutely sure that the container already exists within the docker api when `waitIfRequested` is called? I am not sure if the docker `create container`-api gives us such guarantees (iirc i moved the healthcheck resolution into the check during development because the container could not be fetched once)

src/main/java/io/fabric8/maven/docker/StartMojo.java
danielwegener: Yeah I wasn't sure about this one. Other checkers report their "content" (as url waits log their url). I think some context would be useful but I agree that the plain printing of the docker HEALTHCHECK line is not that beautiful.

danielwegener: what do you think about the logOut then?

src/main/java/io/fabric8/maven/docker/model/InspectedContainer.java
rhuss: Ok, understand.

src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
rhuss: I think its worth the extra round trip (maybe it could be even cached somewhere else ?), because although the plugin won't break if there are no health-check support I think its important to inform the user that the health-check doesn't work because he is running against an too-old Docker daemon. 

So the user configuration can fail for two case: Either no health-check has been provided **or** the Docker daemon doesn't support health-check. We should report that differently. In the first case its a configuration error (doesn't make sense to wait in this case) in the later it's because a potentially supported configuration is run in an non-compatible environment.

src/main/java/io/fabric8/maven/docker/StartMojo.java
rhuss: Yes, I hope we can be sure since e.g. the [TcpWaitChecker](https://github.com/rhuss/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/StartMojo.java#L382-L383) also relies on it when it is created. Haven't got any bug report for this yet ;-) so it should work.

src/main/java/io/fabric8/maven/docker/StartMojo.java
rhuss: I would only use a short, fixed message for `logOut` and post the full health-check with `log.verbose` (once). That is in line with how Dockerfile instruction are only printed with verbose level when an image is build.

src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
danielwegener: >Actually while looking at the code I wonder about the value to have constants and not using it literally (because they are probably only once anyway). I know, there are already constants, but that's maybe a dead end ? wdyt ?

There is an ancient tale about magic numbers and strings and code conventions that argue against them :)

danielwegener: > So the user configuration can fail for two case: Either no health-check has been provided or the Docker daemon doesn't support health-check. We should report that differently. In the first case its a configuration error (doesn't make sense to wait in this case) in the later it's because a potentially supported configuration is run in an non-compatible environment.

Okay. Since we require a certain server api version for this feature anyway, we could also simply check if the `DockerAccessContext.getMinimalApiVersion()` is configured to >=1.25.0 and advice the user to configure it explicitly otherwise (+ add this requirement to the docs).  What do you think?

rhuss: I just saw we could add this check simply to a `WatchConfiguration.initAndValidate()` much like we do for `BuildImageConfiguration.initAndValidate()`. This method is supposed to return [a minimal API version](https://github.com/rhuss/docker-maven-plugin/blob/d431b076a59d97217e2df58ff9d2771a0d83d82e/src/main/java/io/fabric8/maven/docker/config/ConfigHelper.java#L74-L74) to operate. This would result to refuse to work when a [too old Docker daemon is used](https://github.com/rhuss/docker-maven-plugin/blob/7f5d0edd2c29edebb28493313df320a9a3807968/src/main/java/io/fabric8/maven/docker/service/DockerAccessFactory.java#L46-L46).

The implementation should be simple and straight forward.

BTW, I think HEALTHCHECK is supported since 1.24.

rhuss: > There is an ancient tale about magic numbers and strings and code conventions that argue against them :) TBH. I'd be fine with explicit Strings. One level less mental indirection.

;-) maybe we just leave it and then change it when we extract the health check off from StartMojo. Not a biggie anyway.

danielwegener:
Whoops, I guess i'll have too do a rebase asap :)

rhuss:
@danielwegener np ;-) just merged in (actually I refactored the healtchecks out into dedicated classes)

rhuss:
[merge]

danielwegener:
Lovely! :) Thanks

rhuss:
I thank *you* :). d-m-p would be nothing without contributions like this ;) 

thanks ...

PR #725, opened 2017-02-28
Improve removal of temporary tags.

thoschaefer:
Make sure to notify the user if removal of temporary image failed. Also set the force parameter when trying to remove temporary image.

We observed at our Jenkins instance that build job 'n' succeeded but failed to remove the docker tag. Build job 'n+1' then failed with

"Cannot temporarily tag %s with %s because target image already exists. " +

from method tagTemporaryImage. I also wonder if tagTemporaryImage should not try to simply overwrite existing target tags as docker would do it this way itself. That could be another PR I guess.


rhuss:
thanks for the PR and sorry for the delay. quite busy weeks ;-(

The reason why tags are not overwritten forcefully is, that there could be an equally named image which has been created outside of and not related to d-m-p, too. Unlikely, I know, but overwriting this tag is probably not very polite.

I agreed that the information that a temporary tag couldn't be removed should bubble up. But maybe we could do it without chaining the DockerAccess exception but merging it into one: 

sth like 


```java
finally {
    if (!removeTag(...)) {
         if (dae == null) {
             throw new DockerAccessException("Image %s could be pushed, but the temporary tag could not be removed", ...);
         }  else {
            throw new DockerAccessException(dae.getCause(), dae.getMessage() + " and the temporary tag could not be removed, too");
        }
   }

}

```

Still not sure whether the removing of the temporary tag should be considered an fatal error breaking call as the image already could be pushed. Would be awesome if there would be a rollback ;-), but there's nothing like removing an image from a registry ....

rhuss:
@thoschaefer what's you opinion on an adapted error handling as mentioned above ?

thoschaefer:
I think your proposal regarding adapting error handling sounds valid. I will update my PR accordingly.

Regarding you thought about "whether removing of the temporary tag should be considered a fatal error": I am not sure anymore if it should be a fatal error. On the other hand it's confusing if a later build fails because it stumbles upon the non-removed temporary image. Don't understand yet why the image could not be removed. Maybe also worth a retry.

rhuss:
looks good to me. Let's fail the build if the temporary image could not be removed.

rhuss:
[merge]

PR #729, opened 2017-03-02
Fix work with maven build directory path and absolute paths in it

dklabukov:
@rhuss When writing tests to check source and output directories for some reason it was desided to set only `.` as a basefile for MavenProject object. Unfortunately that field was designed to be a link to pom-file not a basepath of a project and because of that some side-effects appeared. Then main one is that when you call `MavenProject.getBasedir` after it, it calls `getParentFile` for a file you set, and it is `null` for `new File('.')` so in this case `new File(params.getProject().getBasedir(), directory)` is equal to `new File(directory)`. Because of this test that checks using an absolute paths as output directories is false positive.

### DIFF
--- a/src/main/asciidoc/inc/_global-configuration.adoc
+++ b/src/main/asciidoc/inc/_global-configuration.adoc
@@ -88,7 +88,7 @@ the URL is:
 | `docker.maxConnections`
 
 | *outputDirectory*
-| If this a relative directory path it is set realtive to ${build.directory} (which normally is the target/ directory within the project directory). The default value is docker and is only used for the goal {plugin}:build.
+| Default output directory to be used by this plugin. If this a relative directory path it is set relative to the project directory. The default value is `target/docker` and is only used for the goal {plugin}:build.
 | `docker.target.dir`
 
 | *portPropertyFile*
--- a/src/main/asciidoc/inc/_global-configuration.adoc
+++ b/src/main/asciidoc/inc/_global-configuration.adoc
@@ -88,7 +88,7 @@ the URL is:
 | `docker.maxConnections`
 
 | *outputDirectory*
-| Default output directory to be used by this plugin. The default value is `{build.directory}/docker` and is only used for the goal `{plugin}:build`.
+| If this a relative directory path it is set realtive to ${build.directory} (which normally is the target/ directory within the project directory). The default value is docker and is only used for the goal {plugin}:build.
 | `docker.target.dir`
 
 | *portPropertyFile*
--- a/src/main/asciidoc/inc/_global-configuration.adoc
+++ b/src/main/asciidoc/inc/_global-configuration.adoc
@@ -88,7 +88,7 @@ the URL is:
 | `docker.maxConnections`
 
 | *outputDirectory*
-| Default output directory to be used by this plugin. The default value is `target/docker` and is only used for the goal `{plugin}:build`.
+| Default output directory to be used by this plugin. The default value is `{build.directory}/docker` and is only used for the goal `{plugin}:build`.
 | `docker.target.dir`
 
 | *portPropertyFile*
--- a/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
@@ -320,6 +320,10 @@ public class EnvUtil {
         if (file.isAbsolute()) {
             return file;
         }
+        File directoryFile = new File(directory);
+        if (directoryFile.isAbsolute()) {
+            return new File(new File(directory), path);
+        }
         return new File(new File(params.getProject().getBasedir(), directory), path);
     }
 
--- a/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
@@ -308,7 +308,7 @@ public class EnvUtil {
     }
 
     public static File prepareAbsoluteOutputDirPath(MojoParameters params, String dir, String path) {
-        return prepareAbsolutePath(params, new File(params.getOutputDirectory(), dir).toString(), path);
+        return prepareAbsolutePath(params, new File(new File(params.getProject().getBuild().getDirectory(), params.getOutputDirectory()), dir).toString(), path);
     }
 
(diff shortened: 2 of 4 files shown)


rhuss:
Thanks ! and sorry for the delay, busy days. I hope I can have a look over the weekend and will review your PR.

In the meantime, could you please do me a favour and sign-off your commits as described in https://github.com/fabric8io/docker-maven-plugin/blob/master/CONTRIBUTING.md ? (you could squash them, too)

thanks ...

rhuss:
Sorry, I probably did not understand actually your use case (and so not really your fix). 

Could you give a concrete example (e.g. configuration) where it causes you problem ?

dklabukov:
@rhuss Our main case is that in some profiles we want to have our ${build.directory} not in `./target` but in `/tmp/maven/${project.groupId}.${project.artifactId}/target`. So at first we desided to set `outputDirectory` in plugin settings to `${build.directory}/docker` but it won't work as expected. It supposed to concatinate module root path and plugin output directory because of missing check that `outputDirectory` is already absolute

src/main/asciidoc/inc/_global-configuration.adoc
```diff
@@ -88,7 +88,7 @@ the URL is:
 | `docker.maxConnections`
 
 | *outputDirectory*
-| Default output directory to be used by this plugin. The default value is `target/docker` and is only used for the goal `{plugin}:build`.
+| Default output directory to be used by this plugin. The default value is `{build.directory}/docker` and is only used for the goal `{plugin}:build`.
```
dklabukov: Done

src/test/java/io/fabric8/maven/docker/assembly/DockerAssemblyConfigurationSourceTest.java
```diff
@@ -51,20 +52,40 @@ public void testCreateSourceRelative() {
...
+    }
+
+    @Test
+    public void testCreateSourceAbsoluteFromLongRelativeProjectBaseDir() {
+        testCreateSource(buildParameters("./projects/project", "/src/docker".replace("/", File.separator), "/output/docker".replace("/", File.separator)));
+    }
+
+    @Test
+    public void testCreateSourceRelativeFromLongRelativeProjectBaseDir() {
+        testCreateSource(buildParameters("./projects/project", "src/docker".replace("/", File.separator), "output/docker".replace("/", File.separator)));
```
dklabukov: Done

src/test/java/io/fabric8/maven/docker/assembly/DockerAssemblyConfigurationSourceTest.java
```diff
@@ -51,20 +52,40 @@ public void testCreateSourceRelative() {
...
         assertTrue(containsDir(image, source.getOutputDirectory()));
         assertTrue(containsDir(image, source.getWorkingDirectory()));
         assertTrue(containsDir(image, source.getTemporaryRootDirectory()));
     }
-    
+
     private MojoParameters buildParameters(String projectDir, String sourceDir, String outputDir) {
         MavenProject mavenProject = new MavenProject();
-        mavenProject.setFile(new File(projectDir));
+        mavenProject.setFile(new File(projectDir + File.separator  + "pom.xml"));
```
dklabukov: Done

src/test/java/io/fabric8/maven/docker/assembly/DockerAssemblyConfigurationSourceTest.java
```diff
@@ -94,18 +115,48 @@ private void testCreateSource(MojoParameters params) {
...
         int length = expectedStartsWith.length();
         assertEquals(expectedStartsWith, path.toString().substring(0, length));
     }
+
+    private void assertChildOfDir(String outputDir, File path) {
+        File absoluteOutput = new File(outputDir).getAbsoluteFile();
+        File absolutePath = path.getAbsoluteFile();
+
+        do {
+            File parentPath = absolutePath.getParentFile();
```
dklabukov: Rewrote that assert, suppose that now it is more clear

rhuss:
Still not 100% sure about your use case and _why_ you would do this, but if I understand you right, you want that if the `outputDirectory` is set to an absolute path, then it should be taken directly. Right ?

I think we should add this. However, we must not change the default value of `outputDirectory` from `target/docker` to `docker`. It can be discussable whether `target/docker` was a good choice from the very beginning, but changing it now would break backwards compatibility without a good reason.

src/main/asciidoc/inc/_global-configuration.adoc
```diff
@@ -88,7 +88,7 @@ the URL is:
 | `docker.maxConnections`
 
 | *outputDirectory*
-| Default output directory to be used by this plugin. The default value is `target/docker` and is only used for the goal `{plugin}:build`.
+| If this a relative directory path it is set realtive to ${build.directory} (which normally is the target/ directory within the project directory). The default value is docker and is only used for the goal {plugin}:build.
```
rhuss: Please don't change the default value

src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -308,7 +308,7 @@ public static String findRegistry(String ... checkFirst) {
     }
 
     public static File prepareAbsoluteOutputDirPath(MojoParameters params, String dir, String path) {
-        return prepareAbsolutePath(params, new File(params.getOutputDirectory(), dir).toString(), path);
+        return prepareAbsolutePath(params, new File(new File(params.getProject().getBuild().getDirectory(), params.getOutputDirectory()), dir).toString(), path);
```
rhuss: Don't really get this line: If `params.getOutputDirectory()` is already an absolute path (it think, this is what this PR is about to support) why would I want it to append it to another directory?

rhuss: tbh, I just would leave the call to `prepareAbsolutePath `here as it was and recommend to use `${build.directory}/docker` in the pom.xml as configuration for the outputDir.

rhuss:
I add some minor changes to simplify things a bit. Please check whether this fits still your use case. If so, I plan to merge this ASAP.

PR #730, opened 2017-03-09
Add PKCS8 private key support

shiroari:
Hi,

This pull request adds support for PKCS8 in key.pem file (default one is PKCS1). Also I changed a bit way of working with readers and stream by using try-with-resources statement and added a few tests.

Note that I remove wrapping FileReader to BufferedReader. I think that was unnecessary because PEMParser is inherited from BufferedReader and prefetches data itself.


rhuss:
Thanks a lot ! I'm bit loaded right now, but will integrate the PR (which makes much sense) over the weekend hopefully.

Could you do me please a favour in the meantime and sign-off the commit as described in https://github.com/fabric8io/docker-maven-plugin/blob/master/CONTRIBUTING.md ?

thanks again ....

shiroari:
Hopefully I did it right. Thank you.

rhuss:
looks good ;-) thanks !

rhuss:
build trigger test - please ignore ....

retest this please

src/main/java/io/fabric8/maven/docker/access/KeyStoreUtil.java
```diff
@@ -41,28 +43,40 @@ public static KeyStore createDockerKeyStore(String certPath) throws IOException,
...
+            PEMParser parser = new PEMParser(reader)) {
+            Object readObject = parser.readObject();
+            if (readObject instanceof PEMKeyPair) {
+                PEMKeyPair keyPair = (PEMKeyPair) readObject;
+                return generatePrivateKey(keyPair.getPrivateKeyInfo());
+            } else if (readObject instanceof PrivateKeyInfo) {
+                return generatePrivateKey((PrivateKeyInfo) readObject);
+            }
+        }
+        return null;
```
rhuss: I would throw an exception in that case, since a private key is mandatory for the further processing. It better to stop where the issue happens and where we have most of the error context (i.e. read object type which is wrong).

src/main/java/io/fabric8/maven/docker/access/KeyStoreUtil.java
```diff
@@ -41,28 +43,40 @@ public static KeyStore createDockerKeyStore(String certPath) throws IOException,
...
-        PEMParser parser = new PEMParser(new BufferedReader(new FileReader(keyPath)));
-        return (PEMKeyPair) parser.readObject();
+    private static PrivateKey generatePrivateKey(PrivateKeyInfo keyInfo) throws IOException, NoSuchAlgorithmException,
+            InvalidKeySpecException {
+        PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(keyInfo.getEncoded());
+        return KeyFactory.getInstance("RSA").generatePrivate(keySpec);
     }
 
-    private static void addCA(KeyStore keyStore, String caPath) throws KeyStoreException, FileNotFoundException, CertificateException {
+    private static void addCA(KeyStore keyStore, String caPath) throws IOException, KeyStoreException,
```
rhuss: Why is FileNotFoundException not good enough here ?

src/main/java/io/fabric8/maven/docker/access/KeyStoreUtil.java
```diff
@@ -41,28 +43,40 @@ public static KeyStore createDockerKeyStore(String certPath) throws IOException,
...
             String alias = crt.getSubjectX500Principal().getName();
             keyStore.setCertificateEntry(alias, crt);
         }
     }
 
-    private static Certificate[] loadCertificates(String certPath) throws FileNotFoundException, CertificateException {
-        InputStream is = new FileInputStream(certPath);
-        Collection<? extends Certificate> certs = CertificateFactory.getInstance("X509").generateCertificates(is);
-        return new ArrayList<>(certs).toArray(new Certificate[certs.size()]);
+    private static Certificate[] loadCertificates(String certPath) throws IOException, CertificateException {
```
rhuss: Same here: Why not keep FileNotFoundException as the more specific exception ?

src/main/java/io/fabric8/maven/docker/access/KeyStoreUtil.java
```diff
@@ -41,28 +43,40 @@ public static KeyStore createDockerKeyStore(String certPath) throws IOException,
         return keyStore;
     }
 
-    public static PrivateKey loadPrivateKey(String keyPath) throws IOException, GeneralSecurityException {
-        PEMKeyPair keyPair = loadPEM(keyPath);
-        PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(keyPair.getPrivateKeyInfo().getEncoded());
-        return KeyFactory.getInstance("RSA").generatePrivate(keySpec);
+    static PrivateKey loadPrivateKey(String keyPath) throws IOException, GeneralSecurityException {
+        try (Reader reader = new FileReader(keyPath);
```
rhuss: thanks for adding resource management here

src/main/java/io/fabric8/maven/docker/access/KeyStoreUtil.java
shiroari: Thanks for the comments. That would be definitely more correct to throw exception here. Will fix.

src/main/java/io/fabric8/maven/docker/access/KeyStoreUtil.java
shiroari: The problem here that try-with-resources statement automatically closes input stream and close method throws `IOException` and then it goes to `addCA`.

rhuss:
Thanks ! Looks good to me

[merge]

PR #735, opened 2017-03-15
Do not attempt to build null-named images, skip them.

ucosty:
Makes it easier to include JAR based submodules in your projects.

Part of the problem is the ```docker:build``` goal gets indiscriminately run against all modules in a project, regardless of whether or not they are configured to build a docker image.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/ConfigHelper.java
+++ b/src/main/java/io/fabric8/maven/docker/config/ConfigHelper.java
@@ -109,23 +109,14 @@ public class ConfigHelper {
         List<ImageConfiguration> ret = new ArrayList<>();
         if (unresolvedImages != null) {
             for (ImageConfiguration image : unresolvedImages) {
-                ret.addAll(imageResolver.resolve(image));
+                if(image.getName() != null) {
+                    ret.addAll(imageResolver.resolve(image));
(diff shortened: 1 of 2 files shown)


rhuss:
Thanks for the PR !

A question: How would you report an error back when a user forgets to add an  image name which is mandatory ? Wouldn't it be better to check for a `<build>` configuration whether, which, if given, indicates that the use want to run a build ?

Having a non-null `<image>` config without a name still looks like an error condition for me. would be awesome if you could provide a multi-module example where the current behaviour fails for you.

rhuss:
@ucosty really need to understand to use case better, so an example would be super awesome.

rhuss:
@ucosty still connected to this PR ? (sorry for the delay). If you could elaborate a bit on your use case that would help me quite a bit to understand it and how to implement it.

ucosty:
@rhuss sorry about that. We've stopped using the docker maven plugin in the project that was affected, so I'm not that fussed now. We had issues with our project, having quite a lot of nested maven projects (some pom docker targets, some jar docker targets, other placeholders and parents)

rhuss:
@ucosty thanks for the feedback. hope you don't mind that I close this PR, please reopen if you need it again.

PR #742, opened 2017-03-24
Option to turn off SSL communication on 2376

Fluxxo:
Added a switch to disable forced SSL communication. This enables the ……user to communicate with docker on port 2376 event when the daemon is not SSL secured. This makes the plugin more aligned with the docker cli that accepts non-ssl connection even on the standard SSL port (2376)


rhuss:
Thanks for the PR and apologies for the late response.

I would like to understand your use case a bit better. In which situation is the Docker daemon running on port 2376 with plain http ? My understanding is that 2375 stands for http and 2376 for https. Even when this is only a convention, what is the use case to serve http on 2376 ?

When you say that the Docker CLI 'accepts non-ssl connection on 2376' do you have a reference, maybe in the Docker docs, where this use case is described?

If this is a valid use case (and if I understand it ;-), I would prefer to mimic the same detection like the Docker CLI, possibly by probing the port. Adding a new configuration option should be done only as a last resort.

Fluxxo:
Hi, 

sorry for the very late feedback from my side this time.

The point is the following:

In your documentation is stated, that if you supply a tcp://HOSTNAME:PORT assignment, it will determine whether to use SSL or not based on the port (if 2375, don't use SSL, if not, use SSL).

The flag forces the docker client in the plugin to not use SSL even though the port is determined as an SSL-port.
The point is, that some docker installations run on different ports but without SSL.

rhuss:
Sorry, something messed up this PR considerably. I will cherry pick you change and force push to make it clean again.

rhuss:
> In your documentation is stated, that if you supply a tcp://HOSTNAME:PORT assignment, it will determine whether to use SSL or not based on the port (if 2375, don't use SSL, if not, use SSL).
> 
> The flag forces the docker client in the plugin to not use SSL even though the port is determined as an SSL-port.
> The point is, that some docker installations run on different ports but without SSL.

Actually, the documentation then is wrong: https is _only_ used when the port on the `tcp://` URL is 2376, otherwise it is always http. What i found in the documentation was:

> The URL of the Docker Daemon. If this configuration option is not given, then the optional `<machine>` configuration section is consulted. The scheme of the URL can be either given directly as `http` or `https`
depending on whether plain HTTP communication is enabled or SSL should
be used. Alternatively the scheme could be `tcp` in which case the
protocol is determined via the IANA assigned port: 2375 for `http`
and 2376 for `https`. 

So doesn't really say something about other ports beside 2375 and 2376 (but I should add that it will be HTTP otherwise)

So the `skipSsl` property would be only in effect for port 2376, but this commonly accepted to be the HTTPs port. So I'd like still to ask for your use case why to run plain HTTP on port 2376 ?

Sorry for the delay btw, it really took me a bit to come back to the plugin (and I was a bit overwhelmed by the state of the PR, which I have cleaned up hopefully ;-)

src/test/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClientSSLTest.java
```diff
@@ -0,0 +1,37 @@
...
+            @Override
+            ApacheHttpClientDelegate createHttpClient(ClientBuilder builder) throws IOException {
+                return mockDelegate;
+            }
+        };
+    }
+
+    @Test
+    public void testPushFailes_noRetry() throws Exception {
+        assertEquals(this.client.getUrlBuilder().getBaseUrl(), "http://1.2.3.4:2376");
```
pepperbot: MAJOR Swap these 2 arguments so they are in the correct order: expected value, actual value. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS3415)

rhuss:
No feedback, so closing it. Feel free to reopen if this is still a thing.

PR #747, opened 2017-03-29
Be more robust when parsing Docker auth config

### LINKED ISSUE #731: fail to build and deploy
The docker maven plugin is unable to build and deploy the maven project when the ~/.docker/config.json has `auths` entries without any auth credentials.

Ref: https://github.com/fabric8io/fabric8-maven-plugin/issues/868

rhuss:
Especially when there is no auth set, don't barf but simply don't
use any auth.

Fixes #731


rhuss:
[merge]

PR #757, opened 2017-04-10
Make wait fail if wait conditions not reached after the container stops

janssk1:
Hi, 

I noticed that wait checkers (eg health/log/..) keep waiting for conditions to be reached, even after the container has stopped. Since nothing will be changing anymore after the container has ended, no need to continue waiting for a timeout. 

This pull request checks during wait if the container is still 'running'. If not, waiting is stopped.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/wait/WaitUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/wait/WaitUtil.java
@@ -1,5 +1,11 @@
 package io.fabric8.maven.docker.wait;
 
+import io.fabric8.maven.docker.access.DockerAccess;
+import io.fabric8.maven.docker.access.DockerAccessException;
+import io.fabric8.maven.docker.log.LogDispatcher;
+import io.fabric8.maven.docker.log.LogOutputSpec;
+
+import java.io.FileNotFoundException;
 import java.util.Arrays;
 import java.util.concurrent.*;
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/wait/WaitUtil.java
```diff
@@ -1,5 +1,11 @@
 package io.fabric8.maven.docker.wait;
 
+import io.fabric8.maven.docker.access.DockerAccess;
```
rhuss: Actually I'd prefer to add a `IsRunningChecker` which is put first in the checker chain. This checker can have access to the `dockerAccess` (ike the `LogWaitChecker` has). Then, if the container is not running I would throw a `WaitResultException`, generalising on the `WaitTimeoutException`

```java
publi class WaitResultException extend Exception {

    private WaitResultException.Reason reason;

   enum Reason { timeout, noRunning }
   ....
}
```

src/main/java/io/fabric8/maven/docker/wait/WaitUtil.java
```diff
@@ -36,17 +42,34 @@ public static long wait(int wait, Callable<Void> callable) throws ExecutionExcep
...
+    public static long wait(DockerAccess access, String containerId, int maxWait, Iterable<WaitChecker> checkers) throws WaitTimeoutException, NotRunningException {
         long max = maxWait > 0 ? maxWait : DEFAULT_MAX_WAIT;
         long now = System.currentTimeMillis();
         try {
             do {
-                if (check(checkers)) {
-                    return delta(now);
+                try {
+                    if (!access.getContainer(containerId).isRunning()) {
+                        if (check(checkers)) {
```
janssk1: This 'double checking' is actually necessary. If the container is not running anymore ,the checkers have to be run again since a checker might have changed his state. EG: If the process prints out an expected string just before it ends. In that case, the wait should just end instead of failing with a 'unexpected end' exception. 

For the same reason, i don't think it will be feasible to implement the running check as a waiter.

rhuss: ok, I see the point. Let me pick this up and try whether I could keep the separation (you are probably right, it's not possible to use this is a check).

rhuss:
[merge]

rhuss:
Fixed build issue and refactored a bit in #773 

Closing this PR in favor of #773 which is build on top of this PR.

Thanks a lot !

Issue #498, opened 2016-06-28
<wait> on container to finish


rhuss:
A new `<wait>` section should be added which waits until a container is finished and exit the build according to the exit status of the container. This is useful for batch like jobs where one simply want to run a container e.g. for initializations or other task.

This could look like

``` xml
<wait>
   <exit>0</exit>
</wait>
```

for an expected exit code of 0 (which could be the default, but since Maven doesnt allow empty tags one probably needs to provide it in any case)

robythomas:
I have the exact requirement to run a container that would eventually exit with a proper code signifying a job completed status. Currently, I <wait> on a <log> pattern in that job container. But I've seen that somtimes this pattern matching doesn't catch it. It has happened in about two or three times already for me. While that perhaps is an issue with the <log> mechanism, having this wait for exit is exactly what I'm looking for.

rhuss:
Implemented with #765

PR #765, opened 2017-04-27
Added exit code checker (see #498)

slauth:
Signed-off-by: Sebastian Lauth <github@slauth.de>


rhuss:
Thanks ! I will have a look asap ...

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
```diff
@@ -0,0 +1,35 @@
...
+    public ExitCodeChecker(int exitCode, ServiceHub hub, String containerId) {
+        this.exitCode = exitCode;
+        this.hub = hub;
+        this.containerId = containerId;
+    }
+
+    @Override
+    public boolean check() {
+        try {
+            Integer exitCode = hub.getQueryService().getMandatoryContainer(containerId).getExitCode();
```
rhuss: Please don't use a local variable with the same name as a field. Might be ok for a constructor but please not in a method.

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
```diff
@@ -0,0 +1,35 @@
...
+            if (exitCode == null) {
+                return false;
+            }
+            return exitCode == this.exitCode;
```
rhuss: When a container exited and the exit code does not match the expected exit code, this will never change. So I wonder whether it would not be better to throw an exception in this case? So that the run aborts immediately as it never can meet the wait condition.

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
```diff
@@ -0,0 +1,35 @@
...
+import io.fabric8.maven.docker.access.DockerAccessException;
+import io.fabric8.maven.docker.service.ServiceHub;
+
+public class ExitCodeChecker implements WaitChecker {
+
+    private final int exitCode;
+    private final ServiceHub hub;
+    private final String containerId;
+
+    public ExitCodeChecker(int exitCode, ServiceHub hub, String containerId) {
```
rhuss: Could you please add a unit test for the checker ? You can use jmockit for mocking the service hub, please have a look at the existing tests.

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
slauth: You're right. But this situation would be also handled by #757, wouldn't it?

rhuss: Yes, think so

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
slauth: Makes sense, added it.

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
slauth: Renamed to expected / actual exit code.

src/main/java/io/fabric8/maven/docker/wait/ExitCodeChecker.java
slauth: So I left this untouched in favor of #757.

rhuss:
[merge]

PR #774, opened 2017-05-11
Document default lifecycle bindings

hho:
This PR adds to the documentation the default bindings of the plugin goals to the Maven lifecycle phases.

I hope this helps folks to better grasp _when_ the plugin does _what_ and reduce the "I have no idea why this happens" moments for newbies (like me).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,19 +12,19 @@ It works with Maven 3.0.5 and Docker 1.6.0 or later.
 
 #### Goals
 
-| Goal                                          | Description                           |
-| --------------------------------------------- | ------------------------------------- |
-| [`docker:start`](https://fabric8io.github.io/docker-maven-plugin/#docker:start)   | Create and start containers           |
-| [`docker:stop`](https://fabric8io.github.io/docker-maven-plugin/#docker:stop)     | Stop and destroy containers           |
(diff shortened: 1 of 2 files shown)


rhuss:
Thanks a lot! Makes much sense. BTW, if you have a suggestion for different default lifecycle bindings, please let us know. Always struggling to find a good compromise ...

rhuss:
[merge]

hho:
Thanks for the super-quick merge!

> BTW, if you have a suggestion for different default lifecycle bindings, please let us know. Always struggling to find a good compromise ...

The only thing I'd probably change would be to do `docker:build` during the `package` phase, so the newly built image then can be used to start a container for the integration tests.

However, it might be difficult to ensure that `docker:build` only runs after all other packaging goals (e.g. `jar:jar`, `spring-boot-maven-plugin:repack` etc.) when all are bound to the same phase.

rhuss:
Yeah, would have loved to move `build` to `package`, too, but as you said, this phase is really crowded and `build` actually needs everything packaged nicely. A `post-package` would have been nice, but until now `install` works not so bad.

btw, I changed the column order a bit in the docs, hope this ok.

PR #783, opened 2017-05-19
Allow system/cli properties to be used in the Dockerfile

pgier:
Currently the Dockerfile interpolation doesn't include properties specified from
the command line, and only uses the values in the POM.  This change allows
the cli to set new properties and override the pom property values.


rhuss:
Thanks ! LGTM.

[merge]

PR #788, opened 2017-05-25
Don't fail build if volume-clean fails.

KedneckInc:
This fix is admittedly simplistic.  It does not take into account any of the other possible causes for an exception.  However, if you are using volume-clean, chances are you're using volume-create, or many of the other goals provided by DMP.  In that case, you will almost certainly receive errors pointing you at the
real problem.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -8,6 +8,7 @@
   - Preserve leading whitespace in logs (#875)
   - Maven property interpolation in Dockerfiles (#877)
   - Allow parameters for the log prefix (#890)
+  - When removing a volume don't error if the volume does not exist (#788)
   
 * **0.22.1** (2017-08-28)
   - Allow Docker compose version "2", too ([#829](https://github.com/fabric8io/docker-maven-plugin/issues/829))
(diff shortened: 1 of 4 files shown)


rhuss:
Sorry, for coming back to this PR that late. However,  it seems to be broken to me when you look into the list of changes and touched files. The conflicts are because we moved on, but the original changeset is still too huge to even review it. 

I'm going to close this PR therefore, it would be awesome if you create a new PR. Alternatively, if you describe the problem in a GitHub issue we could try to find a solution together.

Again sorry for the very late response.

rhuss:
I cherry picked your change now, should be clean again.

rhuss:
I just allowed 404 as a valid response code for a volume delete operation. This is returned when the volume does not exists. I think this is the better solution.

KedneckInc:
Wunderbar!  I'll give it a look because I finally got back to that part of the project!

KedneckInc:
Sorry I took so long to get back to this.  Yes, this definitely fixed the problem.  Many Thanks!  Now our developers don't have to manually clean up the old volumes!

PR #790, opened 2017-05-25
PropertyConfigHandler fixes

adrianluisgonzalez:
PropertyConfigHandler#buildConfigured should also return true if the dockerFile or dockerFileDir properties are set

BuildImageConfiguration do not overwrite filter and cleanup default values with null properties

Signed-off-by: Adrian Gonzalez <adrian.gonzalez@ef.com>


rhuss:
Thanks ! lgtm

[merge]

PR #810, opened 2017-07-08
Allow comma separated list of container names in dependsOn elements

chonton:
### Description
Allow comma separated list of <dependsOn> elements.  (Similar to #558)
### Info

* Feature request: Allow something like:
```
<dependsOn>
   <container>mongo,identity</container>
</dependsOn>
```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
@@ -97,7 +97,7 @@ public class ImageConfiguration implements StartOrderResolver.Resolvable, Serial
 
     private void addLinks(RunImageConfiguration runConfig, List<String> ret) {
         // Custom networks can have circular links, no need to be considered for the starting order.
-        if (runConfig.getLinks() != null && !runConfig.getNetworkingConfig().isCustomNetwork()) {
+        if (!runConfig.getNetworkingConfig().isCustomNetwork()) {
             for (String[] link : EnvUtil.splitOnLastColon(runConfig.getLinks())) {
                 ret.add(link[0]);
             }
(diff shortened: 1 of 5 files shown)


src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
```diff
@@ -97,7 +97,7 @@ private void addVolumes(RunImageConfiguration runConfig, List<String> ret) {
 
     private void addLinks(RunImageConfiguration runConfig, List<String> ret) {
         // Custom networks can have circular links, no need to be considered for the starting order.
-        if (runConfig.getLinks() != null && !runConfig.getNetworkingConfig().isCustomNetwork()) {
```
rhuss: Are you sure that getLinks() is never null ?

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
```diff
@@ -114,7 +114,7 @@ private void addContainerNetwork(RunImageConfiguration runConfig, List<String> r
 
     private void addDependsOn(RunImageConfiguration runConfig, List<String> ret) {
         // Only used in custom networks.
-        if (runConfig.getDependsOn() != null && runConfig.getNetworkingConfig().isCustomNetwork()) {
```
rhuss: why can't getDependsOn() be null ?

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
chonton: I've added @Nonnull annotations to show the deduction chain.  EnvUtil.splitAtCommasAndTrim(Iterable<String> input) never returns null.

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
chonton: likewise.

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
rhuss: Ok, I'm settled then ;-)

Issue #812, opened 2017-07-13
useAllReactorProjects in assembly descriptor causes build failure


ph33rtehgd:
_Please fill out the following information to help us in analyzing the issue, but feel free to skip it if you don't have the information at hand or if it does not apply. Please remove everything which does not apply to the issue. Including this info paragraph ;-)_

### Description
When using the <useAllReactorProjects>true</useAllReactorProjects> option of <moduleSet> in the assembly descriptor the maven build will fail with: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0

### Info

* d-m-p version : 0.21.0
* Maven version (`mvn -v`) : 3.3.9 (Spring Tool Suite embedded)
```

```
* Docker version : Docker version 17.05.0-ce, build 89658be
* If it's a bug, how to reproduce 
1. Setup a basic project that will invoke the plugin's build goal.
2. Attach an assembly descriptor within the <build> tag whose contents contain a <moduleSet> which has<useAllReactorProjects>true</useAllReactorProjects>

Full maven stacktrace:
`[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.21.0:build (build-image) on project assembly: Execution build-image of goal io.fabric8:docker-maven-plugin:0.21.0:build failed: Index: 0, Size: 0 -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:docker-maven-plugin:0.21.0:build (build-image) on project assembly: Execution build-image of goal io.fabric8:docker-maven-plugin:0.21.0:build failed: Index: 0, Size: 0
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
[...]

PR #814, opened 2017-07-14
Fix for Issue #812 - useAllReactorProjects causes build failure

ph33rtehgd:
I tried to squash the commits but I'm a Git neophyte and it doesn't seem to have worked (actually made it worse :( )?

This fix takes the reactorProjects property from Maven and places it into the MojoParameters so that the DockerAssemblyConfigurationSource has access to the proper reactor projects list (and subsequentially the maven assembly plugin will ultimately have access to it).

The specific use case where this issue was discovered was in a multi-module maven project and the project which was assembling the docker image (using the build goal) was the last project in the multi-module project list. Adding useAllReactorProjects would cause the build to fail as the existing method for getting reactor projects was returning a list of 0 projects (instead of all the projects in the reactor as expected).


rhuss:
No problem with the squash, going to squash it for you ;-) (btw, if you are on OS X I can not recommend "gitup" enough, awesome tool for doing exactly that kind of stuff).

PR looks good to me, so let's merge it.

ph33rtehgd:
Much appreciated! I'm on Windows so gitup isn't an option for me :(.  I'll have to poke around for something else :)

rhuss:
No problem, you just missed a rebase on master (for the change happened since you branched).

I did it and opened #815 for this, which I'm going to merge after the test passed.

Thanks !

PR #816, opened 2017-07-14
Ignore empty List<String> configuration values

chonton:
Allow empty list elements.  This often occurs when a parent (corporate) pom has configuration values that are overridden by some, but not all, children.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -7,6 +7,8 @@ import java.util.*;
 import io.fabric8.maven.docker.util.*;
 import org.apache.maven.plugins.annotations.Parameter;
 
+import javax.annotation.Nonnull;
+
 /**
  * @author roland
  * @since 02.09.14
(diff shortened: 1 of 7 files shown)


src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -175,16 +177,19 @@ public AssemblyConfiguration getAssemblyConfiguration() {
         return assembly;
     }
 
+    @Nonnull
     public List<String> getPorts() {
-        return ports;
+        return EnvUtil.removeEmpties(ports);
```
rhuss: minor nitpick: Could we rename it to removeNullEntries() ?

chonton: Sure, however it removes both null and empty values.

rhuss: Ah. ok. I thought it was about null entries. We should keep it of course like this, maybe calling it 'removeEmptyEntries()' ?

Trim is find for me, too.

chonton-elementum:
@rhuss, did you intend to merge this into master or not?  Do I need to create another pull request with new base?

rhuss:
Sorry, I simply forgot to merge ;-)

[merge]

Issue #806, opened 2017-07-01
Support docker credential store


denisa:
Docker has recently introduced the concept of a credential store.
With a credential store, docker saves the credentials in a platform specific safe place, such as macOS keychain.

For example, my ~/.docker/config.json looks like:
```
{
    "auths": {
        "https://index.docker.io/v1/": {},
        "https://us.gcr.io": {},
        "us.gcr.io": {}
    },
    "credsStore": "osxkeychain"
}
```
Clean & safe.

However, this doesn't work with version 0.21.0 of the plugin, as it only supports having credential in clear text in that file.

[Docker documentation for the credential store](https://docs.docker.com/engine/reference/commandline/login/)

rhuss:
Sounds like a reasonable addition the plugin, thanks for the heads up. If you feel fancy and might want to contribute and could give you some pointers for where to start. 

Otherwise it might take a bit unti we pick up this feature, we are all currently quite busy otherwise.

denisa:
fair ;-)

From the [instruction](https://github.com/docker/docker-credential-helpers), the helper must be on $PATH and be called with a `get` argument and the desired server on STDIN.

Example:
```
> echo "us.gcr.io" | docker-credential-osxkeychain get
{"ServerURL":"us.gcr.io","Username":"oauth2accesstoken","Secret":"… my secret key …"}
```

denisa:
It seems like docker now also supports credHelpers as an alternative to credsStore, to specify per-registry helpers, see https://github.com/moby/moby/pull/25851

I will look into adding to existing merge request.

denisa:
This is the default behavior of any recent docker installation of docker for mac. This feature is thus important to support the ‘fall-back to docker credentials’ mode of the plugin.

rhuss:
Support has been added.

PR #821, opened 2017-07-20
use credential helper if defined

denisa:
Signed-off-by: Denis N. Antonioli <d_n_antonioli@yahoo.com>

Fix issue #806.
In need of feedback, especially any idea on how to test this…

I have only performed happy-path tests manually until now.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,7 +2,8 @@
 
 * **0.22-SNAPSHOT**
   - Support relative paths when binding volumes in `docker-compose.yml` (#846)
-  - Allo the session token for AWS authetication to be oncluded in order to allow temporary security credentials provided by the AWS Security Token Service (AWS STS) to sign requests (#883)
+  - Allow  the session token for AWS authetication to be oncluded in order to allow temporary security credentials provided by the AWS Security Token Service (AWS STS) to sign requests (#883)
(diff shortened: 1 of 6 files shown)


rhuss:
Thanks ! I will have a look asap, but will be on PTO this week, so might take a bit. Please ping me, if I forget this after this week.

denisa:
any chance of getting this reviewed in a neat future?

rhuss:
sorry, still under full load with other stuff ;-( But I definitely consider this PR for the next release, which is not so far away. Hopefully on the weekend, but if not, it still might take three weeks as I'm fully booked with other work for the next time.

really sorry, but it's not forgotten ...

rhuss:
I rebased and made some minor updates.

Look good to me, lets merge it (sorry for the ultra long delay).

wrt testing: I think we could do this by including some shell scripts into the code which are written out to a tempdir during the test and then mock the `.dockercfg` read to point to this shell scripts which return expected JSON when called as external command. It's a bit involved and only works for certain environments (unix like), but might be worth to try. If you feel fancy, you can give it a try (I'm unfortunately still under high load so have no time for this :(

Thanks a lot ! (and sorry again, you can expect a release on the weekend)

rhuss:
@denisa 0.23.0 is out. Thanks for your contribution !

PR #834, opened 2017-08-03
Fix test failures when build under Windows

akuhtz:
Fix test failures when build under Windows.
The validation of the Windows filename is copied from stackoverflow: https://stackoverflow.com/a/6804755

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -501,7 +501,10 @@ public class BuildImageConfiguration implements Serializable {
             if (dockerFileDir == null) {
                 return dFile;
             } else {
-                if (dFile.isAbsolute() || !EnvUtil.isValidWindowsFileName(dockerFile)) {
+                if (dFile.isAbsolute()) {
+                    throw new IllegalArgumentException("<dockerFile> can not be absolute path if <dockerFileDir> also set.");
+                }
+                if (EnvUtil.isWindows() && !EnvUtil.isValidWindowsFileName(dockerFile)) {
                     throw new IllegalArgumentException("<dockerFile> can not be absolute path if <dockerFileDir> also set.");
                 }
                 return new File(dockerFileDir, dockerFile);
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -501,7 +501,7 @@ public class BuildImageConfiguration implements Serializable {
             if (dockerFileDir == null) {
                 return dFile;
             } else {
-                if (dFile.isAbsolute()) {
+                if (dFile.isAbsolute() || !EnvUtil.isValidWindowsFileName(dockerFile)) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -501,7 +501,7 @@ private File findDockerFileFile(Logger log) {
             if (dockerFileDir == null) {
                 return dFile;
             } else {
-                if (dFile.isAbsolute()) {
+                if (dFile.isAbsolute() || !EnvUtil.isValidWindowsFileName(dockerFile)) {
```
rhuss: Could you please add this as an extra check with an extra error message for the invalide windows file name case ? Otherwise its confusing when the exception message does not fit to the error.

src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -392,4 +392,32 @@ public static Date loadTimestamp(File tsFile) throws MojoExecutionException {
     public static boolean isWindows() {
         return System.getProperty("os.name").toLowerCase().contains("windows");
     }
+    
+    public static boolean isValidWindowsFileName(String text) {
+    	
+    	if (!isWindows()) {
```
rhuss: Please put the windows checks outside this method (e.g. where it is called) as otherwise the semantics of this check is confusing (e.g. returning different return values depending on the same argument, depending on on which OS the method is called)

src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -392,4 +392,32 @@ public static Date loadTimestamp(File tsFile) throws MojoExecutionException {
...
+    		// only care on windows
+    		return true;
+    	}
+    	
+        Pattern pattern = Pattern.compile(
```
rhuss: Could you please add a reference to the SO answer as comment above ?

src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -392,4 +392,32 @@ public static Date loadTimestamp(File tsFile) throws MojoExecutionException {
     public static boolean isWindows() {
         return System.getProperty("os.name").toLowerCase().contains("windows");
     }
+    
+    public static boolean isValidWindowsFileName(String text) {
```
rhuss: Some unit tests for this methods would be awesome (shouldn't be hard to build)

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -504,6 +504,9 @@ private File findDockerFileFile(Logger log) {
                 if (dFile.isAbsolute()) {
                     throw new IllegalArgumentException("<dockerFile> can not be absolute path if <dockerFileDir> also set.");
                 }
+                if (EnvUtil.isWindows() && !EnvUtil.isValidWindowsFileName(dockerFile)) {
+                    throw new IllegalArgumentException("<dockerFile> can not be absolute path if <dockerFileDir> also set.");
```
rhuss: Shouldn't this be something like 

```
throw new IllegalArgumentException(String.format("Invalid windows file name %s for dockerFile", dockerFile))
```

rhuss:
Thanks !

PR #835, opened 2017-08-04
Adding customizer to service to allow changing the archive root

nicolaferraro:
This will expose the archive customizer to be able to write to the root of the generated tar file (assembly config allows to write into a subfolder, e.g. "maven" by default). S2I files should be written to the ".s2i" folder in the root of the archive.

@rhuss I've tried not to expose maven/plexus objects but I can't find a straightforward approach.

This is to fix the Issue: https://github.com/fabric8io/fabric8-maven-plugin/issues/1009
Related change to f-m-p: https://github.com/nicolaferraro/fabric8-maven-plugin/tree/1009-java-options

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/AllFilesExecCustomizer.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/AllFilesExecCustomizer.java
@@ -31,11 +31,11 @@ import org.codehaus.plexus.util.StringUtils;
  * @author roland
  * @since 26/06/16
  */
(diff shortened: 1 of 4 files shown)


nicolaferraro:
I also though about that, but that way you cannot allow users e.g. to customize file and directory modes. The interface should be more complex.

rhuss:
ok. as you mentioned, the coupling to Maven is a bit unfortunate, I hope sometime we could extract this part to fabric8-build, too (to the non-maven section). But for now thats good enough, lets go on ...

rhuss:
[merge]

Issue #846, opened 2017-08-23
Support relative paths in docker-compose 'volumes:' statements


emetsger:
### Description
Relative paths in docker-compose `volumes:` statements are not supported.  For example, the volumes stanza below 
```
version: '2.2'
services:
  solr:
  ...
     volumes:
     - "./cores:/opt/solr/server/solr/mycores"
```
results in the following error on `docker:start`:
```
[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.21-SNAPSHOT:start (default-cli) on project rmap-indexing-solr: I/O Error: Unable to create container for [solr:6.6.0-alpine] : create ./cores: "./cores" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed (Internal Server Error: 500) -> [Help 1]
```

### Info
* d-m-p version : `e50a7a5`, `0.21-SNAPSHOT`
* Maven version (`mvn -v`) :
```
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T10:41:47-06:00)
Maven home: /Users/esm/maven-3
Java version: 1.8.0_131, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.11.6", arch: "x86_64", family: "mac"
```
* Docker version : `Docker version 1.11.2, build b9f10c9`, `docker-machine version 0.7.0, build a650a40`

emetsger:
FYI, after upgrading the docker instance to `v17.06.1-ce`, the error message is a little more clear, with the additional text: `If you intended to pass a host directory, use absolute path`

```
DOCKER> I/O Error [Unable to create container for [solr:6.6.0-alpine] : create ./cores: "./cores" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path (Internal Server Error: 500)]
```

emetsger:
Closed by merge `802fa9b` of pr #848

PR #848, opened 2017-08-23
Support relative paths in `volumes` statements in docker-compose.yaml

emetsger:
* Moves path resolution logic to a utility class
* Includes tests
* Adds mockito-core as a test dependency

Addresses #846

Signed-off-by: Elliot Metsger <emetsger@jhu.edu>


src/test/java/io/fabric8/maven/docker/config/handler/compose/ComposeUtilsTest.java
```diff
@@ -0,0 +1,50 @@
+package io.fabric8.maven.docker.config.handler.compose;
+
+import org.apache.maven.project.MavenProject;
+import org.junit.Test;
+
+import java.io.File;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
```
rhuss: would it be able to use jmockit as Mock framework (as we already use this and I'm not so inclined to use multiple mock frameworks in d-m-p)

You could take e.g. https://github.com/rhuss/docker-maven-plugin/blob/674e80fdf56d31fdd263366d53e7a5cf0dc064b3/src/test/java/io/fabric8/maven/docker/access/log/LogRequestorTest.java#L220 as a sample how expectations are setup in jmockit.

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
```diff
@@ -396,4 +406,97 @@ private Arguments asArguments(Object command, String label) {
...
+     * <p>
+     * Given {@code baseDir} equal to "/path/to/basedir" and a {@code volume} string equal to
+     * "~/reldir:/some/other/dir", this method returns {@code /path/to/user/home/reldir:/some/other/dir}
+     * </p>
+     *
+     * @param baseDir the base directory used to resolve paths beginning with {@code ./}; <em>must</em> be absolute
+     * @param volume the volume string from the docker-compose file
+     * @return the volume string, with any relative paths resolved as absolute paths
+     */
+    static String resolveRelativeVolumeBinding(File baseDir, String volume) {
```
rhuss: It would be awesome if we could use the same kind of parsing for the regular configuration, too. 

E.g. in 

* https://github.com/rhuss/docker-maven-plugin/blob/674e80fdf56d31fdd263366d53e7a5cf0dc064b3/src/main/java/io/fabric8/maven/docker/service/RunService.java#L258
* https://github.com/rhuss/docker-maven-plugin/blob/674e80fdf56d31fdd263366d53e7a5cf0dc064b3/src/main/java/io/fabric8/maven/docker/service/RunService.java#L332

The challenge there is to get the basedir into the method (which must not have a reference to any Maven code). Needs probably given through top-down.

emetsger: It appears that relative paths in volume binding strings can only appear in `docker-compose.yaml` files. 

The Docker [Run reference](https://docs.docker.com/engine/reference/run/#volume-shared-filesystems) indicates that relative paths are _not_ allowed in image `VOLUME`s:

> -v, --volume=[host-src:]container-dest[:<options>]: Bind mount a volume.
The comma-delimited `options` are [rw|ro], [z|Z],
[[r]shared|[r]slave|[r]private], and [nocopy].
The 'host-src' is an absolute path or a name value.

And later in the same section:

> The container-dest must always be an absolute path such as /src/docs. The host-src can either be an absolute path or a name value. If you supply an absolute path for the host-dir, Docker bind-mounts to the path you specify. If you supply a name, Docker creates a named volume by that name.

> A name value must start with an alphanumeric character, followed by a-z0-9, _ (underscore), . (period) or - (hyphen). An absolute path starts with a / (forward slash).

> For example, you can specify either /foo or foo for a host-src value. If you supply the /foo value, Docker creates a bind-mount. If you supply the foo specification, Docker creates a named volume.

Let me know if you don't agree, but it seems that relative paths only need to be accounted for when using a `docker-compose.yml` file.

In that case, the base directory used to resolve relative paths should be the one defined in the compose handler:
```xml
  <external> 
     <type>compose</type> 
     <basedir>src/main/docker</basedir> 
     <composeFile>docker-compose.yml</composeFile>
  </external>
```

Overall, I'm not sure of the wider applicability this code.  What are your thoughts?  Maybe I am missing something.  I can certainly clean up the PR a bit, and have the resolution logic ready to be used by other classes as needed.

emetsger:
Hi @rhuss,

Let me know if you think I've got all the bases covered.  I tried to address each of your comments, but note my response [here](https://github.com/fabric8io/docker-maven-plugin/pull/848#discussion_r135066012) which may be hidden depending on how the GitHub UI renders the comment in your browser.

I think the major portions of the logic are `public` and have decent test coverage.  If other portions of the code base want to support relative volume binding strings, the `VolumeUtil` class has two methods to support this.

Again, based on  my reasoning [above](https://github.com/fabric8io/docker-maven-plugin/pull/848#discussion_r135066012), I'm not sure if there are other portions of the codebase that can take advantage of this, but I just may not be understanding the code base, or your request.  Let me know if there's anything else you need.

emetsger:
@rhuss just a ping on this PR-  many thanks for your help/direction!

rhuss:
@emetsger Thanks, will have a look ASAP. hope for today, but work is killing me ...

emetsger:
@rhuss fixed test failures, added some javadoc, rebased against master.

rhuss:
Thanks  a lot ! 

@nicolaferraro could you have a brief look, I'm really so overloaded ;-( This is an awesome PR, I would like to get it merged asap (before my PTO next week)

nicolaferraro:
Sure @rhuss, I can look at it this morning.

emetsger:
@nicolaferraro @rhuss I think I've hit everything, please let me know if more updates are needed!  Thanks for the review!

nicolaferraro:
@emetsger it would be good if you can test it on windows (and/or evaluate what's needed to be changed). Unfortunately (or fortunately :smile:) I'm a lot far from any windows machine right now...

emetsger:
@nicolaferraro @rhuss OK!  This PR has been tested on Windows; it does include a minor update to `DockerFileUtilTest.java`, and existing test that failed on the Windows platform due to end-of-line character differences.

Please feel free to squash the PR, and/or let me know if you need anything else!

nicolaferraro:
Great @emetsger, I'll take another look later.

nicolaferraro:
LGTM @emetsger. Great work!
Let's wait for a final check by @rhuss before merging (he may not be available this week).

rhuss:
[merge]

rhuss:
@emetsger thanks !

emetsger:
@rhuss @nicolaferraro thank you folks for a great plugin!  Happy to make a small contribution :)

PR #852, opened 2017-08-28
Fix docker-compose version handling

### LINKED ISSUE #818: Update the required version of docker-compose in DockerComposeConfigHandler to be 2.1
### Description
Update the required version of docker-compose in DockerComposeConfigHandler to be 2.1

### Info
Updating this would allow docker compose to use the healthcheck option

* d-m-p version : 0.21.0
* Maven version (`mvn -v`) : apache-maven-3.0.5
```

Having multiple containers start up with a dependency chain fails.  This is solved with the healthcheck.

Maybe a simple update of :

    private void validateVersion(Map<String, Object> compose, File file) {
        Object version = compose.get("version");
        if (version == null || !version.toString().trim().equals("2")) {
            throw new ExternalConfigHandlerException("Only version 2 of the docker-compose format is supported for " + file);
        }
    }

to

    private void validateVersion(Map<String, Object> compose, File file) {
        Object version = compose.get("version");
        if (version == null || !version.toString().trim().equals("2.1")) {
            throw new ExternalConfigHandlerException("Only version 2.1 of the docker-compose format is supported for " + file);
        }
    }

would help lots.

Many Thanks

rhuss:
And added some tests, too.

Fixes #818.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeConfigHandler.java
+++ b/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeConfigHandler.java
@@ -31,7 +31,7 @@ public class DockerComposeConfigHandler implements ExternalConfigHandler {
 
     // Enable later when issue above is fixed. In the meantime its declared in the components.xml, too
     // @Requirement(role = MavenReaderFilter.class)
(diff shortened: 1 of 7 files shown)


rhuss:
[merge]

PR #875, opened 2017-10-07
Preserve leading whitespace in logs

scoplin:
Changes the log line match pattern to preserve leading whitespace in log
lines by only consuming a single space delimiter between the timestamp
and the message body.


scoplin:
I'm not sure what the original reason for the \\s+ pattern was in the log line matching, but looking at the docker api source code, it looks like it has used a single space to delimit timestamp from body for at least several years now, even through several refactorings.  I tracked it back at least as far as this commit:

    commit 5104b34c443be90644b1ddad12b7b128ed4f168c
    Author: Brian Goff <cpuguy83@gmail.com>
    Date:   Fri Jul 3 09:50:06 2015 -0400

PR #876, opened 2017-10-07
Allow logging prefix to be disabled

scoplin:
Adds a new "disablePrefix" option to the log configuration that will
cause container logs to be presented without a prefix.


rhuss:
Since I'm not super keen to introduce an extra option only for disabling the prefix, I allowed now to use parameters in the prefix. See #890 for the corresponding PR.

For your use case, just use `%z` as prefix which is just an empty string. 'hope this is ok for you, so I'm closing this PR for now.

scoplin:
Yup, that works just fine too.  Tried it out on the v0.23.0 tag and behaved just fine.  Thanks a lot!

PR #883, opened 2017-10-24
Extended Authentication: allow usage of temporary security credentials

sebastiankirsch:
In order to use temporary security credentials provided by the AWS
Security Token Service (AWS STS) to sign requests, the session token has
to be included.


sebastiankirsch:
Thanks, that was fast!
I forgot to ask if there's some integration/system tests that I've been missing. Although it would be a bit artificial to test against anything but AWS itself.

When will this be released?

rhuss:
@sebastiankirsch 0.23.0 is out in the wild now. Thanks for providing this feature !

PR #884, opened 2017-10-26
Fix warning that results when providing new --chown COPY flag

rkhmelichek:
This fixes logging an invalid warning that results when providing the new `--chown` flag to the `COPY` command in a custom Dockerfile.


rkhmelichek:
Forgot to mention, the `--chown` option is part of the release notes from 2017-09-26: https://docs.docker.com/release-notes/docker-ce/

rhuss:
Thanks ! I hope I can review the PR soon ...

rhuss:
Thanks ! I adapted a bit and the fix will be available in the version to be released on this weekend.

PR #897, opened 2017-11-10
[Extended Authentication] Enable usage of system properties for HttpClient 

sebastiankirsch:
This allows using Extended Authentication behind a proxy, as it advices
the HttpClient to use the JVM's standard system properties for
proxies (https://docs.oracle.com/javase/6/docs/technotes/guides/net/proxies.html).

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/ecr/EcrExtendedAuth.java
+++ b/src/main/java/io/fabric8/maven/docker/access/ecr/EcrExtendedAuth.java
@@ -91,7 +91,7 @@ public class EcrExtendedAuth {
     }
 
     CloseableHttpClient createClient() {
-        return HttpClients.createDefault();
+        return HttpClients.custom().useSystemProperties().build();
     }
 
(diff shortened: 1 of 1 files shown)


sebastiankirsch:
To really test this, I'd have to use wiremock or something similar, and I'm not sure if that's really worth the effort - but I can look into this if required...

rhuss:
I don't think its worth the effort as it simply would test http client which we are using. I'm perfectly fine with this PR, thanks ;-)

sebastiankirsch:
Any chance for a new release?
Since we're sitting behind a corporate proxy, we can't use that feature otherwise.

rhuss:
@sebastiankirsch I plan a release until the end of the week. sorry for the delay ...

rhuss:
@sebastiankirsch 0.24.0 has been just released. sorry for the delay.

sebastiankirsch:
Thank you!
Now we can finally get rid of these bash scripts that build & push to AWS ECR.

PR #900, opened 2017-11-17
stop and remove using regex names

transamericamoon:
So we use the maven artifactID as the image name and the version number as the tag for naming our images and containers, the problem is once the version number changes, we can no longer stop or remove old images to deploy the new ones.  

This was a quick fix we put together, i don't know how useful this would be to others, But thought i'd at least throw it out there.  If it's not useful feel free to close.

Usage:
Enabled stop and remove using regex.

```
<properties>
        <docker.stopUsingRegex>true</docker.stopUsingRegex>
        <docker.removeUsingRegex>true</docker.removeUsingRegex>
...
```

Then under the image tags:

  ```
<image>
    <alias>${project.artifactId}_${project.version}</alias>
    <aliasRegex>${project.artifactId}.*${revision.branch}-${revision.type}</aliasRegex>
    <nameRegex>${project.artifactId}.*${revision.branch}-${revision.type}</nameRegex>
....
```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
+++ b/src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
@@ -44,6 +44,11 @@ public final class UrlBuilder {
                 .build();
     }
 
+    public String listImages() {
+        return u("images/json").p("all", true)
+                .build();
+    }
+
     public String containerLogs(String containerId, boolean follow) {
         return u("containers/%s/logs", containerId)
                 .p("stdout",true)
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -11,6 +11,7 @@
   - Enhance @sha256 digest for tags in FROM (image_name:image_tag@sha256<digest>) ([#541](https://github.com/fabric8io/docker-maven-plugin/issues/541))
   - Support docker SHELL setting for runCmds (#1157)
   - Added 'autoRemove' option for running containers (#1179)
+  - Stop and remove using regex names (#900)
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java [hunk 1]
```diff
@@ -81,7 +86,7 @@ public String inspectExecContainer(String containerId) {
     }
 
     public String listContainers(String ... filter) {
-        Builder builder = u("containers/json");
+        Builder builder = u("containers/json").p("all", true);
```
rhuss: Not sure why do need "all" here ?

rhuss:
fyi, just did a rebase and pushed.

transamericamoon:
thanks!

src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
transamericamoon: Yes, this is so even if the container is stopped it can still be removed.  Without "all" only running containers are returned.

rhuss:
@rohanKanojia this is another PR which would need a bit of love. If you feel fancy, you can resolve the conflicts (shouldn't be hard, and then let's try to get this merged).

rhuss:
@rohanKanojia @wrosenuance I think this PR overlaps with #1215 so we should decide which one to continue.

rhuss:
We added now #1215, so this should cover your use case, too. You can expect a release over the weekend.

Thanks a lot for the PR and sorry for being that late ;-)

Issue #862, opened 2017-09-12
mvn initialize docker:start stuck


vinayak03:
when I am trying to start the container using mvn initialize docker:start it stucks does not proceed further.
I am able to create docker image with the same configuration and it push this image in local repo.
but after that when I try to start this it stucks.

* d-m-p version : 0.22.1
* Maven version (`mvn -v`) : 3.3.9
```

```
* Docker version 17.06.2-ce

Added debug file
[debug.txt](https://github.com/fabric8io/docker-maven-plugin/files/1295743/debug.txt)

Any pointers appreciated.

Thanks,
Vinayak

rhuss:
As I see, you are running on Windows. Do you use Docker for Windows here ? Any chance that you could share your project (pom.xml).

TBH, its always hard for me to check Windows issues and it takes a fair amount of time to get and update my Window environment. I hope I can tackle the issue ASAP, but I can't promise anything as I'm currently super busy with other works.

novityFS:
Hi Roland,

I've created a very simple project demonstrating this error. Have a look on:

https://github.com/novityFS/docker-maven-plugin-waitlog-scenario

The readme will give you more detailed info.

I've analyzed that the error is not with the `LogWaitChecker`. Via debug output I could see, that the wait matched the given pattern. Also I think that it's not a race condition. For this reason I've created an alternative version of the `LogWaitChecker` where checking and signaling the match is not runtime dependend. If you want to try this, please have a look on my fork of the plugin:

https://github.com/novityFS/docker-maven-plugin

Kind regards,

Frank.

novityFS:
Dear Roland,

with my projects mentioned above I was able to get some light here. After rearranging the thread synchronisation of the `LogRequestor` (used by `LogWaitChecker`) and the signalling the tasks started by `startImage`, I could reproduce the error in Windows and Linux.

The problem is this piece of code in the method `parseResponse` of the `LogRequestor`:

```Java
try (InputStream is = response.getEntity().getContent()) {
    while (true) {
        if (!readStreamFrame(is)) {
            return;
        }
    }
} catch (IOException e) {
    callback.error("Cannot process chunk response: " + e);
    finish();
} catch (LogCallback.DoneException e) {
    // Can be thrown by a log callback which indicates that we are done.
    finish();
}
```

[...]

PR #904, opened 2017-11-28
#862 - Fix for hanging wait on log

novityFS:
Fixed entity stream closing
Added callback mechanism to signal wait lock condition to main thread including tests
Changed mocked DockerAccess dependency including expectations in LogWaitChecker tests


rhuss:
Thanks a lot and sorry for the long delay.

I did a rebase and some minor cosmetics.

'hope I can get out 0.24.0 with this fix this week.

PR #914, opened 2017-12-27
Fail the build if postStart/preStop commands return a non-zero exit code

isavin:
Extend #272 to check the exit code of the command and if non-zero fail the build.

Use case: start a DB server and in <postStart/> create the db schema which might fail; fail-early when the container is created

I can look further into adding a test for it if the approach looks fine.


rhuss:
Hey, thanks a lot ! (and sorry for the delay).

I just rebased it to the current master and added an option 'breakOnError' for the wait exec condition, which can be set to break the build if an error occurs. Just wanted to add this as I don't want to break backwards compatibility for now.

For version 1.0 we could consider to switch to true by default.

Do you have any chance to try this out and check whether it fits your needs ? thanks!

isavin:
Thank you for the feedback @rhuss! I've tested the changes and made a few fixed.

I've pushed a commit which includes these fixes and posted the [sample project](https://github.com/isavin/issues/tree/master/github.com/fabric8io/docker-maven-plugin/PR914) I've used to test them.

I didn't know what was the expected commit message for a fixup so I've just listed the fixed. Should I use the same msg as your last commit or something else? Please let me know and I'll updated it alternatively if it can be merged into yours as a fixup please feel free to do that.

rhuss:
Thanks, looks good to me ! wrt. to the commit message: There is no standard established yet for dmp, but I'm used to use commitizen like commit logs --> https://commitizen.github.io/cz-cli/ 

I'm not dogmatic here, though ;-)

Let's merge it and many thanks for the contribution ! Feel free to add more if you feel fancy ;-)

isavin:
Thank you for the cz pointer and for merging this!

Issue #917, opened 2017-12-29
Log more information when verbose=true 


kennymacleod:
I'd like to see the plugin logging more information when the verbose=true option is set.

Setting the -X maven flag is not practical, given that this generates insane amounts of junk from the HTTP client (see #868).

Currently, setting that flag only seems to generate some extra logging related to image building, but it would be great if we also saw logging around the commands issued to the docker daemon.

For example, in `DockerAccessWithHcClient#createContainer` the code currently logs the JSON that gets sent to the `/container/create` endpoint, but this is logged at debug level (and for the reasons mentioned above, getting -X debug out isn't an option). This would be a perfect candidate for verbose logging.

PR #918, opened 2017-12-29
#917 additional verbose-level logging on docker PUT/POST/DELETE

kennymacleod:
As described in #917 it'd be most excellent to perform verbose-level logging when we issue commands to the docker daemon. This is extremely useful when reconciling the maven plugin configuration against what the plugin is actually issuing to docker.

Currently, the verbose=true option isn't used for much of anything other than logging image build commands.


rhuss:
@kennymacleod I agree, that -X is definitely too verbose as this is also used by the underlying HTTP service for logging all traffic. However, the POST requests can get quite verbose, too, and tbh I consider them more an internal implementation detail useful mostly for debugging the plugin itself (not a concrete build). E.g. the Docker CLI prints out build steps, but API communication only at higher logging levels. 

Wdyt about introducing an dedicated 'docker.verbose.api'  parameter, or even better using a value for `docker.vebose` like in `-Ddocker.verbose=build,api` ? (with `build` as the default ?)

kennymacleod:
Yeah, I like that idea. Requires more changes to the logging code, obviously, but that would be a nicer solution.

rhuss:
@kennymacleod ok, let's go this road then. Unfortunately I'm a bit loaded these days, but if you feel fancy, I would happily integrate the PR (typically at the weekends).

rhuss:
@rohanKanojia I think the idea mention in https://github.com/fabric8io/docker-maven-plugin/pull/918#issuecomment-355582761 is still valid, but needs to be implemented. It's not a big deal, though, so if you want to give it a try, that would be awesome.

rhuss:
This feature has been implemented with #1192, so closing this PR

PR #922, opened 2018-01-02
More reliably clean up failed/aborted containers

### LINKED ISSUE #921: Stop/remove containers from aborted builds
### Description

When attempting to run a batch-mode dockerized program, it is very useful to use the wait condition to wait for a zero exit status and follow mode on.  In my case, this would be for running a local Hadoop program, but it could also be employed for other dockerizable tools.

However, if the build is interrupted by the user with control-C, instead of killing the container, it leaves it running disconnected in the background.  This appears to be because the wait condition has not yet been satisfied, and so the stop container logic is not active.

The fix, I think, is to simply move the registration of the shutdown hook ahead of starting any of the containers when in follow mode.

### Info

* d-m-p version : 0.23.0
* Maven version (`mvn -v`) : 3.5.2
* Docker version : 1.12.6

1. Start a container with a wait condition that will not be satisfied quickly and follow mode enabled (`docker:run`, for example).
2. Abort the build with control-C.
3. Observe that the container is not stopped or removed.

scoplin:
Moves the shutdown hook in the StartMojo earlier so that containers that
will be cleaned up even if they haven't been fully started per their
wait conditions.  This is especially likely to occur on aborted builds.
[fixes #921]

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/StartMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StartMojo.java
@@ -121,6 +121,12 @@ public class StartMojo extends AbstractDockerMojo {
             // Queue of images to start as containers
             final Queue<ImageConfiguration> imagesStarting = new ArrayDeque<>();
 
+            // Prepare the shutdown hook for stopping containers if we are going to follow them.  Add the hook before starting any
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks ! Makes much sense to me.

Issue #925, opened 2018-01-02
"push" called twice (or more) when image is already on the remote


glennji:
### Description
The `docker:push` goal does two pushes — first the image by name, then iterating over tags and pushing the "tagged" image separately. I don't think this is always necessary.

In my case, I use an image name (in plugin config in my project POM) that includes a repository, but not a version e.g. `<name>docker.internal/${docker.namespace}/pulse/${project.artifactId}</name>`. This, then, defaults to the `latest` tag. In `RegistryService` we push this image to the repo, and then attempt to push the same image in the "for tag in tags" loop below.

I think something like the following makes more sense: we only want to push a second time if the tagged image has a different ID to the one we've already pushed.

[fix-repeated-push.patch.txt](https://github.com/fabric8io/docker-maven-plugin/files/1598957/fix-repeated-push.patch.txt)

### Info
* d-m-p version : 0.23.0
* Maven version (`mvn -v`) : 3.0.3
* Docker version : 17.09.1-ce, build 19e2cf6

glennji:
https://github.com/fabric8io/docker-maven-plugin/pull/926

PR #926, opened 2018-01-02
Issue #925: Push called multiple times for the same image

glennji:
A small fix to the `push` goal that ensures we don't attempt to push the
same image more than once (i.e. if it is tagged multiple times).

Signed-off-by: Glenn J. Mason <glenn@glennji.com>

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/RegistryService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/RegistryService.java
@@ -24,10 +24,12 @@ import org.apache.maven.settings.Settings;
 public class RegistryService {
 
     private final DockerAccess docker;
+    private QueryService queryService;
     private final Logger log;
 
-    RegistryService(DockerAccess docker, Logger log) {
+    RegistryService(DockerAccess docker, QueryService queryService,  Logger log) {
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks for the PR. 

However, I'm not sure that skipping tags to be pushed should be the desired behaviour. Actually (as it should be mentioned in the docs, too, going to verify this), then extra tagged images should be pushed, too. 

There is no extra traffic involved as the created layers are already pushed in the first image. Actually every tagged image should be the _very same_ as the original image, since these are just _extra tags_ (like adding a `2` for a major version tag which always points to the the latest 2 release).

What exactly is your use case ? Would it make sense to introduce an option to avoid pushing all tagged images ?

rohanKanojia:
@glennji: Thanks for contributing :+1:  Could you please rebase with the latest master and push again?

glennji:
Oh, that was QUITE some time ago (and @rhuss may be entirely correct).

If I get a chance, I'll try to get back into it and do the rebase, but if someone else is free then by all means ...

rhuss:
Let's close this 2-year old PR and reopen it, if we get a chance to come back on it if double push is still an issue.

PR #929, opened 2018-01-11
Credentials helper version string fix

### LINKED ISSUE #896: [BUG] Credentials helpers assumed to return a version number
### Description
Starting with 0.23, docker-maven-plugin supports docker credential helpers. By default, they are called with the `version` argument to get the plugin version and use it in a log message.
Not all the helpers, however, return a version number when called with the `version` argument (tested with `docker-credential-gcloud` here, part of google-cloud-sdk).

### Info
* d-m-p version: 0.23.0

giannello:
Don't throw an exception if the credentials helper doesn't return a version string

Fixes #896

Signed-off-by: Giuseppe Iannello <giuseppe.iannello@brokenloop.net>

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/CredentialHelperClient.java
+++ b/src/main/java/io/fabric8/maven/docker/util/CredentialHelperClient.java
@@ -73,7 +73,7 @@ public class CredentialHelperClient {
         public String getVersion() throws IOException {
             execute();
             if (version == null) {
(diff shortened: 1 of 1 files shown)


rhuss:
Looks good ! I would even go to switch to `verbose` level for the logging as it seems to be a valid use case anyway. Let me fix this directly on master.

Thanks !

PR #941, opened 2018-02-12
Add separate properties for build.env.build and run.env

stromnet:
Addresses #386, discussed in #938

Note that it was not possible to use "env.run/build", which would have
matched well with other properties, because we already have the "env"
matcher (which would have matched env.run too, injecting bad
properties).


stromnet:
Re-pushed with updated docs.

stromnet:
Closing in favour for handling in #948.

Issue #931, opened 2018-01-19
[FEATURE] Global configuration for namingStrategy


giannello:
### Description

The current implementation of the plugin requires `namingStrategy` to be defined for each image. A global configuration parameter would be nice to have, to avoid cluttering the POMs.

PR #944, opened 2018-02-14
Add new naming strategy "auto"

marcust:
Enable a new default naming strategy "auto" to create container names based on
a prefix and the image simple name. Adding a global configration for the naming
strategy in order to change it on a global level and thus closing #931.

I don't know if that is a little bit over the top or very narrow minded from our use case point of view, feedback is more than welcome.


rhuss:
Thanks a lot ! I'm currently on the road (and was on PTO last week), but I will have a look as soon as I'm back (early next week)

Issue #869, opened 2017-09-26
`docker.push` should not push the tags when the `docker.skip.tag` is presented


meggarr:
### Description
When I have my `pom.xml` define the tags in both `<name>` and `<tags>`, e.g.
```
<name>myimage:${project.version}.${build.number}</name>
<build>
  <tags>
    <tag>latest</tag>
  </tags>
</build>
```
I want to build the image for default tag only and push the image to remote, via
```
mvn clean package docker:build docker:push \
        -Ddocker.push.registry=<my-registry> \
        -Ddocker.skip.tag=true
```
The it still tries to push the image with tag `latest` while it's skipped in build.

### Info
The `docker.push` should not push the tags when the `docker.skip.tag` is presented.

* d-m-p version : 0.20.2
* Maven version (`mvn -v`) : 3.0.5
```
Apache Maven 3.0.5 (r01de14724cdef164cd33c7c8c2fe155faf9602da; 2013-02-19 21:51:28+0800)
Maven home: /works/apps/apache-maven-3.0.5
Java version: 1.8.0_111, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-oracle/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-79-generic", arch: "amd64", family: "unix"
```
* Docker version : 1.12.6

maximusfloydus:
@rhuss - Have you guys looked at this one?

I'm having this same issue.  In our "snapshot" builds, we create the image and tag, then on releases, we only create the image, but it still tries to push the tag that wasn't created.

rhuss:
Sorry, haven't looked at it yet. I'm currently on the road and quite busy these days, but I will have a look when I'm back.

If you feel fancy (and want accelerate things), I'm happily integrating pull requests, too ;-)

PR #954, opened 2018-02-22
Use docker.skip.tag property on push and remove

maximusfloydus:
When pushing and removing images, the docker.skip.tag property is not
used and can cause errors because the plugin will try to push and/or
remove image tags that don't exist because they were skipped during the
build process.

https://github.com/fabric8io/docker-maven-plugin/issues/869

Fix bug in removing tagged images

Code to remove tagged images was not creating the image name correctly

Signed-off-by: Mark Whelan <mbwhelan@gmail.com>

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -3,7 +3,8 @@
 * **0.24-SNAPSHOT**
   - Fix possible NPE when logging to a file and the parent directory does not exist yet (#911) (#940)
   - Change content type to "application/json" when talking to the Docker daemon (#945)
-
+  - Use docker.skip.tag property on push and remove (#954) ([#869](https://github.com/fabric8io/docker-maven-plugin/issues/869))
+  
 * **0.24.0** (2018-02-07)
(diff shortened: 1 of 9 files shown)


maximusfloydus:
@rhuss - Here is the PR for https://github.com/fabric8io/docker-maven-plugin/issues/869

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -158,6 +158,12 @@
      */
     @Parameter(property = "docker.skip.extendedAuth", defaultValue = "false")
     protected boolean skipExtendedAuth;
+    
+    /** 
+     * Skip building tags
+     */
+    @Parameter(property = "docker.skip.tag", defaultValue = "false")
+    protected boolean skipTag;
```
rhuss: Can we move this to the Mojos only where it makes sense (Build/Remove/Push) ? This leads to some duplication, true, but makes it clear that its not usable in other Mojos.

maximusfloydus:
@rhuss - I made the requested updates to code, docs, and changelog.

PR #959, opened 2018-02-27
Remove TODO comment in DockerAssemblyManager

rohanKanojia:

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
@@ -507,8 +507,6 @@ public class DockerAssemblyManager {
         }
         File artifactFile = new File(target, finalName + "." + project.getPackaging());
         if (artifactFile.exists() && artifactFile.isFile()) {
-            // TODO: Maybe check also that the artifact file is 'young' ? (so that it could be assumed that
-            //       it has been created recently by a 'package' in the same run.)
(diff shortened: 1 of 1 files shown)


src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -506,14 +514,29 @@ private File ensureThatArtifactFileIsSet(MavenProject project) {
             return null;
         }
         File artifactFile = new File(target, finalName + "." + project.getPackaging());
-        if (artifactFile.exists() && artifactFile.isFile()) {
-            // TODO: Maybe check also that the artifact file is 'young' ? (so that it could be assumed that
-            //       it has been created recently by a 'package' in the same run.)
+        if (artifactFile.exists() && artifactFile.isFile() && isRecentlyCreated(artifactFile)) {
```
rhuss: Thanks for the implementation and jumping on dmp, but now I'm not really sure anymore whether this makes sense after all because its maybe to much of magic and there is no really good default value. So this could also make build scripts unpredictable and fragile wrt timing issues. 

@rohanKanojia wdyt ? Maybe we just should remove the TODO comment ;-) ?

rohanKanojia: Hm, okay

rhuss:
Thanks !

Issue #913, opened 2017-12-25
user command not work in order


fairjm:
### Description
I use `<user>` before `<runCmds>`,but `<runCmds>` still run earlier than `<user>`.
### Info  
```
<build>
    <from>jetty</from>
    <assembly>
        <descriptorRef>rootWar</descriptorRef>
        <targetDir>/var/lib/jetty/webapps</targetDir>
    </assembly>
    <env>
        <JAVA_OPTIONS>-Xmx1g</JAVA_OPTIONS>
    </env>
    <user>root</user>
    <runCmds>
        <runCmd>mkdir -p /root/xxx</runCmd>
        <runCmd>touch /root/xxx/yyy</runCmd>
        <runCmd>echo zzz > /root/xxx/yyy</runCmd>
    </runCmds>
</build>
```
build output:  
```
[INFO] DOCKER> [project:0.0.1-SNAPSHOT]: Created docker-build.tar in 3 seconds
[INFO] DOCKER> Step 1/7 : FROM jetty
[INFO] DOCKER> ---> 2938aa22262c
[INFO] DOCKER> Step 2/7 : ENV JAVA_OPTIONS -Xmx1g
[INFO] DOCKER> ---> Using cache
[INFO] DOCKER> ---> d047a2ce0867
[INFO] DOCKER> Step 3/7 : COPY maven /var/lib/jetty/webapps/
[INFO] DOCKER> ---> 272466f0b477
[INFO] DOCKER> Removing intermediate container 82de144a0b48
[INFO] DOCKER> Step 4/7 : RUN mkdir -p /root/xxx
[INFO] DOCKER> ---> Running in 14165585160c
[INFO] DOCKER> [91mmkdir: cannot create directory /root: Permission denied
```  
because parent image use `jetty` as user so I can't create the file  that is required by app in `/root`.  
using Dockerfile is ok.  
Is it on purpose?

* d-m-p version :0.23.0
* Maven version (`mvn -v`) :
```
Apache Maven 3.3.3 (7994120775791599e205a5524ec3e0dfe41d4a06; 2015-04-22T19:57:37+08:00)
Maven home: D:\maven\apache-maven-3.3.3\bin\..
Java version: 1.8.0_101, vendor: Oracle Corporation
Java home: D:\java8\jdk1.8.0_101\jre
Default locale: zh_CN, platform encoding: GBK
OS name: "windows 7", version: "6.1", arch: "amd64", family: "dos"
```
* Docker version :
[...]

PR #963, opened 2018-03-01
 Changing the place where the "USER" command is put in the generated Dockerfile

amanzag:
Right now, the "USER" command is always put last if you generate the Dockerfile from the pom. It isn't very useful this way because normally you want to run some commands after USER.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
@@ -105,6 +105,7 @@ public class DockerFileBuilder {
         addPorts(b);
 
         addCopy(b);
+        addUser(b);
         addWorkdir(b);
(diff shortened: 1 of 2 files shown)


amanzag:
This might fix #913

rhuss:
Unfortunately its not that easy as explained in https://github.com/fabric8io/docker-maven-plugin/issues/913#issuecomment-378725546 because

* it makes sometimes still sense to switch to the user only as last step and run the commands with the original user from the base image
* it for sure breaks backwards compatibility

So please don't mind when I'm closing this PR. Maybe we can together look for a more complete solution like suggested in #913 ?

amanzag:
No worries. Your suggestion to extend runCmds makes sense.
I'd still like to have a POM-based solution and not have to resort to a Dockerfile because POM configuration inheritance is pretty handy.

PR #964, opened 2018-03-06
Fixes #960 Property placeholders are not interpolated when they are only thing in XML

### LINKED ISSUE #960: Property placeholders are not interpolated when they are the only thing in the XML element value
### Description

Property interpolation like `${docker.container.<container>.ip}` only seem to work when they are adjacent to other text in the target XML element.

From the sample pom.xml:
`<CONSUL_HTTP_ADDR>${docker.container.consul.ip}:8500</CONSUL_HTTP_ADDR>`
Produces an environment variable of:
`CONSUL_HTTP_ADDR=172.17.0.2:8500` 

However 
`<CONSUL_HTTP_ADDR>${docker.container.consul.ip}</CONSUL_HTTP_ADDR>` 
Just by itself produces an environment variable:
`CONSUL_HTTP_ADDR=` 
When one would expect it to just contain: `CONSUL_HTTP_ADDR=172.17.0.2` 

Also `${docker.container.<container>.ip}` only seems to work when `<link>`ed with the container being referenced. If you comment out the `<links>` in the `consul-test` image, but leave the `<dependsOn>`, `${docker.container.consul.ip}` fails to be interpolated even when other text is present in the xml element.

### Info

* d-m-p version : 0.23.0
* Maven version (`mvn -v`) :
```
Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T00:58:13-07:00)
Maven home: /opt/apache-maven-3.5.2
Java version: 1.8.0_141, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_141.jdk/Contents/Home/jre
OS name: "mac os x", version: "10.12.6", arch: "x86_64", family: "mac"
```
* Docker version :17.12.0-ce

Sample pom.xml to reproduce the problem
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

    <groupId>test</groupId>
    <artifactId>test</artifactId>
    <packaging>pom</packaging>
    <version>1.0.0-SNAPSHOT</version>

    <build>
        <plugins>
            <plugin>
                <groupId>io.fabric8</groupId>
                <version>0.23.0</version>
                <configuration>
                    <allContainers>true</allContainers>
                    <removeVolumes>true</removeVolumes>
                    <showLogs>true</showLogs>

                    <images>
                        <image>
                            <alias>consul</alias>
                            <name>consul:1.0.6</name>
                            <run>
                                <env>
                                    <CONSUL_BIND_INTERFACE>eth0</CONSUL_BIND_INTERFACE>
                                </env>
                                <wait>
                                    <log>consul: New leader elected</log>
                                    <time>60000</time>
                                </wait>
                            </run>
                        </image>

                        <image>
                            <alias>consul-test</alias>
                            <name>consul:1.0.6</name>
                            <run>
                                <dependsOn>
                                    <container>consul</container>
                                </dependsOn>
                                <links>
                                    <link>consul</link>
                                </links>

                                <env>
                                    <!--suppress MavenModelInspection -->
                                    <!--<CONSUL_HTTP_ADDR>${docker.container.consul.ip}:8500</CONSUL_HTTP_ADDR>-->
                                    <CONSUL_HTTP_ADDR>${docker.container.consul.ip}</CONSUL_HTTP_ADDR>
                                </env>

                                <entrypoint>
                                    <shell>env</shell>
                                </entrypoint>

                                <wait>
                                    <time>60000</time>
                                    <exit>0</exit>
                                </wait>
                            </run>
                        </image>
                    </images>
                </configuration>

                <executions>
                    <!-- Start all the containers -->
                    <execution>
                        <id>start-containers</id>
                        <phase>pre-integration-test</phase>
                        <goals>
                            <goal>start</goal>
                        </goals>
                    </execution>

                    <!-- Stop everything -->
                    <execution>
                        <id>stop-containers</id>
                        <phase>post-integration-test</phase>
                        <goals>
                            <goal>stop</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>

</project>
```

rohanKanojia:
Fixes #960 

Somehow these ${...} parameters are not being picked up by maven
properly when used solely without any suffix string literal, the maven
parameter string comes off as NULL. So adding a workaround for this.
We'll use +${...} as a parameter which would be handled by dmp afterwards

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
+++ b/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
@@ -57,6 +57,12 @@ public class ContainerCreateConfig {
                 String value = entry.getValue();
                 if (value == null) {
                     value = "";
+                } else if(value.matches("^\\+\\$\\{.*\\}$")) {
+                    /*
+                     * This case is to handle the Maven interpolation issue which used
+                     * to occur when using ${..} only without any suffix.
+                     */
+                    value = value.substring(1, value.length());
                 }
                 envProps.put(entry.getKey(), StrSubstitutor.replace(value, mavenProps));
             }
--- a/src/test/java/io/fabric8/maven/docker/access/ContainerCreateConfigTest.java
+++ b/src/test/java/io/fabric8/maven/docker/access/ContainerCreateConfigTest.java
@@ -41,11 +41,14 @@ public class ContainerCreateConfigTest {
         cc.environment(copyPropsToFile(), envMap, Collections.<String, String>emptyMap());
         JSONArray env = getEnvArray(cc);
         assertNotNull(env);
-        assertEquals(3, env.length());
+        assertEquals(6, env.length());
         List<String> envAsString = convertToList(env);
         assertTrue(envAsString.contains("JAVA_OPTS=-Xmx512m"));
         assertTrue(envAsString.contains("TEST_SERVICE=SECURITY"));
         assertTrue(envAsString.contains("EXTERNAL_ENV=TRUE"));
+        assertTrue(envAsString.contains("TEST_HTTP_ADDR=${docker.container.consul.ip}"));
+        assertTrue(envAsString.contains("TEST_CONSUL_IP=+${docker.container.consul.ip}:8080"));
+        assertTrue(envAsString.contains("TEST_CONSUL_IP_WITHOUT_DELIM=${docker.container.consul.ip}:8225"));
     }
 
     @Test
@@ -53,7 +56,7 @@ public class ContainerCreateConfigTest {
         ContainerCreateConfig cc = new ContainerCreateConfig("testImage");
         cc.environment(null, getEnvMap(),Collections.<String, String>emptyMap());
         JSONArray env = getEnvArray(cc);
-        assertEquals(2, env.length());
+        assertEquals(5, env.length());
     }
 
     @Test
@@ -122,6 +125,9 @@ public class ContainerCreateConfigTest {
         Map<String,String> envMap = new HashMap<>();
         envMap.put("JAVA_OPTS", "-Xmx512m");
         envMap.put("TEST_SERVICE", "LOGGING");
+        envMap.put("TEST_HTTP_ADDR", "+${docker.container.consul.ip}");
+        envMap.put("TEST_CONSUL_IP", "+${docker.container.consul.ip}:8080");
+        envMap.put("TEST_CONSUL_IP_WITHOUT_DELIM", "${docker.container.consul.ip}:8225");
         return envMap;
     }
 
--- a/src/main/asciidoc/inc/_implicit-properties.adoc
+++ b/src/main/asciidoc/inc/_implicit-properties.adoc
@@ -1,0 +1,7 @@
+= Implicit properties
+
+There are some implicit configurations in docker maven plugin that are not so straightforward. These are simply workarouds to get docker-maven-plugin's flow right; just to overcome limitations of Maven and other things. Some of these are mentioned below:
+
+* If the only value of the `env` parameter is a docker-maven-plugin internal property which has been set implicitly you have to prefix the property with a single `+` like in `+${docker.container.test.ip}`. This is necessary due to some Maven limitations which simply interpolates a lone, non defined property, to an empty string which can't then be replaced by this plugin after the initial interpolation phase.
+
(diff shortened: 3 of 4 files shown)


rhuss:
@rohanKanojia thanks for picking that up ! Gimme a bit, just on my way to a conference, so hope to review it soonish.

src/main/asciidoc/inc/misc/_env.adoc
```diff
@@ -1,6 +1,22 @@
 
 
-When creating a container one or more environment variables can be set via configuration with the `env` parameter
+When creating a container one or more environment variables can be set via configuration with the `env` parameter. Somehow maven is showing weird behavior in case of interpolating single ${..} parameter; so we have provided a workaround for that. An `env` can take the following forms:
+
+.Env format
```
rhuss: Thanks for caring about the documentation, too ;-)

However I would describe it a bit shorter. Maybe just: 

> If the only value of the env parameter is a docker-maven-plugin internal property which has been set implicitely you have to prefix the property with a single `+` like in `+${docker.container.test.ip}`. This is necessary due to some Maven limitations which simply interpolates a lone, non defined property, to an empty string which can't then be replaced by this plugin after the initial interpolation phase.

Maybe we can list the affected properties here. Or we should put the whole documentation to the place, where the implicit properties are explained (i think this is even better)

src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
```diff
@@ -57,6 +57,12 @@ public ContainerCreateConfig environment(String envPropsFile, Map<String, String
                 String value = entry.getValue();
                 if (value == null) {
                     value = "";
+                } else if(value.contains("+")) {
```
rhuss: wold prefer a `startsWith("+$")` here because otherwise its far too unspecific. (Think about values which can validly contain `+` characters)

Even better, use a regexp `^\+\$\{.*\}$`. This would really only match our use case.

src/test/java/io/fabric8/maven/docker/access/ContainerCreateConfigTest.java
```diff
@@ -122,6 +123,7 @@ private String copyPropsToFile() throws IOException {
         Map<String,String> envMap = new HashMap<>();
         envMap.put("JAVA_OPTS", "-Xmx512m");
         envMap.put("TEST_SERVICE", "LOGGING");
+        envMap.put("TEST_HTTP_ADDR", "+${docker.container.consul.ip}");
```
rhuss: Please add also tests where the + sign is part of a value or when there is something in addition like in `+${docker.container.consul.ip}additional_data` (here, there should be no removal of the +)

src/test/java/io/fabric8/maven/docker/config/RunImageConfigurationTest.java
```diff
@@ -0,0 +1,138 @@
...
+    public void testNetworkingConfig() {
+        runImageConfiguration = new RunImageConfiguration.Builder()
+                .network(null)
+                .build();
+        Assert.assertNotNull(runImageConfiguration);
+        Assert.assertNotNull(runImageConfiguration.getNetworkingConfig());
+    }
+
+    @Test
+    public void testSerialization() {
```
rhuss: what is this test about ?

rohanKanojia: Sorry, I thought I should add a test to check whether these RunImageConfiguration objects are properly serialized into objects or not. If this doesn't look like a sane to you, I would remove it.

Also, Can there be other tests related to this class?

rhuss: well, don't know exactly where this class should be serialized. 

Adding tests are very valuable, but:

* they add maintenance costs, which is ok if the test has some value and tests the right thing. E.g they need to be updated if ImageConfiguration changes
* should not test trivial things which is out of scope of the project. E.g. testing Java serialization (which is the task of the JVM) is not in scope of this project. The same btw for testing setter and getter.

So, testing simple, dump configuration classes does not make sense normally, except when it add some extra logic.

What should be tested for sure are classes which _use_ these configuration. 

E.g. the latest PR about allowing to merge property based configuration with XML based configuration introduces some kind of logic to the configuration classes which could (and probably should) be tested. However, the plan is to move these classes to a `Optional` to handle the situation better to distinguish between default values and values set by the user when it comes to mergign (where default values should be ignored).

tl;dr - yes, please remove this test ;-) (but thanks for caring about testing, not every contribution does this)

rohanKanojia: Sorry for my naive mistakes. I'll update it

rhuss: hey @rohanKanojia no need to apologize ! it's awesome that you picked up the story and did the implementation. Highly appreciated !

Issue #947, opened 2018-02-16
docker:stop hangs when using -Ddocker.keepRunning


jpraet:
As of dmp 0.24.0 docker:stop hangs when using -Ddocker.keepRunning.
In 0.23.0 the build just completed successfully, but keeping the containers running.

Don't know if this change was intentional or not. I don't see it mentioned in the changelog.

rhuss:
This change was not intentional. Does it happen all the time to you or only sporadically ?

When it hangs, could you please trigger a threadump (CTRL-\ or SIGQUIT) and post it please.

jpraet:
```
[INFO] --- docker-maven-plugin:0.24.0:stop (default-stop) @ openshift-poc-openshift ---
2018-02-22 15:54:54
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.151-b12 mixed mode):

"resolver-5" #54 daemon prio=5 os_prio=0 tid=0x0000000022d70000 nid=0x23b8 waiting on condition [0x0000000028bce000]
   java.lang.Thread.State: WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000006c2cad4e0> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
        at java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
        at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1074)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1134)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)

"Thread-21" #52 prio=5 os_prio=0 tid=0x0000000022d6e800 nid=0x18e0 runnable [0x0
000000028a3e000]
   java.lang.Thread.State: RUNNABLE
        at java.net.SocketInputStream.socketRead0(Native Method)
        at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
        at java.net.SocketInputStream.read(SocketInputStream.java:171)
[...]

Issue #767, opened 2017-05-01
waiting on log randomly doesn't work


redijedi:
### Description

The wait on logging will often fail even though the logging statement is plainly visible in the logs. This is most often observed when running in Jenkins. There doesn't appear to be any pattern as running the build manually afterwards will usually fix it, but not always. Sometimes we have to rerun it two or three times in a row until it decides to see the logging.

### Info

* d-m-p version : 0.20.1
* Maven version (`mvn -v`) : 3.3.9
* Docker version : Docker version 1.13.1, build 092cba3
* Simply running the build will fail about half of the time. We have a dockerized DropWizard service starting up and then we wait for the service to come up using the following setting:
```
                                    <wait>
                                        <log>(.*)(org\.eclipse\.jetty\.server\.Server: Started)(.*)</log>
                                        <time>60000</time>
                                    </wait>
```

We've tried various permutations from simply looking for "Started" to the regex noted above. The issue is never resolved. As often as not the following failed build output can be observed:

```
omitted> INFO  [2017-05-01 19:16:45,229] org.eclipse.jetty.server.AbstractConnector: Started admin@3804a9a8{HTTP/1.1,[http/1.1]}{0.0.0.0:8081}
omitted> INFO  [2017-05-01 19:16:45,229] org.eclipse.jetty.server.Server: Started @7161ms
[ERROR] DOCKER> [omitted] "omitted": Timeout after 120039 ms while waiting on log out '(.*)(org\.eclipse\.jetty\.server\.Server: Started)(.*)'
[ERROR] DOCKER> Error occurred during container startup, shutting down...
```

rhuss:
@redijedi Could you please try the following pattern: 

```
(?s)org\.eclipse\.jetty\.server\.Server:\s+Started
```

Does this help ?

redijedi:
@rhuss 

Thanks for the prompt reply! Unfortunately it's more of the same:

```
omitted> INFO  [2017-05-02 17:35:21,668] org.eclipse.jetty.server.handler.ContextHandler: Started i.d.j.MutableServletContextHandler@9687f55{/,null,AVAILABLE}
omitted> INFO  [2017-05-02 17:35:21,674] org.eclipse.jetty.server.Server: Started @6974ms
[ERROR] DOCKER> [omitted] "omitted": Timeout after 120040 ms while waiting on log out '(?s)org\.eclipse\.jetty\.server\.Server:\s+Started'
[...]

PR #965, opened 2018-03-06
Fix deadlock waiting on docker log pattern to match

rkhmelichek:
This fixes a regression introduced in the 0.24.0 version of the plugin.
I have a small test program to demonstrate the problem below.

Normally, Apache HTTP Client will drain an http request upon closing the input stream.
In this case we consume the docker logs in 'follow' mode which will cause the docker daemon to never close the connection on its end.
Thus the http client `close()` call will block forever.
The main thread is responsible for calling `request.abort()`, which will signal Apache HTTP Client to close the socket on it's end and thus the `close()` call can unblock.
Due to the refactoring in 0.24.0, there is a deadlock on this object in LogRequestor.java
```
    // Lock for synchronizing closing of requests
    private final Object lock = new Object();
```
Since the log requestor thread is waiting on `close()` and the main thread cannot reach the code to call `request.abort()`

I believe this is the same issue as #767 .

In the patch, I removed the lock completely, as I made sure that `request.abort()` is only called from the main thread. It doesn't seem necessary anymore.

I have been testing my build of the plugin for a few days and it's been working on Docker for Mac and on our Linux build boxes.
I have briefly tested on Docker for Windows and seems to work as well...but I will run some more tests there.

```java
package io.fabric8.maven.docker.access.log;

import io.fabric8.maven.docker.access.UrlBuilder;
import io.fabric8.maven.docker.access.hc.unix.UnixSocketClientBuilder;
import io.fabric8.maven.docker.util.Logger;
import io.fabric8.maven.docker.util.Timestamp;
import org.apache.commons.lang3.mutable.MutableBoolean;
import org.apache.commons.lang3.mutable.MutableObject;

public class LogWaitDeadlockTest {
    // Replace this with the container ID of a running container on your system.
    private static final String CONTAINER_ID = "35193bc7c489";

    public static void main(String[] args) throws Exception {
        final UnixSocketClientBuilder builder = new UnixSocketClientBuilder("/var/run/docker.sock", 2, new Logger() {
            @Override
            public void debug(String format, Object... params) {

            }

            @Override
            public void info(String format, Object... params) {

            }

            @Override
            public void verbose(String format, Object... params) {

            }

            @Override
            public void warn(String format, Object... params) {

            }

            @Override
            public void error(String format, Object... params) {

            }

            @Override
            public String errorMessage(String message) {
                return null;
            }

            @Override
            public boolean isDebugEnabled() {
                return false;
            }

            @Override
            public boolean isVerboseEnabled() {
                return false;
            }

            @Override
            public void progressStart() {

            }

            @Override
            public void progressUpdate(String layerId, String status, String progressMessage) {

            }

            @Override
            public void progressFinished() {

            }
        });

        final UrlBuilder urlBuilder = new UrlBuilder("unix://127.0.0.1:1/", "v1.18");
        final MutableBoolean stop = new MutableBoolean(false);

        // Plugin normally launches the LogRequestor in it's own thread...
        // The main thread checks every 500ms whether the LogRequestor thread has matched something...

        final MutableObject<LogRequestor> logRequestorHandle = new MutableObject<>();
        new Thread(new Runnable() {
            @Override
            public void run() {
                try {
                    LogRequestor logRequestor = new LogRequestor(builder.buildBasicClient(), urlBuilder, CONTAINER_ID, new LogCallback() {
                        @Override
                        public void log(int type, Timestamp timestamp, String txt) throws DoneException {
                            // Simulate matching on the first log line.
                            if (!stop.booleanValue()) {
                                stop.setTrue();
                                throw new LogCallback.DoneException();
                            }
                        }

                        @Override
                        public void error(String error) {
                            System.err.println(error);
                        }

                        @Override
                        public void open() {
                        }

                        @Override
                        public void close() {
                        }
                    });
                    logRequestorHandle.setValue(logRequestor);
                    logRequestor.run();
                } catch (Exception e) {
                    e.printStackTrace();
                }
            }
        }).start();

        boolean done = false;
        while (!done) {
            if (stop.booleanValue()) {
                LogRequestor logRequestor = logRequestorHandle.getValue();
                logRequestor.finish();
                done = true;
            }
            Thread.sleep(100);
        }

        System.out.println("FINISHED!");
    }
}
```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/log/LogRequestor.java
+++ b/src/main/java/io/fabric8/maven/docker/access/log/LogRequestor.java
@@ -58,9 +58,6 @@ public class LogRequestor extends Thread implements LogGetHandle {
 
     private final UrlBuilder urlBuilder;
 
-    // Lock for synchronizing closing of requests
-    private final Object lock = new Object();
-
     /**
      * Create a helper object for requesting log entries synchronously ({@link #fetchLogs()}) or asynchronously ({@link #start()}.
      *
@@ -86,10 +83,10 @@ public class LogRequestor extends Thread implements LogGetHandle {
         try {
             callback.open();
             this.request = getLogRequest(false);
-            final HttpResponse respone = client.execute(request);
-            parseResponse(respone);
+            final HttpResponse response = client.execute(request);
+            parseResponse(response);
         } catch (LogCallback.DoneException e) {
-            finish();
+            // Signifies we're finished with the log stream.
         } catch (IOException exp) {
             callback.error(exp.getMessage());
         } finally {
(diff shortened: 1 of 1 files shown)


rhuss:
@rkhmelichek sorry, for the delay. I'm just on the road, but hopefully will have a look to the PR later on. My plan is to get to it before this weekend, please ping me again, if I dont make it.

jpraet:
This also fixes #947

rhuss:
thanks ;-) fyi, 0.25.0 has just been released, should bubble up to Maven central very soon.

PR #968, opened 2018-03-10
support multiline labels and empty labels

chonton:
<labels> section should be able to support multiline labels and empty labels


rhuss:
Thanks ! I'm on the road currently but will have a look soonish.

rhuss:
Thanks !

PR #969, opened 2018-03-10
Fixes #957 Simplify configuration

### LINKED ISSUE #957: Simplify configuration
When only a single image should be build with a Dockerfile, the current configuration syntax has too much boilerplate.

E.g.

```xml
<plugin>
  <groupId>io.fabric8</groupId>
  <configuration>
    <images>
      <image>
        <name>${project.groupId}/${project.artifactId}:${project.version}</name>
        <build>
          <dockerFile>Dockerfile</dockerFile>
        </build>
      <image>
    </images>
  </configuration>
</plugin>
```

should be possible with 

```xml
<plugin>
  <groupId>io.fabric8</groupId>
</plugin>
```

when a single `Dockerfile` in the base directory (or `src/main/docker`). The image name defaults to `${project.groupId}/${project.artifactId}:${project.version}` then, but can be configured with a global `imageName`.

Also think about a compatibility mode for the configuration syntax of https://github.com/spotify/docker-maven-plugin or https://github.com/spotify/dockerfile-maven

rohanKanojia:
When only a single image is build with Dockerfile, no need to provide
too much parameters in configuration; build ImageConfiguration dynamically
with the given defaults.
#957

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/BuildMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/BuildMojo.java
@@ -1,6 +1,7 @@
 package io.fabric8.maven.docker;
 
 import java.util.Date;
+import java.util.List;
 
 import io.fabric8.maven.docker.access.DockerAccessException;
 import io.fabric8.maven.docker.config.BuildImageConfiguration;
@@ -9,6 +10,7 @@ import io.fabric8.maven.docker.service.BuildService;
 import io.fabric8.maven.docker.service.ImagePullManager;
 import io.fabric8.maven.docker.service.ServiceHub;
 import io.fabric8.maven.docker.util.EnvUtil;
+import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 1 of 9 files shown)


src/main/java/io/fabric8/maven/docker/BuildMojo.java
```diff
@@ -34,7 +35,11 @@ protected void executeInternal(ServiceHub hub) throws DockerAccessException, Moj
         if (skipBuild) {
             return;
         }
-        for (ImageConfiguration imageConfig : getResolvedImages()) {
+        List<ImageConfiguration> resolvedImages = getResolvedImages();
+        if(resolvedImages == null || resolvedImages.isEmpty()) {
```
rhuss: resolvedImges can't be null otherwise the next for loop would fail with an NPE.

Actually I would either put this into an extra method which returns early in this case or put this into an if-else with the following loop.

src/main/java/io/fabric8/maven/docker/config/ImageConfiguration.java
```diff
@@ -153,6 +154,21 @@ public String initAndValidate(ConfigHelper.NameFormatter nameFormatter, Logger l
         return minimalApiVersion;
     }
 
+    public static ImageConfiguration getDefaultImageConfiguration(MavenProject project, String defaultImageName) {
```
rhuss: Can we keep the dependency on Maven low ? actually I would keep the extraction of the default name from maven coordinates in the Mojos.

The reason is, that eventually, we might want to separate the Maven and non-Maven parts (see https://github.com/fabric8io/fabric8-build for more details of these plans).

src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -119,6 +119,26 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
                 .withExpressionMarkers(delimiters[0], delimiters[1]);
     }
 
+    /**
+     * Fetch the location of docker file in case of simgle configuration
+     *
+     * @return directory containing the Dockerfile
+     */
+    public static String getDockerFileDirectory() {
+        if(checkFileExists("./Dockerfile")) {
```
rhuss: I would not rely on the current working directory (which might e.g. completely elsewhere in a multi-module project). Instead you `${project.baseDir}` (or the MavenProject's equivalent) should be taken into account (which is the absolute path to the module base directory).

src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -119,6 +119,26 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
...
+            return new String(".");
```
rhuss: same here: Best to always work with absolute directories, that will work everywhere. With relative directories this is always fragile and it really depends where this method is used.

src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -119,6 +119,26 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
...
+        } else if(checkFileExists("src/main/docker/Dockerfile")){
```
rhuss: dito

src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -119,6 +119,26 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
...
+            return new String("src/main/docker");
```
rhuss: what's the difference between `new String("src/main/docker")` and `"src/main/docker"` ?

sorry, but that's an oldtimer ;-) and does not make sense at all as String is always immutable. so you can always return the plain string.

rhuss:
@rohanKanojia Thanks ! Looks good to me ....

Issue #970, opened 2018-03-12
postStart exec breakOnError only fails when all containers have started


jpraet:
I have a build with multiple docker images, and have configured startParallel = false, so the containers start sequentially.

The first container (db2express) is configured with a postExec wait action, with breakOnError=true.

I notice because of delay in the logs that the postExec script of the first container is executed before the second container is started. This is as expected. However, when the postExec script fails, I would expect the build to fail immediately. Instead, it first starts the remaining containers, and only once they are all started am I seeing the error from the postExec script.

This is docker-maven-plugin 0.24.0 and docker 0.12.6 on Windows 7.

```
[INFO] DOCKER> [docker-stable.cbss.inet/cbsstools/db2express-c:latest] "db2express-openshift-poc-1.12": Start container d2e3dc52c8e2
[INFO] DOCKER> [rodolpheche/wiremock:2.1.11] "wiremock-openshift-poc-1.12": Start container 7b420510f7eb
[INFO] DOCKER> Pattern 'port' matched for container 7b420510f7eb
[INFO] DOCKER> [rodolpheche/wiremock:2.1.11] "wiremock-openshift-poc-1.12": Waited on log out 'port' 1708 ms
[INFO] DOCKER> [selenium/standalone-chrome:3.8.1] "selenium-openshift-poc-1.12": Start container 71c42c7d2368
[INFO] DOCKER> Pattern 'Selenium Server is up and running' matched for container 71c42c7d2368
[INFO] DOCKER> [selenium/standalone-chrome:3.8.1] "selenium-openshift-poc-1.12": Waited on log out 'Selenium Server is up and running' 3883 ms
[INFO] DOCKER> [docker-snapshots.cbss.inet/openshift-poc:1.12-dev-SNAPSHOT] "jboss-openshift-poc-1.12": Start container da71fdb63e18
[ERROR] DOCKER> Error occurred during container startup, shutting down...
[INFO] DOCKER> [docker-snapshots.cbss.inet/openshift-poc:1.12-dev-SNAPSHOT] "jboss-openshift-poc-1.12": Stop and removed container da71fdb63e18 after 0 ms
[INFO] DOCKER> [selenium/standalone-chrome:3.8.1] "selenium-openshift-poc-1.12": Stop and removed container 71c42c7d2368 after 0 ms
[...]

PR #976, opened 2018-03-18
#970 postStart exec breakOnError should fail fast

jpraet:
Signed-off-by: Jimmy Praet <jimmy.praet@ksz-bcss.fgov.be>

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/StartMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StartMojo.java
@@ -142,21 +142,14 @@ public class StartMojo extends AbstractDockerMojo {
                     // Move from waiting to starting status
                     imagesStarting.add(image);
                     imagesWaitingToStart.remove(image);
+                    
+                    if (!startParallel) {
+                        waitForStartedContainer(hub, containerStartupService, startedContainerAliases, imagesStarting);
+                    }
                 }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -142,21 +142,14 @@ public synchronized void executeInternal(final ServiceHub hub) throws DockerAcce
...
-
-                    updateAliasesSet(startedContainerAliases, imageConfig.getAlias());
-                    exposeContainerProps(hub.getQueryService(), startedContainer);
-
-                    // All done with this image
-                    imagesStarting.remove(imageConfig);
-                } catch (ExecutionException e) {
-                    rethrowCause(e);
+                if (startParallel) {
+                    waitForStartedContainer(hub, containerStartupService, startedContainerAliases, imagesStarting);
```
rhuss: looks good, I have a synctactic suggestion though: I would use a Future<StartedContainer> as argument instead of constainterStartupService() (so fetching the fututer before this method) and don't feed in `imageStarting` but also handle it outside. Within the loop we then could avoid adding the images to imagesStarting for the non-parallel case when we wait for it to start anyway.

But well, properly a matter of tast, so perfeclty fine for me if we merge it like this.

PR #978, opened 2018-03-26
Unbreak merge of LogConfiguration config 

stromnet:
Follow-up on #948, bugfix. LogConfiguration was always created with enabled=true, even if no log properties where set.

Desired functionality: if any config or property in LogConfiguration is set, enabled shall be true.

Also added short note to changelog.


stromnet:
ping @rhuss

PR #989, opened 2018-04-10
fix: Quote labels and env also when having a space

rhuss:


src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
```diff
@@ -244,23 +244,25 @@ private String createKeyValue(String key, String value) {
...
             char c = value.charAt(i);
             switch (c) {
                 case '"':
                 case '\n':
                 case '\\':
-                    escaped = true;
                     // escape the character
                     valBuf.append('\\');
-                    // fall into writing the character
+                case ' ':
```
pepperbot: BLOCKER End this switch case with an unconditional break, return or throw statement. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS128)

PR #994, opened 2018-04-13
Allow retries for pull

bkelley17:
Currently the plugin allows configurable retries for push, but not for pull. This change allows retries to be configured for pull as well.


bkelley17:
I wonder if someone can help me understand what the codecov/patch check is that failed. Is that because coverage of introduced code is not high enough?

rhuss:
@bkelley17 yes, it's because that we generally require that new code should be at least as much tested then the existing code. Not really sure how this is measure as I would expect than if the overall coverage gets increased, then the contribution must be at least have a higher coverage than the rest.

But don't worry too much about it, we also do exceptions from the rule.
Please give me some time to do a review, I hope I can jump on it this week.

thanks for your contribution ...

src/main/java/io/fabric8/maven/docker/BuildMojo.java
```diff
@@ -26,6 +26,9 @@
     @Parameter(property = "docker.skip.build", defaultValue = "false")
     protected boolean skipBuild;
 
+    @Parameter(property = "docker.pull.retries", defaultValue = "0")
+    private int retries;
```
rhuss: Could we please rename this in `pullRetries` for the member configuration ? "retries" alone is a bit unspecific.

rhuss: Same for the other MoJos please and the documentation.

rhuss:
I will work on this PR for the next release (resolving the conflicts) and add the suggestions as I think this is a valuable feature.

rohanKanojia:
Superseded by #1697

PR #1000, opened 2018-04-15
Network aliases management

giafar:
Hi all,
I had some problems to start an external docker-compose.yml because of network aliases configuration mismatch. 
The DockerComposeServiceWrapper does not manage correctly the network name nor the aliases.
There is a unit test, the changes and a docker-compose file to check the behaviour.

Regards Gianluca.

```
version: '2.2'
services:
  service1:
    image: image
    networks:
      network1:
        aliases:
          - alias1
          - alias2
  service2:
    image: image
    networks:
      - network1
  service3:
    image: image
    networks:
      network1:
        aliases:
          - alias1

```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
+++ b/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
@@ -190,12 +190,9 @@ class DockerComposeServiceWrapper {
             	} else if(aliases instanceof LinkedHashMap) {
             		LinkedHashMap<String, ArrayList<String>> map = (LinkedHashMap<String, ArrayList<String>>)aliases;
             		if(map.containsKey("aliases")) {
-            			map.get("aliases").forEach(new Consumer<String>() {
-							@Override
-							public void accept(String t) {
-								ret.addAlias(t);
-							}
-						});
+            			for(String alias : map.get("aliases")) {
+            				ret.addAlias(alias);
+            			}
             		} else {
             			throwIllegalArgumentException("'networks:' Aliases must be given as a linked has map of strings. 'aliases' key not founded");
             		}
--- a/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
+++ b/src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
@@ -1,9 +1,20 @@
 package io.fabric8.maven.docker.config.handler.compose;
 
 import java.io.File;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
-import io.fabric8.maven.docker.config.*;
+import io.fabric8.maven.docker.config.Arguments;
+import io.fabric8.maven.docker.config.ImageConfiguration;
+import io.fabric8.maven.docker.config.LogConfiguration;
+import io.fabric8.maven.docker.config.NetworkConfig;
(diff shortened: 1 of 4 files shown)


rhuss:
Thanks ! but not sure what going on, as when I select https://github.com/fabric8io/docker-maven-plugin/pull/1000/files I still see that everything has changed. Could you try to squash your commits to one and force push it ?

thanks !

giafar:
I don't know what is really going on: https://github.com/fabric8io/docker-maven-plugin/pull/1000/commits/0e05cc905b6ea501ca743a06a3c0f67b5ab2ae5b

I'll squash everything

giafar:
@rhuss Is it possible to merge the PR ?
Thanks in advance.

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
```diff
@@ -182,11 +182,22 @@ NetworkConfig getNetworkConfig() {
...
-                if (!(aliases instanceof List)) {
-                    throwIllegalArgumentException("'networks:' Aliases must be given as a list of string");
-                }
-                for (String alias : (List<String>) aliases) {
-                    ret.addAlias(alias);
+                if (aliases instanceof List) {
+                    for (String alias : (List<String>) aliases) {
+                        ret.addAlias(alias);
+                    }
+                } else if (aliases instanceof LinkedHashMap) {
```
rhuss: wouldn't it be better to check just for `Map` ?

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
```diff
@@ -182,11 +182,22 @@ NetworkConfig getNetworkConfig() {
...
+                    LinkedHashMap<String, ArrayList<String>> map = (LinkedHashMap<String, ArrayList<String>>) aliases;
```
rhuss: `Map<String,List<String>> map` because using the implementation classes as type parameters really causes a tight coupling and you really don't need this.

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
```diff
@@ -182,11 +182,22 @@ NetworkConfig getNetworkConfig() {
...
+                    if (map.containsKey("aliases")) {
+                        for (String alias : map.get("aliases")) {
+                            ret.addAlias(alias);
+                        }
+                    } else {
+                        throwIllegalArgumentException(
+                                "'networks:' Aliases must be given as a linked has map of strings. 'aliases' key not founded");
```
rhuss: why does it have to be a LinkedHashMap ? Just a "... as a map of strings", so I would use this error message:

```
'networks:' No aliases entry found in network config map
```

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
giafar: You are absolutly right

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
giafar: You are right again

src/main/java/io/fabric8/maven/docker/config/handler/compose/DockerComposeServiceWrapper.java
giafar: Message updated

PR #1001, opened 2018-04-16
Add possiblity to use docker.imagePropertyConfiguration with multipe images

stromnet:
By default it works as previous, but if explicit <external> is added to each image, and i.e. setting a custom prefix, it work.

This is useful when i.e. env has `docker.imagePropertyConfiguration` set globally (i.e. settings.xml), and a Maven project happens to have multiple images.

Also add support for docker.imagePropertyConfiguration=skip, for i.e. overriding parents config and disable it (note that setting this in pom <properties> will not override a value in settings.xml, mostly useful to override parent pom).

### DIFF
--- a/src/main/asciidoc/inc/external/_property_configuration.adoc
+++ b/src/main/asciidoc/inc/external/_property_configuration.adoc
@@ -330,19 +330,20 @@ configuration (since version 0.25.0), which can have one of the following values
 [[combining-property-config-externally]]
 .Activating property configuration externally
 It also possible to activate property configuration by setting the property `docker.imagePropertyConfiguration` to a
-valid `property mode`, without adding an `<external>` section.
-This will use any properties with default prefix.
+valid `property mode`, without adding an `<external>` section. The plugin will then use any properties with default `docker.` prefix.
 This can be useful if most of the configuration is specified in XML/POM file, but there
-is need to override certain configuration values without altering the POM file (by i.e. adding this to a parent POM or
+is need to override certain configuration values without altering the POM file (instead add this to a parent POM or
 global settings.xml).
 
-If set globally (i.e. parent POM), but not wanted locally, the property could be overriden to `skip` to disabled it.
-Note that settings.xml properties take precedence over any properties defined in pom.xml.
+If set in parent POM, but not wanted in specific project, the property could be overriden locally with the value `skip`
+to disabled property configuration for that particular project.
+If set in settings.xml however, by Maven design, that value will always take precedence over any properties defined in
+pom.xml.
 
-For configurations with multiple images, using this property will produce an error. All images would use the docker prefix,
-they would all be given the same configuration, and that is probably not intended.
+For configurations with multiple images, using this property will produce an error. All images would then
+use the same `docker.` property prefix, resulting in multiple identical configurations.
 This error can be avoided by adding an explicit <external> configuration element with an alternative <prefix> to all
-images (or at least all but one).
+images (or at least all but one). Since they then have different prefixes, it would be OK.
 
 .Merging POM and property values
 For some fields it may be desired to merge values from both POM and properties. For example, in a certain run environment
--- a/src/main/asciidoc/inc/external/_property_configuration.adoc
+++ b/src/main/asciidoc/inc/external/_property_configuration.adoc
@@ -348,6 +348,8 @@ it does allow you to use the same prefix (even `docker`) on all images. This is 
 a few properties. This only makes sense when `property mode` is _override_ or _fallback_ and image-specific configuration
 are defined in the POM configuration.
 
+For examples, see <<externally-property-config-example, here>>
+
 .Merging POM and property values
 For some fields it may be desired to merge values from both POM and properties. For example, in a certain run environment
 we might want to inject a `http_proxy` environment variable, but we do not want to add this to the POM file.
@@ -405,6 +407,8 @@ For example, to not merge ports, set `docker.ports._combine=replace`, and to ena
 </build>
 ----
 
+
+[[externally-property-config-example]]
 .Example, combining properties and XML/POM configuration
 [source,xml]
 ----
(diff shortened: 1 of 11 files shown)


stromnet:
@rhuss Follow up for my previous work, any chance of merge & release? :)

src/test/java/io/fabric8/maven/docker/config/ConfigHelperTest.java
```diff
@@ -71,11 +69,41 @@ public void externalPropertyActivation() throws MojoFailureException {
             ConfigHelper.validateExternalPropertyActivation(project, images);
             fail();
         }catch(MojoFailureException ex) {
-            assertTrue(ex.getMessage().contains("Configuration error: Cannot use property"));
+            assertEquals("Configuration error: Cannot use property " + ConfigHelper.EXTERNALCONFIG_ACTIVATION_PROPERTY + " on projects with multiple images without explicit image external configuration.",
```
rhuss: I would only check on parts with a essential to provide a meaningful context in the error message. That way the check is more robust on changes wrt rewording of the message. So, I always prefer contains on the essential parts (like `contains(ConfigHelper.EXTERNALCONFIG_ACTIVATION_PROPERTY)`)

src/main/java/io/fabric8/maven/docker/config/ConfigHelper.java
```diff
@@ -72,14 +74,35 @@ public static void validateExternalPropertyActivation(MavenProject project, List
...
+        if(images.size() == 1)
+            return;
+
+        // With more than one image, externally activating propertyConfig get's tricky. We can only allow it to affect
+        // one single image. Go through each image and check if they will be controlled by default properties.
+        // If more than one image matches, fail.
+        int imagesWithoutExternalConfig = 0;
+        for (ImageConfiguration image : images) {
+            if(PropertyConfigHandler.canCoexistWithOtherPropertyConfiguredImages(image.getExternalConfig()))
+                continue;
```
rhuss: can we please agree on adding brackets even around single line conditionals ? This is according to the existing style and less error-prone (believe, I hit that when adding a second line).

Also, would prefer to have spaces between `if(` but I'm less dogmatic here ;-)

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -327,12 +327,22 @@ configuration (since version 0.25.0), which can have one of the following values
...
+valid `property mode`, without adding an `<external>` section.
+This will use any properties with default prefix.
+This can be useful if most of the configuration is specified in XML/POM file, but there
+is need to override certain configuration values without altering the POM file (by i.e. adding this to a parent POM or
+global settings.xml).
+
+If set globally (i.e. parent POM), but not wanted locally, the property could be overriden to `skip` to disabled it.
+Note that settings.xml properties take precedence over any properties defined in pom.xml.
+
+For configurations with multiple images, using this property will produce an error. All images would use the docker prefix,
```
rhuss: maybe highlight "docker" as this is a literal.

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -327,12 +327,22 @@ configuration (since version 0.25.0), which can have one of the following values
...
+.Activating property configuration externally
+It also possible to activate property configuration by setting the property `docker.imagePropertyConfiguration` to a
+valid `property mode`, without adding an `<external>` section.
+This will use any properties with default prefix.
+This can be useful if most of the configuration is specified in XML/POM file, but there
+is need to override certain configuration values without altering the POM file (by i.e. adding this to a parent POM or
+global settings.xml).
+
+If set globally (i.e. parent POM), but not wanted locally, the property could be overriden to `skip` to disabled it.
+Note that settings.xml properties take precedence over any properties defined in pom.xml.
```
rhuss: How would one then could add 'skip' in the pom.xml when the properties in settings.xml has higher prio ? It probably only makes sense when used in pom.xml inheritance.

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -327,12 +327,22 @@ configuration (since version 0.25.0), which can have one of the following values
...
-For single-image configurations it is also possible to active property
-configuration by setting the property `docker.imagePropertyConfiguration` to a
-valid `property mode`, without adding a `<external>` section.
-This will use properties with default prefix.
-This can be useful if most of configuration is specified in XML/POM file, but there
-is need to override certain configuration values without altering the POM file.
+[[combining-property-config-externally]]
+.Activating property configuration externally
+It also possible to activate property configuration by setting the property `docker.imagePropertyConfiguration` to a
+valid `property mode`, without adding an `<external>` section.
```
rhuss: would add here, that the configuration then looks for the default prefix of `docker`.

src/main/asciidoc/inc/external/_property_configuration.adoc
rhuss: maybe: "All images would use the same `docker` prefix ..." ?

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -327,12 +327,22 @@ configuration (since version 0.25.0), which can have one of the following values
...
+This will use any properties with default prefix.
+This can be useful if most of the configuration is specified in XML/POM file, but there
+is need to override certain configuration values without altering the POM file (by i.e. adding this to a parent POM or
+global settings.xml).
+
+If set globally (i.e. parent POM), but not wanted locally, the property could be overriden to `skip` to disabled it.
+Note that settings.xml properties take precedence over any properties defined in pom.xml.
+
+For configurations with multiple images, using this property will produce an error. All images would use the docker prefix,
+they would all be given the same configuration, and that is probably not intended.
```
rhuss: "probably" is a bit weak.

Maybe: "All images would then use the same `docker` property prefix, resulting in multiple identical configurations"

src/main/java/io/fabric8/maven/docker/config/ConfigHelper.java
stromnet: Gah, I thought I configured IntelliJ to do that...

src/main/asciidoc/inc/external/_property_configuration.adoc
stromnet: That comment is more in general for all properties, inherently by maven any properties in settings.xml seems to override my projects properties.

stromnet:
The use case is actually not mixing <external> and docker.imagePropertyConfiguration. 

The use case here was that I globally (through settings.xml) set `docker.imagePropertyConfiguration=override` and `docker.runEnv.http_proxy=xxxx`. But that effectively breaks any of our projects (although very few) with two images in the same d-m-p config.

Thus I don't really have a *need* to add <external> to either of the images, that was more of a way to work around it.. since both images cannot use the same prefix :)

The opposite solution would be, if more than one image exists, we ignore any `docker.imagePropertyConfiguration`, and enforce the caller to explicitly set <external> sections. in case they actually want any merging to happen.. But that hides logic (`docker.imagePropertyConfiguration` suddenly not working if >1 image), so I don't really like that.

stromnet:
All code comments fixed. Also adjusted the external activation & multi-image check a bit: if prefix is explicitly set (to anything), allow any number of images with same prefix. It is useful if most of config is set in POM, but still want to use override for some properties such as `docker.envRun.http_proxy=...` which should apply to all images.

stromnet:
@rhuss any additional feedback? :)

rhuss:
@stromnet just back from Red Hat Summit, tomorrow is docker-maven-plugin day ;-) 'hope to work on as many PRs as possible, including this one.

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -327,12 +327,26 @@ configuration (since version 0.25.0), which can have one of the following values
...
+If set in settings.xml however, by Maven design, that value will always take precedence over any properties defined in
+pom.xml.
+
+For configurations with multiple images, using this property will by default produce an error. All images would then
+use the same `docker` property prefix, resulting in multiple identical configurations.
+This can be overruled by adding an explicit <external> configuration element with an explicit <prefix> to all
+images (or at least all but one). Normally you'd want to use different prefix for each image, but if explicitly set
+it does allow you to use the same prefix (even `docker`) on all images. This is useful in case you just want to share
+a few properties. This only makes sense when `property mode` is _override_ or _fallback_ and image-specific configuration
+are defined in the POM configuration.
```
rhuss: Can we add an example here please, e.g. the example for setting the http proxy for all configured images with a single property (i.e. you use case) ? I think this would help to understand people why it would make sense sometimes to use one property for all configured images.

stromnet: Good point. An example has been added for http_proxy and a single image, as well as an example with two images (using the same prefix).

stromnet:
Improved docs, and added some verbose logging when the external activation property is handled. I was thinking of adding something like "for image <thename>", but that is tricky to fetch since it does not necessarily have a name at that time (we don't yet know if it is from config or from properties).
But at least it hints that a global property is present!

Not sure what more verbose log to add. Perhaps dump the whole image configuration after they have been resolved, or is that more suitable for debug perhaps?

rhuss:
Looks good to me, thanks for the docs and examples. Let's get that merged, also to gather some feedback.

thanks !

stromnet:
Great, thanks! Hoping for a 0.25.3 soon then (and hopefully no more bugs...) :)

PR #1007, opened 2018-04-23
Fix #986 : Create volumes with proper configuration during "docker:start"

### LINKED ISSUE #986: Create volumes with proper configuration during "docker:start"
If a volume is refered to during startup in a bind mount mapping and if this does not exist **and** a volume configuration exists, then this should be taken to create the volume upfront, not relying on Docker to create it implicitly during startup with default params.

This would avoid the necessity to call `volume-create` explicitly before `start` when a custom volume configuration is used.

See #985 for a concrete use case.

The documentation needs also an update to reflect that you cannot only configure real host paths in bind mounts, but also volumes (this is also weak in the Docker documentation itself).

rohanKanojia:
#986


src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -331,6 +336,26 @@ public StartedContainer call() throws Exception {
         return imagesWaitingToStart;
     }
 
+    private void createVolumesAsPerVolumeBinds(ServiceHub hub, List<String> volumeBinds, List<VolumeConfiguration> volumesConfigs)
```
rhuss: can we move this out of the StartMojo into the RunService ? The Mojo should stay as slim as possible.

src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -331,6 +336,26 @@ public StartedContainer call() throws Exception {
...
+        throws DockerAccessException {
+        Map<String, Integer> aIndexMap = new HashMap();
+        for(Integer aIndex = 0; aIndex < volumesConfigs.size(); aIndex++)
```
rhuss: please add braces

src/main/java/io/fabric8/maven/docker/StartMojo.java
```diff
@@ -331,6 +336,26 @@ public StartedContainer call() throws Exception {
...
+            aIndexMap.put(volumesConfigs.get(aIndex).getName(), aIndex);
+
+        for(String aVolumeBind : volumeBinds) {
+            if(aVolumeBind.contains(":")) {
+                aVolumeBind = aVolumeBind.substring(0, aVolumeBind.indexOf(":"));
+            }
+            Integer nVolumeConfigIndex = aIndexMap.get(aVolumeBind);
```
rhuss: What's the difference between an `n` and a `a` prefix for your variables names ? tbh, I would avoid that kind of prefixing as we don't do it nowhere else in this code base. It's always good to adapt to a given codestyle instead of introducing ones own (like avoid braces or spaces after `for` or `if`). This helps to keep the code consistently readable.

src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -516,4 +516,40 @@ public Void call() throws Exception {
...
+        Map<String, Integer> indexMap = new HashMap();
+        List<String> volumesCreated = new ArrayList<>();
+
+        for(Integer index = 0; index < volumesConfigs.size(); index++) {
+            indexMap.put(volumesConfigs.get(index).getName(), index);
+        }
+
+        for(String volumeBind : volumeBinds) {
+            if(volumeBind.contains(":")) {
+                volumeBind = volumeBind.substring(0, volumeBind.indexOf(":"));
```
pepperbot: MAJOR Put single-quotes around ':' to use the faster "indexOf(char)" method. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS3027)

rohanKanojia:
@rhuss : Could you please take a look at it again? I've modified it according to your comments :)

rhuss:
thanks, will do ASAP !

jmayday:
I'm afraid it caused regression:
https://github.com/fabric8io/docker-maven-plugin/issues/1093

rohanKanojia:
@jmayday : Sorry for causing regression :disappointed:

jmayday:
It's OK, I'm fine with downgrading for now. It just took time to analyze what's going on.

rohanKanojia:
@jmayday : bdw, Roland fixed this in #1092 . It would be released soon :-)

rhuss:
np, its was a bad reviewer ;-) 'going to release 0.27.1 this evening ...

PR #1019, opened 2018-05-08
Add option for regenerating certificates after starting Docker Machine

markhalm256:
If using Docker Machine with the AWS EC2 driver, the remote host will have a new IP address every time it is started up again.  This causes the Docker Maven plug-in to fail with an error about certificates being invalid when starting up the machine automatically.  Add an option to automatically regenerate the certificates for the new IP address after startup.  Setting this option to true will allow Docker Machines running in AWS EC2 to be automatically started up into a usable state by the Docker Maven plug-in. (Without this option, such automatically started machines will fail the Maven Docker plug-in invocation complaining about certificates being invalid as the existing certificates are associated with the old IP address)

The changes have been tested with the option both on and off, and are working as expected. Also the documentation has been updated to mention the new option and when it is useful.

### DIFF
--- a/src/main/asciidoc/inc/_global-configuration.adoc
+++ b/src/main/asciidoc/inc/_global-configuration.adoc
@@ -174,6 +174,9 @@ This plugin supports also Docker machine (which must be installed locally, of co
 | *autoCreate*
 | if set to `true` then a Docker machine will automatically created. Default is `false`.
 
+| *regenerateCertsAfterStart*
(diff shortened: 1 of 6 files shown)


markhalm256:
Would be possible to have this merged? The code coverage results are mainly due to Docker Machine code already not being covered, and adding coverage for anything new related to Docker Machine is challenging without any existing testing setup.

PR #1020, opened 2018-05-09
Unbreak setting of entrypoint in exec form when property mode is enabled

stromnet:
Any `<entrypoint><exec><arg>...</arg>` would disappear when any property mode was enabled (but corresponding property not defined as property), due to bad string conversion.

Fix is to get rid of all string conversions dealing with Arguments (which was mostly from test).

Bug in #948

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -413,9 +413,9 @@ public class BuildImageConfiguration implements Serializable {
             return this;
         }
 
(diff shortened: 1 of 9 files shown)


rhuss:
There are currently conflicts because of another merge for health check. 
Also, I think its a good point in time now to switch to Java 8 and lambdas (instead of using `Function`). Wdyt ?

stromnet:
Rebased on master, ready for merge.

I'd rather merge this one before putting effort into changing into Java8, as this bug was a blocker for one of our projects. But yes, that's yet another use-case where lambdas would be nice :)

PR #1023, opened 2018-05-16
Unbreak implicit enabling of log configuration

### LINKED ISSUE #1010: log file output broken in 0.24.0, 0.25.0, 0.25.1, 0.25.2
### Description
With 0.24.0 container starts and output file is created, but plugin does not complete
With 0.25.0, 0.25.1, 0.25.2 container starts and plugin completes, but output file is not created

With 0.23.0 container starts and output file is created, and plugin completes

### Info

* d-m-p version :  0.24.0, 0.25.0, 0.25.1, 0.25.2
* Maven version (`mvn -v`) :
```
Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T00:58:13-07:00)
Java version: 1.8.0_92, vendor: Oracle Corporation
```
* Docker version : 18.03.0-ce-mac60 (23751)

Use mvn docker:start with the following pom.

* Sample project : 
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <groupId>repro</groupId>
  <artifactId>bug</artifactId>
  <version>0.0.1-SNAPSHOT</version>
  <build>
    <plugins>
      <plugin>
        <groupId>io.fabric8</groupId>
        <version>0.23.0</version>
        <configuration>
          <images>
            <image>
              <name>docker.io/vault:0.7.2</name>
              <alias>vault</alias>
              <run>
                <namingStrategy>alias</namingStrategy>
                <network>
                  <name>${docker.network}</name>
                </network>
                <env>
                  <!-- hack to allow older docker hosts to start the vault image -->
                  <SKIP_SETCAP>true</SKIP_SETCAP>
                </env>
                <wait>
                  <time>30000</time>
                  <log>Vault server started! Log data will stream in below:</log>
                </wait>
                <log>
                  <file>vault.log</file>
                </log>
              </run>
            </image>
          </images>
        </configuration>
        <executions>
          <execution>
            <goals>
              <goal>start</goal>
              <goal>stop</goal>
            </goals>
          </execution>
        </executions>
      </plugin>
    </plugins>
  </build>
</project>
```

stromnet:
My previous `enabled` logic change was flawed, and failed to activate LogConfiguration for pom-only configs.

In addition, there was never any property for `log.file`, causing `<file>` in pom to be dropped entirely (as no merge could happen).

Fixes #1010

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/LogConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/LogConfiguration.java
@@ -11,7 +11,7 @@ import org.apache.maven.plugins.annotations.Parameter;
  */
 public class LogConfiguration implements Serializable {
 
-    public static final LogConfiguration DEFAULT = new LogConfiguration(false, null, null, null, null, null);
+    public static final LogConfiguration DEFAULT = new LogConfiguration(null, null, null, null, null, null);
 
     @Parameter(defaultValue = "true")
     private Boolean enabled;
@@ -58,6 +58,25 @@ public class LogConfiguration implements Serializable {
         return enabled;
     }
 
+    /**
+     * If explicitly enabled, or configured in any way and NOT explicitly disabled, return true.
+     *
+     * @return
+     */
+    public boolean isActivated() {
+        return enabled == Boolean.TRUE ||
+                (enabled != Boolean.FALSE && !isBlank());
+    }
+
+    /**
+     * Returns true if all options (except enabled) are null, used to decide value of enabled.
+     *
+     * @return
+     */
+    private boolean isBlank() {
+        return prefix == null && date == null && color == null && file == null && driver == null;
+    }
+
     public String getFileLocation() {
         return file;
     }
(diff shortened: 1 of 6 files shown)


src/main/java/io/fabric8/maven/docker/config/LogConfiguration.java
```diff
@@ -58,6 +58,25 @@ public Boolean isEnabled() {
...
     }
 
+    /**
+     * If explicitly enabled, or configured in any way and NOT explicitly disabled, return true.
+     *
+     * @return
+     */
+    public boolean isActivated() {
+        return enabled == Boolean.TRUE ||
+                (enabled != Boolean.FALSE && !isBlank());
```
rhuss: shouldn't this be better

```java 
enabled == Boolean.TRUE || (enabled == null && !isBlank());
```

which might be easier to understand.

stromnet: They'd be the same yes. But imo it's clearer that we want to explicitly check for false; the thing we want is that if not explicitly disabled, then we let !isBlank() decide if it is "activated" or not.

Issue #1034, opened 2018-05-28
Can't match GCR auth config in ~/.docker/config.json


seboudry:
### Description

DMP can't match an existing auth section in `~/.docker/config.json` to GCR.

### Info

* d-m-p version : `0.25.2`
* Docker version : `18.03.1-ce, build 9ee9f40`

### Reproduce bug

Log to GCR:
```
docker login eu.gcr.io
```

Try to use the auth section in `~/.docker/config.json`:
```
fabric8:push -Ddocker.push.registry=eu.gcr.io/awesome
```

The auth configuration does not match `"eu.gcr.io" != "eu.gcr.io/awesome"`

```
[INFO] --- fabric8-maven-plugin:3.5.38:push (default-cli) @ qs-framework-fabric8-python-samples-echo ---
[DEBUG] F8> AuthConfig: no credentials found
```

rhuss:
"eu.gcr.io/awesome" doesn't look like a registry, its the registry + repository name. Both concepts are clearly separate within d-m-p, so my question to you is why is "awesome" not part of the image name (as specified within the image configuration), using only "eu.gcr.io" as registry name ?

seboudry:
For my use case, repository name can't be part of image name.
For now (with f-m-p), image names consist of last part of groupId and artifactId. So, repository (in this case GCP project) don't have place there. Can have multiple projects in same repository (like in Nexus).
Moreover, I want to be free for where I store my Docker images. Same image can be in multiple registries (CI and production for example). I mention the target registry only at deployment phase.

PS : I use both "repository" and "registry" words on this answer. They maybe not have same signification.

rhuss:
So in the context of d-m-p a "registry" is alwasy a pure registry. It's quay.io, docker.io, or eu.gcr.io. The "repository" is the user name within this registry. It's part of the image name. Please don't confuse them.

So, when pushing to different registry with the same user name, used docker.registry (or the equivalent XML config). If you want to have different user names for you images use something like this in the configuration:

```xml
<images>
  <image>
    <name>${image.user}/myimage</name>
  </image>
</images>
```

and then user `-Dimage.user=itsme` on the commandline or something like this in the properties section.

But dont add the user part to the registry name when using `-Ddocker.push.registry`. That's not a registry anymore then.

seboudry:
To be clear and use f-m-p dialect, in the case of GCR:
[...]

PR #1035, opened 2018-05-28
Issue #1034 : Can't match GCR auth config in ~/.docker/config.json

seboudry:
After searching for exact registry URL match in `~/.docker/config.json`, do the same search with only the hostname part (but keeping scheme if given).


seboudry:
Arf, failed tests on Travis build are OK at home.

rhuss:
Sorry, but the assumption `registry name == hostname` is quite strong, so I wouldn't attach the project id to the registry but to the image name, which makes that kind of lookup unnecessary.

I'm so free to close this PR for now, please reopen if you think we should reconsider this.

seboudry:
Hi @rhuss !
No problem to close this PR as we use combination of `gcloud login` (with a service account) and `docker push` directly for now.
Sorry fot not behing so responsive about this.

PR #1052, opened 2018-07-13
Minor fix on volume configuration example

doandzhi:
Just gave this example a try and maven reported a parsing problem because of this closing tag. :)


rhuss:
Cool, thanks !

PR #1083, opened 2018-09-14
Respect DOCKER_CONFIG when looking for config.json

jakub-bochenski:
Keeping config.json under user home is a problem on CI (e.g. Jenkins)

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
+++ b/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
@@ -414,12 +414,16 @@ public class AuthConfigFactory {
     }
 
     private JSONObject readDockerConfig() {
-        Reader reader = getFileReaderFromHomeDir(".docker/config.json");
+        String dockerConfig = System.getenv("DOCKER_CONFIG");
+        
+        Reader reader = dockerConfig == null
+                    ? getFileReaderFromDir(new File(getHomeDir(),".docker/config.json"))
+                    : getFileReaderFromDir(new File(dockerConfig,"config.json"));
         return reader != null ? new JSONObject(new JSONTokener(reader)) : null;
     }
 
     private Map<String,?> readKubeConfig() {
-        Reader reader = getFileReaderFromHomeDir(".kube/config");
+        Reader reader = getFileReaderFromDir(new File(getHomeDir(),".kube/config"));
         if (reader != null) {
             Yaml ret = new Yaml();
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -439,6 +442,11 @@ private Reader getFileReaderFromHomeDir(String path) {
         } else {
             return null;
         }
+    }    
+    
+    private Reader getFileReaderFromHomeDir(String path) {
+        File file = new File(getHomeDir(),path);
+        return getFileReaderFromDir(file)
```
rhuss: missing semi-colon, but I suggest to inline this method anyway (using directly `getFileReaderFromDir(new File(getHomeDir(),path)` where used.

jakub-bochenski:
@rhuss  sure, and thank for the quick reply -- I just wanted to give you an idea of the change

jakub-bochenski:
@rhuss  done
Now I have a problem with covering this with automated tests. There is no way in Java to override an environment variable. Any suggestions?

rhuss:
I think, that's fine (change is overseeable ;-). Btw, while you are at could also do the same for `KUBECONFIG`, this is the env which overrides the kubelet configuration.

Also, a short entry in the docs would be awesome as well as little entry in the changelog.md

Thanks !

jakub-bochenski:
Done.

Note that I don't have an environment at hand to test the kubernetes change manually.

rhuss:
@jakub-bochenski yes, will do soonish. Thanks !

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -375,7 +363,7 @@ private Map getAuthConfigMapToCheck(LookupMode lookupMode, Map authConfigMap) {
     }
 
     // Parse OpenShift config to get credentials, but return null if not found
-    private AuthConfig parseOpenShiftConfig() {
+    private AuthConfig parseOpenShiftConfig(String useOpenAuthModeProp) throws MojoExecutionException {
```
pepperbot: CRITICAL Refactor this method to reduce its Cognitive Complexity from 79 to the 15 allowed. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS3776)

rhuss:
Thanks ! I udpdated some docs and did some minor refactorings while we were at it ;)

PR #1095, opened 2018-10-01
Fix #1060: Smarter API version handling

### LINKED ISSUE #1060: Smarter API version handling
### Description

As described in https://github.com/fabric8io/fabric8-maven-plugin/issues/1200 the minimal version 1.18 used by d-m-p gets rejected for newer Docker daemons (e.g. those having 1.35). 

Actually we are compliant with the newer versions (its a minimal version after all), so I see two solutions:

* If possible, we don't send an API version. However when we use features which require a minimal version of 1.21 this still will still break.

* We send a test request to the Daemon, and check whether it succeeds. In the return response there's a header `Api-Version:` which holds the actual Docker daemon version. So if this test request because of a too low version, we can just pick up that version and use it as minimal version for us.

rohanKanojia:
#1060 Send a test request to the Daemon, and check whether it succeeds. In the return response
there's a header Api-Version: which holds the actual Docker daemon version. If this test
request because of a too low version, we can just pick up that version and use it as
minimal version for us.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
+++ b/src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
@@ -5,6 +5,7 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.http.HttpHeaders;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpResponseException;
 import org.apache.http.client.ResponseHandler;
@@ -58,6 +59,8 @@ import io.fabric8.maven.docker.util.JsonFactory;
 import io.fabric8.maven.docker.util.ImageName;
 import io.fabric8.maven.docker.util.Logger;
 import io.fabric8.maven.docker.util.Timestamp;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
 
 import static java.net.HttpURLConnection.HTTP_CREATED;
 import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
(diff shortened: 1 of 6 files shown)


src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -376,6 +371,26 @@ private HttpBodyAndStatus inspectImage(String name) throws DockerAccessException
         }
     }
 
+    private static Map<ApacheHttpClientDelegate, UrlBuilder> createHttpClientAndUrlBuilder(String baseUrl, String certPath, Integer maxConnections, String apiVersion, Logger log) throws IOException {
```
rhuss: nice trick with the return value, but as we are using commons-lang anyway, you could also use a `Pair<AppacheHttpClientDelegate,UrlBuilder>` and return an `ImmutablePair` here. I think this is more direct (as a map return value does not really make sense here)

src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -707,4 +722,22 @@ public Object handleResponse(HttpResponse response) throws IOException {
...
+        Map.Entry<ApacheHttpClientDelegate, UrlBuilder> httpClientToUrlBuilderMap = createHttpClientAndUrlBuilder(baseUrl, certPath, maxConnections, DockerAccessFactory.API_VERSION, log).entrySet().iterator().next();
+        ApacheHttpClientDelegate delegate = httpClientToUrlBuilderMap.getKey();
+        UrlBuilder urlBuilder = httpClientToUrlBuilderMap.getValue();
+
+        HttpGet get = new HttpGet(urlBuilder.version());
+        get.addHeader(HttpHeaders.ACCEPT, "*/*");
+        get.addHeader(HttpHeaders.CONTENT_TYPE, "application/json");
+        CloseableHttpResponse response = delegate.getHttpClient().execute(get);
+
+        for(Header header : response.getAllHeaders()) {
```
rhuss: Better call `response.getFirstHeader("Api-Version")` instead of iterating.

src/main/java/io/fabric8/maven/docker/service/DockerAccessFactory.java
```diff
@@ -35,7 +35,8 @@ public DockerAccess createDockerAccess(DockerAccessContext dockerAccessContext)
             DockerConnectionDetector dockerConnectionDetector = createDockerConnectionDetector(dockerAccessContext, dockerAccessContext.getLog());
             DockerConnectionDetector.ConnectionParameter connectionParam =
                     dockerConnectionDetector.detectConnectionParameter(dockerAccessContext.getDockerHost(), dockerAccessContext.getCertPath());
-            String version = dockerAccessContext.getMinimalApiVersion() != null ? dockerAccessContext.getMinimalApiVersion() : API_VERSION;
+            String version = dockerAccessContext.getMinimalApiVersion() != null ? dockerAccessContext.getMinimalApiVersion()
+                    : DockerAccessWithHcClient.fetchApiVersionFromServer(connectionParam.getUrl(), connectionParam.getCertPath(), dockerAccessContext.getMaxConnections(), dockerAccessContext.getLog());
             DockerAccess access = new DockerAccessWithHcClient("v" + version, connectionParam.getUrl(),
```
rhuss: Can't we put the Api detection within the constructor itself ? That way we could also verify whether we can reach the Daemon and we would avoid that static method call (as well maybe even the createHttpClientAndUrlBuilder() method)

src/main/java/io/fabric8/maven/docker/access/UrlBuilder.java
```diff
@@ -187,7 +187,7 @@ private static String encode(String param) {
         }
     }
 
-    private String stripSlash(String url) {
+    public static String stripSlash(String url) {
```
rhuss: I wouldn't expose that as public, but just do the strip directly when querying the API

src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -707,4 +709,13 @@ public Object handleResponse(HttpResponse response) throws IOException {
             return null;
         }
     }
+
+    public String fetchApiVersionFromServer(String baseUrl, ApacheHttpClientDelegate delegate) throws IOException {
+        HttpGet get = new HttpGet(baseUrl + "/version");
```
rhuss:
```java
baseUrl + (baseUrl.endsWith("/") ? "" : "/") + "version"
```

src/test/java/integration/DockerAccessWinIT.java
```diff
@@ -98,7 +97,7 @@ public void testPullStartStopRemove() throws DockerAccessException {
     private DockerAccessWithHcClient createClient(String baseUrl, Logger logger) {
         try {
             String certPath = createDockerConnectionDetector(logger).detectConnectionParameter(null, null).getCertPath();
-            return new DockerAccessWithHcClient("v" + DockerAccessFactory.API_VERSION, baseUrl, certPath, 1, logger);
```
rhuss: Do we still need `DockerAccessFactory.API_VERSION` ?

PR #1125, opened 2018-11-22
Added 'readOnly' option to run configuration to mount container's roo…

bgutjahr:
'docker run' provided the option --read-only to mount a container's root file system as read only.
The docker-maven-plugin didn't provide such an option, so I have added it to be able to run
integration tests with that option.


bgutjahr:
Please let me know if the failed tests on CircleCI are caused by my changed, but to me it doesn't look so. All unit tests worked fine in my local build before I pushed my commit.

PR #1144, opened 2018-12-21
Refactor Mojos to get rid of direct member field access of @Parameter properties

rohanKanojia:
Related to https://github.com/fabric8io/fabric8-maven-plugin/issues/215


src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -194,6 +194,74 @@
 
     private String minimalApiVersion;
 
+    protected String getAutoPull() {
+        return getProperty("autoPull");
```
rhuss: What's about the field which would be set by Maven from the XML configuration ? I don't see where these are used in the getters. These should be actually the fallback values when no property is specified and should mimic the same algorithm that a `@Parameter(property=...)` would do. E.g. I'm not 100% sure what has precedence, the XML value, a system property and a Maven property.

rohanKanojia: Sorry, I didn't consider XML while testing

rhuss: I think this should have been caught by integrations tests, but I don't think we have any ;-)

PR #1157, opened 2019-01-24
Support docker SHELL setting for runCmds

### LINKED ISSUE #1156: runCmds are only executable with /bin/sh set
### Description
I am trying to create a docker image with `RUN` parameters and a linux distribution as `FROM` param, that hasn't linked `/bin/sh`. Dockerfiles support two `RUN` syntaxes to set the executable that should  your command or to use the standard shell (`/bin/sh`). (https://docs.docker.com/engine/reference/builder/#run).

When you are in the situation that `/bin/sh` is not linked you are in a deadlock situation with the plugin. You need a RUN cmd to link the shell but you can't execute the link command.

I can think of two possible solutions:
1. Extend the runCmd section to support the exec syntax
2. Support the SHELL param to set a standard shell without linking it per RUN parameter (https://docs.docker.com/engine/reference/builder/#shell)

If I didn't miss something, I can make a pull request if you tell me which of the options you would prefer.

MeiSign:
Fixes #1156

### DIFF
--- a/src/main/asciidoc/inc/build/_configuration.adoc
+++ b/src/main/asciidoc/inc/build/_configuration.adoc
@@ -90,6 +90,9 @@ A provided `<from>` takes precedence over the name given here. This tag is usefu
 | *ports*
 | The exposed ports which is a list of `<port>` elements, one for each port to expose. Whitespace is trimmed from each element and empty elements are ignored. The format can be either pure numerical ("8080") or with the protocol attached ("8080/tcp").
 
+| *shellParams*
+| Shell to be used for the *runCmds*. It contains *shell* elements which are defining the executable and its params. Whitespace is trimmed from each element and empty elements are ignored.
+
 | *runCmds*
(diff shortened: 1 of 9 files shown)


rhuss:
Thanks ! Looks good to me, except that I would use `<shell>` as top-level element and `<arg>` (as we use it for an <exec> entrypoint, too) as configuration elements.

Could you update these please?

MeiSign:
@rhuss I was thinking about using the args but args give the option to be outputted in exec syntax if I understood the code correctly. The shell command in a dockerfile doesn’t allow that unfortunately: “The SHELL instruction must be written in JSON form in a Dockerfile.”

Should I try to extend the args to work only with Json format with shell?

MeiSign:
I don't mean to rush you but any opinions on my changes?

tcompart:
with which release is this change included or when will it be merged into master?

rhuss:
@tcompart I plan to make a release over the weekend. Sorry for the delay.

tcompart:
thank @rhuss

PR #1170, opened 2019-03-05
Removed check for existing temporary tag

### LINKED ISSUE #838: Cannot temporarily tag because target image already exists
Sometimes when we have target image already existing in system, and so caused the docker-maven-plugin failed to temporarily tag. After checked, found following code.       
```
 if (!name.hasRegistry() && registry != null) {
            if (hasImage(targetImage)) {
                throw new DockerAccessException(
                    String.format("Cannot temporarily tag %s with %s because target image already exists. " +
                                  name.getFullName(), targetImage));
            }
```
I guess it's abnormal to interrupt the docker push with such errors, is it possible to remove the exception or add a parameter to toggle on/off the exception?

jroper:
Fixes #838

There are a few reasons why such a tag might exist:

* Previous run was killed part way through push tag, resulting in the finally block not running
* In my case, was previously manually tagging and pushing, but I didn't clean up old tags, then I switched to using Fabric8.

At any rate, I see no harm in overwriting the tag, since if doing it manually, that's what you'd do anyway.


src/main/java/io/fabric8/maven/docker/access/hc/DockerAccessWithHcClient.java
```diff
@@ -659,12 +659,6 @@ private void doPushImage(String url, Map<String, String> header, HcChunkedRespon
     private String tagTemporaryImage(ImageName name, String registry) throws DockerAccessException {
         String targetImage = name.getFullName(registry);
         if (!name.hasRegistry() && registry != null) {
-            if (hasImage(targetImage)) {
```
rohanKanojia: @jroper : rather than removing exception thrown, can we log it maybe in warn mode? Wdyt?

rohanKanojia:
@jroper : Could update this PR to add a line to CHANGELOG regarding this change?

rhuss:
> At any rate, I see no harm in overwriting the tag, since if doing it manually, that's what you'd do anyway.

the issue is, that this temporary tag will be removed after the push, so that is not something you will expect when the tag already existed because it won't be there after the push.

A possible solution would be to remove the tag only when it has not been present previously. But it will be a different image than before, so this might be an issue, too. 

Not sure how we proceed.

ccampo133:
Was this issue ever completed as part of another PR? Because I notice that this is not merged.

rhuss:
@ccampo133 no it hasn't been added yet (not that I'm aware of)

rhuss:
I tend to close this PR (but feel free to reopen it) as I still have the feeling that is solves the problem on partially and in unexpected ways (see comments above, especially because a temporary tag is removed uncoditionally).

If we want to be more robust for existing tags, we need a PR which also checks for exisiting tags and maybe even keep the old tag's digest and then re-tag agan after the operation.

PR #1176, opened 2019-03-19
Enhance Fix #541: Allow @sha256 digest for tags in FROM

### LINKED ISSUE #541: Allow @sha256 digest for tags in FROM
When I try to use a digest tag in my Dockerfile FROM: such as "FROM: mytomcat@sha256:b61b1b823ec66a2875c44572885fa332990b3aa0055ed50abf7f94837c9f7925" I get this:

[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.15.14:build (docker-build) on project kestrel-docker: Execution docker-build of goal io.fabric8:docker-maven-plugin:0.15.14:build failed: Given repository name 'mytomcat@sha256:b61b1b823ec66a2875c44572885fa332990b3aa0055ed50abf7f94837c9f7925' is invalid:
[ERROR] \* image part 'mytomcat@sha256' doesn't match allowed pattern '(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]_[a-zA-Z0-9])(?:(?:.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]_[a-zA-Z0-9]))+)?(?::[0-9]+)?/)?[a-z0-9]+(?:(?:[._]|__|[-]_)[a-z0-9]+)_(?:/[a-z0-9]+(?:(?:[._]|__|[-]_)[a-z0-9]+)_)?'

I see various mentions in docker re: digest usage and our docker-compose.yml uses it just fine. Do the regex's need to be updated?

mmkonrad:
ImageName.java and respective test class adjusted to be able to handle
image names with tag AND sha256 digest like:
image_name:image_tag@sha256<digest>

Signed-off-by: Marcus Konrad <M.M.Konrad@web.de>

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/ImageName.java
+++ b/src/main/java/io/fabric8/maven/docker/util/ImageName.java
@@ -64,25 +64,35 @@ public class ImageName {
             throw new NullPointerException("Image name must not be null");
         }
 
+        // set digest to null as default
+        digest = null;
+        // check if digest is part of fullName, if so -> extract it
         if(fullName.contains("@sha256")) { // Of it contains digest
             String[] digestParts = fullName.split("@");
             digest = digestParts[1];
-            parseComponentsBeforeTag(digestParts[0]);
-        } else {
-            digest = null;
-            Pattern tagPattern = Pattern.compile("^(.+?)(?::([^:/]+))?$");
-            Matcher matcher = tagPattern.matcher(fullName);
-            if (!matcher.matches()) {
-                throw new IllegalArgumentException(fullName + " is not a proper image name ([registry/][repo][:port]");
-            }
-            tag = givenTag != null ? givenTag : matcher.group(2);
-            String rest = matcher.group(1);
+            fullName = digestParts[0];
+        }
 
-            parseComponentsBeforeTag(rest);
+        // check for tag
+        Pattern tagPattern = Pattern.compile("^(.+?)(?::([^:/]+))?$");
+        Matcher matcher = tagPattern.matcher(fullName);
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException(fullName + " is not a proper image name ([registry/][repo][:port]");
+        }
+        // extract tag if it exists
+        tag = givenTag != null ? givenTag : matcher.group(2);
+        String rest = matcher.group(1);
 
-            if (tag == null) {
-                tag = "latest";
-            }
+        // extract registry, repository, user
+        parseComponentsBeforeTag(rest);
+
+        /*
+         * set tag to latest if tag AND digest are null
+         * if digest is not null but tag is -> leave it!
+         *  -> in case of "image_name@sha256" it is not required to get resolved to "latest"
+         */
+        if (tag == null && digest == null) {
+            tag = "latest";
         }
 
         doValidate();
@@ -172,10 +182,14 @@ public class ImageName {
      * @return full name with original registry (if set) or optional registry (if not <code>null</code>).
      */
     public String getFullName(String optionalRegistry) {
-        if(digest != null) {
-            return getNameWithoutTag(optionalRegistry) + "@" + digest;
+        String fullName = getNameWithoutTag(optionalRegistry);
+        if (tag != null) {
+            fullName = fullName +  ":" + tag;
         }
-        return getNameWithoutTag(optionalRegistry) + ":" + tag;
+        if(digest != null) {
+            fullName = fullName + "@" + digest;
+        }
+        return fullName;
     }
 
     /**
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -5,6 +5,7 @@
   - docs: Correct default image naming
   - close api version http connection ([#1152](https://github.com/fabric8io/docker-maven-plugin/issues/1152))
   - Update to jnr-unixsocket 0.22
+  - Enhance @sha256 digest for tags in FROM (image_name:image_tag@sha256<digest>) ([#541](https://github.com/fabric8io/docker-maven-plugin/issues/541))
 
 * **0.28.0** (2018-12-13)
   - Update to JMockit 1.43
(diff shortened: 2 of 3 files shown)


mmkonrad:
Updated PR without indent issues for [Fabric8 MVN Plugin Issue 1386](https://github.com/fabric8io/fabric8-maven-plugin/issues/1386)

src/main/java/io/fabric8/maven/docker/util/ImageName.java
```diff
@@ -64,25 +64,35 @@ public ImageName(String fullName, String givenTag) {
             throw new NullPointerException("Image name must not be null");
         }
 
+        // set digest to null as default
+        digest = null;
```
rohanKanojia: I don't understand the intention of this change, why do you need tag when you're not passing it in REST call to docker daemon?

rohanKanojia: Do you need it only for ImageName?

mmkonrad: Do you mean why only ImageName.java ist modified but not e.g. UrlBuilder.java?

rohanKanojia: Yeah, since this would only have an effect on `ImageName` object and this won't be propagated to api server. Maybe I misunderstood your use case earlier and delayed it because I wasn't sure how docker api server handles digests.

mmkonrad: True. I did not dive deep into the entire architecture. (i do not have enough time) As I said in the F8 MVN thread, it is not final. 
I just wanted to make a start towards solving our/my request instead of just asking all the time when it will be done.
If you have some hints where to look and probably some docs about the missing parts, that would be great.

mmkonrad: As far as the (by us) required feature is concerned, the tool must be able to create docker images using this pattern:  "registry/repository/docker_image_name:docker_image_tag@sha256<digest>"

mmkonrad: I tested my changes with the dockerfile sample of this project and am able to build docker images using `FROM my-registry.com/my-repository/jetty:9.4.14-jre8@sha256:36bf904c5a2ce97da398cefacf725a870c388c022ba07ef64d1cd516870e49a8`

But I still do not know what else I affect with the proposed changes. All tests succeed so far, but it that is no proof of absence.....

rohanKanojia: Right now we're ignoring tag and sending image digest in place of tag, need to check in docker api on how to pass both(but I'm still not sure why do we need both)

mmkonrad: I am not sure whether others have the necessity for this.
But we do.
We have multiple applications with each several docker images being involved.
We have the demand, to know (and control) exactly what base images are used and whether we adapt newer versions/patches/updates or not.
In case we decide to adapt a change in a base image, follow-up processes will be triggered.
Going with the latest image does not work for us.
Instead we will use an open-source tool called "renovate" to pin the hash digest to the used base images.
This tool then will then check if there is any change to the base image, based on the hash digest that is pinned to the image and its tag.
This can be an update within the same tag e.g. version 1.2.3.4-alpine, a minor patch and also major patches.
Version tags like 1.2.3.4-alpine may consist of different version even though it is displayed to the outside as version 1.2.3.4-alpine.
Even these small changes within the same version may cause issues and side-effects to us. 
But then we may also have to adapt to certain patches and updates at some time. 
Right now multiple use-cases exist for us.
Relying on image tags AND hash digests together, gives us the possibility to meet the demand of controlling what updates/patches/changes we want to adapt.
Additionally fabric8 is a major component in the build processes of our docker images.
As a result we need the docker-maven-plugin to allow tags AND hash digest in the FROM statement.

Maybe there is a better solution out there, but time and available resource do not allow it right now to rethink and reconstruct all of our processes just to meet that demand.
Therefore it is the convenient way to enable the docker-maven-plugin to handle tags and hash digests together in the FROM statement.

schuch: @mmkonrad so the only reason for specifying the tag and the digest together is that a tool like https://github.com/renovatebot/renovate can automatically update the digest of a tag, if a tag changed, right?

Imho when pulling the "pinned" image only the digest needs to be sent to the registry. 

According to the author of renovatebot, when specifiying tag and digest in an image reference, the tag will be indeed ignored:
> One useful trick know is that you don’t have to remove the tag if you want to add a digest. If both are present then the tag is ignored, so you can leave it in for human readability, e.g
(refering to https://renovatebot.com/blog/docker-mutable-tags)

But i don't know if the pulling client is handling this by not sending the tag or if the registry just ignores it. I did not find any specification for that. May be we can try to look how the docker daemon behaves on this by looking at pull requests with a man-in-the-middle proxy.

mmkonrad: @schuch Thank you for pointing this out. 
Human readability is indeed the key factor. 

> If a source tag is updated with a new semver version, you will receive a PR with a new FROM line that looks something like FROM node:8.10.1@sha256:eks9abc802b2a8464b3bfc1f8c3c409f89e9b70a31f1dccce70bd14mxwris. i.e. tag has changed and digest has changed. If only the digest is updated, you would see that too (**and the PR will tell you that it’s a digest update, not a version update**).

@rohanKanojia Sorry for not stating it so clearly from the beginning

rohanKanojia: @mmkonrad: Hi, Sorry for late reply(I saw this in notifications but somehow forgot to reply). Are you trying to say that you're okay with tag to stay just in ImageName class(like you've done here)? I think we're good to merge this then ;-) . I think we should handle this case as well and simply ignore tag even if it's parsed properly.

rohanKanojia:
@rhuss : Could you please share your thoughts? I think we can merge this.

mmkonrad:
Thanks a lot!

PR #995, opened 2018-04-13
Added 'contextDir' configuration option. Used to specify Docker build context

blaketsk:
There is no possibility to put Dockerfile outside of build context now (Dockerfile's parent directory is used as build context).
It can be useful to have 'contextDir' property in some cases.
For example, currently I'm working on multi-module maven project. And I want to use single Dockerfile for multiple modules without copying it at every module.. I found a workaround for that: copy Dockerfile to 'target' directory using maven-resources-plugin..
It would be more convenient to put Dockerfile at parent-module, and specify configuration options like:
`<dockerFile>${project.parent.basedir}/Dockerfile</dockerFile>`
`<contextDir>${project.basedir}`

PR #1189, opened 2019-04-06
Added 'contextDir' configuration option. Used to specify Docker build context

rohanKanojia:
This is the continuation of #995

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
@@ -130,7 +130,7 @@ public class DockerAssemblyManager {
                 archiveCustomizers.add(new ArchiverCustomizer() {
                     @Override
                     public TarArchiver customize(TarArchiver archiver) throws IOException {
-                        DefaultFileSet fileSet = DefaultFileSet.fileSet(dockerFile.getParentFile());
+                        DefaultFileSet fileSet = DefaultFileSet.fileSet(buildConfig.getAbsoluteContextDirPath(params));
                         addDockerIncludesExcludesIfPresent(fileSet, params);
                         // Exclude non-interpolated dockerfile from source tree
                         // Interpolated Dockerfile is already added as it was created into the output directory when
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -19,6 +19,13 @@ public class BuildImageConfiguration implements Serializable {
     public static final String DEFAULT_CLEANUP = "try";
 
     /**
+     * Directory is used as build context.
+     * If not specified, dockerfile's parent directory is used as build context.
+     */
+    @Parameter
+    private String contextDir;
+
+    /**
      * Directory holding an external Dockerfile which is used to build the
      * image. This Dockerfile will be enriched by the addition build configuration
      */
@@ -26,7 +33,7 @@ public class BuildImageConfiguration implements Serializable {
     private String dockerFileDir;
 
     /**
-     * Path to a dockerfile to use. Its parent directory is used as build context (i.e. as <code>dockerFileDir</code>).
+     * Path to a dockerfile to use.
      * Multiple different Dockerfiles can be specified that way. If set overwrites a possibly givem
      * <code>dockerFileDir</code>
      */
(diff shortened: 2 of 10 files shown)


rohanKanojia:
This is still wip, I haven't tested it properly and haven't updated docs yet.

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -130,7 +130,9 @@ public File createDockerTarArchive(String imageName, final MojoParameters params
                 archiveCustomizers.add(new ArchiverCustomizer() {
                     @Override
                     public TarArchiver customize(TarArchiver archiver) throws IOException {
-                        DefaultFileSet fileSet = DefaultFileSet.fileSet(dockerFile.getParentFile());
+                        DefaultFileSet fileSet = DefaultFileSet.fileSet(buildConfig.getContextDir() != null
```
rhuss: According to your change below, `buildConfig.getContextDir()` can never be null.

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -548,20 +576,48 @@ private void initDockerFileFile(Logger log) {
...
             File dFile = new File(dockerFile);
-            if (dockerFileDir == null) {
+            if (dockerFileDir == null && contextDir == null) {
                 return dFile;
             } else {
-                if (dFile.isAbsolute()) {
-                    throw new IllegalArgumentException("<dockerFile> can not be absolute path if <dockerFileDir> also set.");
+                if(contextDir != null) {
+                    if(dockerFileDir != null) {
+                        log.warn("using contextDir as it has more precedence for Dockerfile");
```
rhuss: Precedence over what ? Suggestion: "Both contextDir (%s) and deprecated dockerFileDir (%s) are configured. Using contextDir." (with inserting the concrete values)

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -548,20 +576,48 @@ private void initDockerFileFile(Logger log) {
...
+                    }
+
+                    if (dFile.isAbsolute()) {
+                        try {
+                            Files.copy(dFile.toPath(), Paths.get(contextDir));
```
rhuss: I don't think that the dockerfile should be copied int the `contextDir` as this is typically a source code controlled directory. Intead it should be copied into the build dir where the context is copied to for creating the tar archive to be send to the Docker daemon.

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -548,20 +576,48 @@ private void initDockerFileFile(Logger log) {
...
+                    }
+                    log.warn("dockerFileDir parameter is deprecated, please migrate to contextDir");
+                    return new File(dockerFileDir, dockerFile);
                 }
-                return new File(dockerFileDir, dockerFile);
             }
         }
 
+
+        if (contextDir != null) {
```
rhuss: I recommend to not duplicate that check but just add it outside the conditionals (and use the comment as mentioned above)

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
rohanKanojia: Could you please tell me where are you creating tar archive? Is it here:

https://github.com/fabric8io/docker-maven-plugin/blob/dc82c82360afca7451b98d89f046204b506bec1e/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java#L130-L140

rhuss: It's here where the Dockerfile is copied to its target directory: https://github.com/fabric8io/docker-maven-plugin/blob/dec36f1b75f86fa462e125172a183b7355ec460c/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java#L196-L202

rohanKanojia: okay, thanks

rohanKanojia: to bo honest, I am still confused. Aren't we copying `Dockerfile` here already into build directory? Could you please tell me also where is `dockerFileDir`/`contextDir` getting copied? do we need to copy `Dockerfile` just inside build or inside the `contextDir`?

rohanKanojia: hey, we were not done with this comment? Is it okay to merge it without addressing it?

PR #1191, opened 2019-04-07
support the new docker build --pull option

rohanKanojia:
Extension of PR: #626


rhuss:
@rohanKanojia not sure what the status of this PR is ? Should we close is ? Looks like it got out-of-sync quite a bit.

rohanKanojia:
Let me check it on Monday, If updating it is painful, then I would close this. Otherwise I would try to modify it to accomodate that `autoPull` option

rhuss:
@rohanKanojia no worries, I probably don't have a chance to work on dmp before my holidays. So take your time ;-)

src/main/java/io/fabric8/maven/docker/BuildMojo.java
```diff
@@ -83,8 +84,18 @@ protected void buildAndTag(ServiceHub hub, ImageConfiguration imageConfig)
...
     private String determinePullPolicy(BuildImageConfiguration buildConfig) {
-        return buildConfig != null && buildConfig.getImagePullPolicy() != null ? buildConfig.getImagePullPolicy() : imagePullPolicy;
+        if (buildConfig != null) {
+            if (buildConfig.getImagePullPolicy() != null) {
+                return buildConfig.getImagePullPolicy();
+            } else if (buildConfig.getAutoPull() != null) {
+                return buildConfig.getAutoPull().equalsIgnoreCase("TRUE") ?
+                        ImagePullPolicy.IfNotPresent.name() :
+                        ImagePullPolicy.Never.name();
+            }
```
rhuss: It was a long time ago, but what actually why do we need both `imagePullPolicy` and `autoPull` as properties ? It looks like they are really doing the same and autoPull could always be interfered from an `imagePullPolicy`.

src/main/java/io/fabric8/maven/docker/service/BuildService.java
```diff
@@ -137,6 +137,7 @@ protected void buildImage(ImageConfiguration imageConfig, MojoParameters params,
         // auto is now supported by docker, consider switching?
         BuildOptions opts =
                 new BuildOptions(buildConfig.getBuildOptions())
+                        .autoPull(buildConfig.getAutoPull())
```
rhuss: Here we could put in the value calculated from image pull policy ?

src/main/asciidoc/inc/build/_configuration.adoc
```diff
@@ -71,6 +71,9 @@ endif::[]
 
 A provided `<from>` takes precedence over the name given here. This tag is useful for extensions of this plugin like the https://maven.fabric8.io[fabric8-maven-plugin] which can evaluate the additional information given here.
 
+| *autoPull*
+| Whether we need to pull a newer version of image.
```
rhuss, line 75:
```suggestion
| If the base image should be automatically pulled if not already present
```

rhuss (commented):
Thanks, the PR makes totally sense, but I'm still wondering why we just can use the `pullPolicy` every instead of having an additional `autoPull`. Historically autoPull was first but IMO serve the same purpose than imagePullPolicy. So instead of introducing yet another autoPull field maybe we should just ensure that `imagePullPolicy` works properly on the BuildConfiguration ?

But maybe I misunderstood something and autoPull is required in addition to imagePullPolicy.

rhuss (commented):
Thanks, the PR makes totally sense, but I'm still wondering why we just can use the `pullPolicy` every instead of having an additional `autoPull`. Historically autoPull was first but IMO serve the same purpose than imagePullPolicy. So instead of introducing yet another autoPull field maybe we should just ensure that `imagePullPolicy` works properly on the BuildConfiguration ?

But maybe I misunderstood something and autoPull is required in addition to imagePullPolicy.

PR #1192, opened 2019-04-07
 Change docker.verbose parameter to String to handle multiple logging levels

### LINKED ISSUE #917: Log more information when verbose=true

rohanKanojia:
Fix #917


src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -78,7 +78,13 @@ public void info(String message, Object ... params) {
 
     /** {@inheritDoc} */
     public void verbose(String message, Object ... params) {
-        if (verbose) {
+        if (verbose != null && verbose.contains("build")) {
```
rhuss: I would enable this also when this parameter is set to true, so that the old behaviour with `-Ddocker.verbose=true` still works (and which would switch on just the 'build' verbose part).

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -106,7 +112,7 @@ public boolean isDebugEnabled() {
     }
 
     public boolean isVerboseEnabled() {
-        return verbose;
+        return !verbose.isEmpty();
```
rhuss: I think if you use -Ddocker.verbose=false then verbose logging should be disabled as well. Please add a check for "false", too.

src/main/java/io/fabric8/maven/docker/util/Logger.java
```diff
@@ -26,13 +26,22 @@
...
     void verbose(String format, Object ... params);
 
+    /**
+     * Verbose API messages
+     *
+     *
+     * @param format verbose message format
+     * @param params parameter for formatting message
+     */
+    void verboseApi(String format, Object ... params);
```
rhuss: Could we generalise this ? I.e. to add an an enum LogVerboseCategory with elements "Build" and "Api", and a `verbose(LogCategory cat, String format, .... params)` method ?

That would make it much easier to add more than the Api for logging.

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -28,7 +28,7 @@
     private final String prefix;
     private final boolean batchMode;
 
-    private boolean verbose;
+    private String verbose = "build";
```
rhuss: better call it "verboseModes" and maybe make it an array of LogVerboseCategory enums ? see below.

src/main/java/io/fabric8/maven/docker/util/Logger.java
```diff
@@ -26,12 +26,13 @@
...
     /**
-     * Verbose message
+     * Verbose message for build
      *
+     * @param logVerboseCategory debug level for logging
      * @param format verbose message format
      * @param params parameter for formatting message
      */
-    void verbose(String format, Object ... params);
+    void verbose(AnsiLogger.LogVerboseCategory logVerboseCategory, String format, Object ... params);
```
rhuss: We should move the category to this interface. It's a bad smell to have the interface to reference to one of its implementations.

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -48,17 +53,36 @@
...
-    public AnsiLogger(Log log, boolean useColor, boolean verbose, boolean batchMode) {
+    public AnsiLogger(Log log, boolean useColor, String verbose, boolean batchMode) {
         this(log, useColor, verbose, batchMode, DEFAULT_LOG_PREFIX);
     }
 
-    public AnsiLogger(Log log, boolean useColor, boolean verbose, boolean batchMode, String prefix) {
+    public AnsiLogger(Log log, boolean useColor, String verbose, boolean batchMode, String prefix) {
         this.log = log;
-        this.verbose = verbose;
+        this.isVerbose = checkBackwardVersionValues(verbose); // For backward compatibility
```
rhuss: What happens when "verbose" is null ? It doesn't look like that the code really checks this case. It should be that in this case verbose logging should be disabled.

PR #1194, opened 2019-04-09
Issue #1185 Prevent docker:save failing when no images are defined.

wrosenuance:
Address issue #1185 by having docker:save handle POMs with no images with a build configuration.

Signed-off-by: William Rose <william.rose@nuance.com>

## BEFORE
$ mvn io.fabric8:docker-maven-plugin:0.28.0:save
[INFO] Scanning for projects...
[INFO]
[INFO] -------------------< io.fabric8:docker-maven-plugin >-------------------
[INFO] Building docker-maven-plugin 0.28-SNAPSHOT
[INFO] ----------------------------[ maven-plugin ]----------------------------
[INFO]
[INFO] --- docker-maven-plugin:0.28.0:save (default-cli) @ docker-maven-plugin ---
[ERROR] DOCKER> More than one image with build configuration is defined. Please specify the image with 'docker.name' or 'docker.alias'.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 2.227 s
[INFO] Finished at: 2019-04-09T00:07:12-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.28.0:save (default-cli) on project docker-maven-plugin: More than one image with build configuration is defined. Please specify the image with 'docker.name' or 'docker.alias'. -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException

## AFTER
$ mvn io.fabric8:docker-maven-plugin:0.28-SNAPSHOT:save
[INFO] Scanning for projects...
[INFO]
[INFO] -------------------< io.fabric8:docker-maven-plugin >-------------------
[INFO] Building docker-maven-plugin 0.28-SNAPSHOT
[INFO] ----------------------------[ maven-plugin ]----------------------------
[INFO]
[INFO] --- docker-maven-plugin:0.28-SNAPSHOT:save (default-cli) @ docker-maven-plugin ---
[INFO] DOCKER> No images have a build configuration defined: save skipped
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[...]

PR #1196, opened 2019-04-09
Prevent docker:save failing when no images are defined.

rhuss:
Followup and refactoring of #1194 // @wrosenuance


rhuss:
If you agree with the changes, I would merge this PR and close #1194 .

rhuss:
@wrosenuance cool, thanks !

Issue #1195, opened 2019-04-09
Test cases fail on Windows


wrosenuance:
### Description
Existing test cases use relative paths in dockerFile, but the EnvUtil.isValidWindowsFileName() test applied to this field fails if there are slashes or backslashes.

### Info
This is only reproducible if running the test suite on Windows as EnvUtil.isWindows() is called before EnvUtil.isValidWindowsFileName(). This also means the restriction that the dockerFile attribute only be a file name and not a path is only applied to Windows.

* d-m-p version : 0.29-SNAPSHOT
* Maven version (`mvn -v`) : 3.5.4
```

```
* If it's a bug, how to reproduce : run mvn clean test on latest checkout on a Windows host

rhuss:
In fact, the dockerFile should be possibly a path, not only a plain filename.

Thanks for reporting this. @rohanKanojia would this be an issue you could have a look if time permits ?

rohanKanojia:
Hmm, let me check it on my brother's laptop.

wrosenuance:
@rohanKanojia I created a PR #1198 that resolves this issue for me by taking out the Windows-specific validations that are inconsistent with other platforms. I think it should be enough to rely on the File() constructor and other java.io class code to validate paths in a cross-platform way.

There was one test case that tried to create an absolute path for the dockerFile parameter by passing "/Dockerfile": I updated this test case to use new File("Dockerfile").getAbsolutePath() so that it generates a valid absolute path for the platform.

rhuss:
Actually this check was introduced with https://github.com/fabric8io/docker-maven-plugin/pull/834 . Not sure anymore what this actually fixed ;-)

rohanKanojia:
@wrosenuance : I tried building this on windows but surefire failed before test suite could run: https://pastebin.com/SiK8wBMT. Could you please tell me how you tested it on windows?

rohanKanojia:
bdw, has anyone considered integrating Azure pipelines into repo?  I've seen on Quarkus repository that they're using it, here is a sample https://dev.azure.com/quarkus-ci/quarkus/_build/results?buildId=912 . I think it's free for OSS projects

wrosenuance:
> @wrosenuance : I tried building this on windows but surefire failed before test suite could run: https://pastebin.com/SiK8wBMT. Could you please tell me how you tested it on windows?

I just checked out the branch, opened the Git bash from Git for Windows, and ran:
[...]

PR #1198, opened 2019-04-09
Issue #1195 Remove Windows file name validations.

wrosenuance:
The Windows file name validations were inconsistent with other platform
behaviour and impede the creation of cross-platform POMs.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
@@ -580,10 +580,6 @@ public class BuildImageConfiguration implements Serializable {
         }
 
         if (dockerFile != null) {
-            if (EnvUtil.isWindows() && !EnvUtil.isValidWindowsFileName(dockerFile)) {
-                throw new IllegalArgumentException(String.format("Invalid Windows file name %s for <dockerFile>", dockerFile));
-            }
-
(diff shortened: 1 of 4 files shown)


rhuss:
Thanks for the PR ! However, I'm pretty sure we added that for some reason (i.e. because of a bug on Windows). Let me find the corresponding issue ...

rhuss:
Actually this change came in with https://github.com/fabric8io/docker-maven-plugin/pull/834 where it was exactly introduced to *fix* test failures on windows. So now I'm not sure anymore ;-)

wrosenuance:
Yeah, I think it may have been the wrong fix for the test case. The one test case that fails without this check is the one I changed in [here](https://github.com/fabric8io/docker-maven-plugin/pull/1198/commits/7b6cb75c26f563eb70d2ad94e0ca66f36f79b92e#diff-d4e798be801faf7be6bbb66640c20d2f), and I think it's because the test tries to create an absolute path but did so in a non-portable way. Rather than making the test case create an absolute path portably, the workaround was to trigger an error for a different reason.

rhuss:
Sounds legit. @rohanKanojia wdyt ? From my side, I think we can merge.

rohanKanojia:
+1

मंगल, 9 अप्रैल 2019, 23:19 को Roland Huß <notifications@github.com> ने लिखा:

Issue #1108, opened 2018-10-06
color works until halfway through 'start'


borisivan:
### Description
Maven colorized output works, until it stops working partway through the 'start' of docker-maven-plugin. But when docker-maven-plugin is finished, it remains broke through the rest of the build.

### Info
All maven output colorized until the point shown below. But after that point, everything is NON colorized -- including the entire maven build that has nothing to do with docker-maven-plugin.

```
10/06/18 09:44:26 [DEBUG] Configuring mojo 'io.fabric8:docker-maven-plugin:0.27.2:start' with basic configurator -->
10/06/18 09:44:26 [DEBUG]   (f) autoCreateCustomNetworks = false
10/06/18 09:44:26 [DEBUG]   (f) execution = io.fabric8:docker-maven-plugin:0.27.2:start {execution: prepare-flyway-dbserver}
10/06/18 09:44:26 [DEBUG]   (s) name = obfuscated/postgres/postgres:10.3
10/06/18 09:44:26 [DEBUG]   (f) alias = flyway-dbserver
10/06/18 09:44:26 [DEBUG]   (f) ports = [flyway-dbserver.port:5432]
10/06/18 09:44:26 [DEBUG]   (f) bind = [/var/lib/jenkins/workspace/obfuscated]
10/06/18 09:44:26 [DEBUG]   (f) volumes = io.fabric8.maven.docker.config.RunVolumeConfiguration@1c4493c6
10/06/18 09:44:26 [DEBUG]   (f) log = (?s)database system is ready to accept connections.*database system is ready to accept connections
10/06/18 09:44:26 [DEBUG]   (f) time = 20000
10/06/18 09:44:26 [DEBUG]   (f) postStart = /opt/build/flyway_db_setup.sh managementdb
10/06/18 09:44:26 [DEBUG]   (f) exec = io.fabric8.maven.docker.config.WaitConfiguration$ExecConfiguration@41f5e6ee
10/06/18 09:44:26 [DEBUG]   (f) wait = io.fabric8.maven.docker.config.WaitConfiguration@30c9ab3f
10/06/18 09:44:26 [DEBUG]   (f) run = io.fabric8.maven.docker.config.RunImageConfiguration@5f5ab1ad
10/06/18 09:44:26 [DEBUG]   (f) images = [ImageConfiguration {name='obfuscated/postgres/postgres:10.3', alias='flyway-dbserver'}]
[...]

PR #1199, opened 2019-04-10
Issue #1108 Restore color log output after plugin executes.

wrosenuance:
Fix for #1108.

If the plugin is set to disable color output, or detects pre-3.5.0 Maven
on Windows, ANSI color logging is disabled. Previously this would then
disable color logging for all subsequent plugins, but now the setting is
restored. Additionally it has been enabled for Windows with Maven 3.5.0+.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -37,11 +37,13 @@ import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.settings.Settings;
+import org.apache.maven.shared.utils.logging.MessageUtils;
 import org.codehaus.plexus.PlexusConstants;
 import org.codehaus.plexus.PlexusContainer;
 import org.codehaus.plexus.context.Context;
 import org.codehaus.plexus.context.ContextException;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Contextualizable;
+import org.fusesource.jansi.Ansi;
 
 /**
  * Base class for this plugin.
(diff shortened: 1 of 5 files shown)


wrosenuance:
I tested this under a few settings - all on Windows, though:
- Windows cmd.exe with Maven 3.3.3
- Windows cmd.exe with Maven 3.5.4
- Windows Git bash with Maven 3.3.3
- Windows Git bash with Maven 3.5.4

Under Git bash, the System.console() method returned null, but Maven would still colorize output and it looked fine, so I switched that part of the test for a test against the Maven MessageUtils.isColorEnabled() that is set by the CLI and takes into account whether batch mode is enabled (disables color), as well as [newer features](https://github.com/apache/maven/pull/114/commits).

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -224,8 +219,8 @@ private void print(String txt) {
 
     private String colored(String message, Ansi.Color color, boolean addPrefix, Object ... params) {
         Ansi ansi = ansi().fg(color);
-        String msgToPrint = addPrefix ? prefix + message : message;
-        return ansi.a(format(evaluateEmphasis(msgToPrint, color), params)).reset().toString();
+        message = ansi.a(format(evaluateEmphasis(message, color), params)).reset().toString();
+        return addPrefix ? prefix + message : message;
```
rhuss: Why don't you want the prefix to be colored ?

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -205,34 +207,43 @@
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
         if (!skip) {
-            log = new AnsiLogger(getLog(), useColor, verbose, !settings.getInteractiveMode(), getLogPrefix());
-            authConfigFactory.setLog(log);
-            imageConfigResolver.setLog(log);
+            boolean ansiRestore = Ansi.isEnabled();
+            log = new AnsiLogger(getLog(),
+                    useColor && MessageUtils.isColorEnabled()
```
rhuss: Could we please move this into an extra method, like 'useColorForLogging()' or so ....

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
wrosenuance: This was just to match the test case that was there. It asserted that the prefix would not be highlighted even when the rest of the message was, so when it was failing, I changed the code to make the test pass. But I was also not sure that made the most sense. I guess having the actual output highlighted vs the prefix might help draw attention to it, but I'm happy to change this.

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
wrosenuance: Sure!

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
rhuss: I'd prefer to keep it like it was (as there was no issue with that yet). So would be awesome if you could fix the test instead (wondering though why it not failed before ...)

wrosenuance: I restored this.

I have a suspicion that the test wasn't failing because the ANSI logging was being disabled during forked tests since System.console() would return null. Because the tests would use the static Ansi.ansi() method to retrieve a helper to generate the expected output, it would also magically stop putting in the escape sequences, so the expected string would match.

Part of the fix for this was to instead construct a new Ansi() instance to use to construct the expected result so that it would always insert the codes.

While testing this I discovered some other bugs - if the emphasis functionality was used and the emphasis ended at the end of the message, it would fail to reset for the next line. I fixed this.

I also spent a bit of time getting the maven test suite to output the ANSI codes during testing. Because there is static usage going on, the test run individually behaved differently to when it was in the suite. Plus, JAnsi would often disable itself because the output was not a console, so in the end I have forced these codes to output just during the tests so that it is always possible to see they are working and that the console doesn't get messed up.

Issue #1203, opened 2019-04-11
NPE in docker:save when supplying file name without directory


wrosenuance:
### Description
If the docker.save.file system property is used to override the output location for docker:save and the property value is just a file name without a directory, the plugin throws an NPE.

### Info
* d-m-p version : 0.29-SNAPSHOT
* Maven version (`mvn -v`) : 3.5.4

The issue is in ensureSaveDir(), which does not check for a null result from getParentFile(), which happens when the File does not specify a directory.

```
    private void ensureSaveDir(String fileName) throws MojoExecutionException {
        File saveDir = new File(fileName).getParentFile();
        if (!saveDir.exists()) {
            if (!saveDir.mkdirs()) {
                throw new MojoExecutionException("Can not create directory " + saveDir + " for storing save file");
            }
        }
    }
```

I will submit a PR to fix, it's easy.

Issue #1200, opened 2019-04-10
NullPointerException after upgrading to 0.29.0


papegaaij:
### Description
After upgrading from 0.28.0 to 0.29.0 we get the folloing NullPointerException when trying to build a docker image:
```
Caused by: java.lang.NullPointerException
    at io.fabric8.maven.docker.config.BuildImageConfiguration.getAbsoluteContextDirPath (BuildImageConfiguration.java:329)
    at io.fabric8.maven.docker.assembly.DockerAssemblyManager$1.customize (DockerAssemblyManager.java:133)
    at io.fabric8.maven.docker.assembly.DockerAssemblyManager.createBuildTarBall (DockerAssemblyManager.java:312)
    at io.fabric8.maven.docker.assembly.DockerAssemblyManager.createDockerTarArchive (DockerAssemblyManager.java:179)
    at io.fabric8.maven.docker.service.ArchiveService.createArchive (ArchiveService.java:123)
    at io.fabric8.maven.docker.service.ArchiveService.createArchive (ArchiveService.java:118)
    at io.fabric8.maven.docker.service.BuildService.buildImage (BuildService.java:117)
    at io.fabric8.maven.docker.service.BuildService.buildImage (BuildService.java:65)
    at io.fabric8.maven.docker.BuildMojo.buildAndTag (BuildMojo.java:73)
    at io.fabric8.maven.docker.BuildMojo.processImageConfig (BuildMojo.java:104)
    at io.fabric8.maven.docker.BuildMojo.executeInternal (BuildMojo.java:60)
    at io.fabric8.maven.docker.AbstractDockerMojo.execute (AbstractDockerMojo.java:225)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
[...]

PR #1204, opened 2019-04-11
Issue #1203 NPE if docker:save called with filename only.

wrosenuance:
Fixes NPE that happens if you run mvn docker:save -Dfile=no-directory-name.tar

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/SaveMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/SaveMojo.java
@@ -106,7 +106,7 @@ public class SaveMojo extends AbstractDockerMojo {
     }
 
     private void ensureSaveDir(String fileName) throws MojoExecutionException {
-        File saveDir = new File(fileName).getParentFile();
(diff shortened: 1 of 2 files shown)


src/main/java/io/fabric8/maven/docker/SaveMojo.java
```diff
@@ -106,7 +106,7 @@ private String completeCalculatedFileName(String file) throws MojoExecutionExcep
     }
 
     private void ensureSaveDir(String fileName) throws MojoExecutionException {
-        File saveDir = new File(fileName).getParentFile();
+        File saveDir = new File(fileName).getAbsoluteFile().getParentFile();
```
rhuss: thanks ! @rohanKanojia : FYI, this fix very likely fixes also #1200 if applied to `getContextDir()`

rohanKanojia: okay, thanks

PR #1209, opened 2019-04-13
Fix #1201: Initial draft for integration tests

### LINKED ISSUE #1201: Add Support for Integration tests
I think we need tests that should run against a real docker daemon(since that's mostly what this plugin is used for) with various configuration cases. So that we can avoid breaking plugin's standard behavior again and again.

These tests don't need to be a part of this PR test suite(not sure how many cases can we check there on CI), but they can be something we can run before/after merging big PRs or doing releases. Also, it can catch the mistakes of new people like me who can potentially break plugin.

rohanKanojia:
Fix #1201 

I didn't have to do much as I just picked some of the working samples and added some executions into some of the poms. We can remove projects you feel are redundant and add some other poms in order to test important workflows. I've mostly added executions of `build`, `start` and `stop` goals, but we can add others too.


rhuss:
Cool, thanks ! Totally overlooked that PR. Awesome work, @rohanKanojia.

Unfortunately the test fails, if you have some spare, could give it a look again ?

rohanKanojia:
seems like maybe samples' dmp version is outdated. Let me try to update it tomorrow or maybe monday.

PR #1210, opened 2019-04-18
Add the ability to attach a saved image archive as a project artifact

wrosenuance:
This PR is to add the ability to attach an image archive produced by docker:save to the project as an artifact. Currently this is not possible and archives have to be attached using something like build-helper-maven-plugin.

To avoid conflict with docker:source, the default attach classifier is `archive` (or `archive-<saveAlias>`), instead of `docker` (or `docker-<alias>`) as used by that mojo.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/SaveMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/SaveMojo.java
@@ -39,9 +39,6 @@ public class SaveMojo extends AbstractDockerMojo {
 	@Parameter(property = "docker.skip.save", defaultValue = "false")
 	private boolean skipSave;
 
-	@Parameter(property = "docker.save.attach", defaultValue = "false")
-	private boolean saveAttach;
-
 	@Parameter(property = "docker.save.classifier")
 	private String saveClassifier;
 
@@ -53,7 +50,8 @@ public class SaveMojo extends AbstractDockerMojo {
 			return;
 		}
 
-		String imageName = getImageName(images);
+		ImageConfiguration image = getImageToSave(images);
+		String imageName = image.getName();
 		String fileName = getFileName(imageName);
 		ensureSaveDir(fileName);
 		log.info("Saving image %s to %s", imageName, fileName);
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/SaveMojo.java
```diff
@@ -39,6 +39,12 @@
 	@Parameter(property = "docker.skip.save", defaultValue = "false")
 	private boolean skipSave;
 
+	@Parameter(property = "docker.save.attach", defaultValue = "false")
+	private boolean saveAttach;
+
+	@Parameter(property = "docker.save.classifier")
+	private String saveClassifier;
```
rhuss: I wonder whether we could combine both properties (as I'm always trying to keep the configuration surface small). What's about having a single 'saveAttach' string property, which can be 'true/false' (then a default classifier is select for 'true') or any other value in which case it enables the featue and uses that classifier ?

wrosenuance: We can, although I always worry about making something that looks boolean, but isn't, because of [previous mockery](https://thedailywtf.com/articles/What_Is_Truth_0x3f_).

I could make specifying a non-empty `saveClassifier` mean "attach with this classifier", and not provide the default logic, since by the time you write `<saveAttach>true</saveAttach>`, you probably could have written `<saveClassifier>archive</saveClassifier>`, and we could use %a for the alias (as in `<saveClassifier>archive-%a</saveClassifier>` or -Ddocker.save.classifier=archive-%a)?

wrosenuance:
I added a commit to remove saveAttach in favour of just saveClassifier. I didn't squash this with the previous commit yet as I think if you merge PR #1208 I will need to resolve a conflict in the changelog, so I can squash it all then.

src/main/java/io/fabric8/maven/docker/SaveMojo.java
rhuss: > We can, although I always worry about making something that looks boolean, but isn't, because of [previous mockery](https://thedailywtf.com/articles/What_Is_Truth_0x3f_).
> 
> I could make specifying a non-empty `saveClassifier` mean "attach with this classifier", and not provide the default logic, since by the time you write `<saveAttach>true</saveAttach>`, you probably could have written `<saveClassifier>archive</saveClassifier>`, and we could use %a for the alias (as in `<saveClassifier>archive-%a</saveClassifier>` or -Ddocker.save.classifier=archive-%a)?

Fair enough. I think having no default name is fine as adding one is trivial (also saves the user a documentation lookup for finding out what the default would be).

Issue #1207, opened 2019-04-12
Support detecting image name during docker:build from docker archive


wrosenuance:
### Description
When an image archive is saved with the docker:save mojo or using docker save on the command line, the archive contains a manifest that records the tags that were saved. Later, when loading the image using docker:build and dockerArchive, these tags are restored. However, if these tags do not match the <name> configured in the image, then later actions by the plugin fail as they attempt to refer to a non-existent tag.

### Info
In a workflow where containers are built, saved to an archive with docker:save, and published to a Maven artifactory, then retrieved later using maven-dependency-plugin and reloaded into the Docker server, it may be difficult to know exactly how the image was tagged when the docker:save happened.

For instance, images tagged with a snapshot timestamp, or built with a Git hash in the tag that is not part of the version, can still be attached and later retrieved by a project that doesn't know (and can't know) the actual build timestamp or hash. Images saved from outside of the plugin may have had a repository in the tag that could vary, or be absent sometimes and present other times.

In these cases, it would be useful to be able to scan the archive to find the tag that matches a pattern (e.g. with the tag containing a wildcard, or an optional prefix).

* d-m-p version : 0.29-SNAPSHOT
* Maven version (`mvn -v`) : 3.5.4

rhuss:
Sound like a legit use case. Happy to help to integrate the PR you sent together with you.

PR #1215, opened 2019-04-20
Support for pattern matching in docker:stop and docker:remove

wrosenuance:
This is a replacement for PR #900 that builds on the pattern matching implemented for the docker:build loadNamePattern, described in PR #1207 .

### Pattern Lists
To try to keep the number of new configuration options to a minimum, the PR allows a single pattern string to be composed of several patterns by separating them with commas. Given commas are not allowed in
Docker names, no special quoting is supported to avoid splitting at a particular comma.

Since patterns could be applied to different parts of the Docker object model -- e.g. filter containers by container name or the running image name -- the list of patterns allows each pattern to be limited to a specific field, or left unnamed, in which case it applies everywhere. This means that by default, a docker:stop pattern will apply to both the container name and the image name, and if either matches, the container is stopped. To have a pattern only apply to the container name, prefix with name=, and for image name, with image=.

### Global and per-image configuration
The new parameters are stopNamePattern (used by docker:stop) and removeNamePattern (used by docker:remove) that can be set globally or per-image, with per-image completely replacing any global setting. Given the pattern list targetting above it is possible that these separate stop and remove parameters could be merged into a single cleanupNamePattern, or just namePattern, but the latter then may be confusing next to other parameters like containerNamePattern (which is used for generating) and loadNamePattern.

### Removing stopped containers
Like PR #900, this PR changes the behaviour of docker:stop to consider already stopped containers, because the docker:stop goal is the only goal that can remove a container, and should support removing stopped containers as well as running ones. Removing of stopped containers is only enabled if keepContainers is false.

### Removing by alias regex
This is not supported because I think the removal happens by container name or image name, and a change in alias is not relevant. The change in alias might affect the behaviour of an image configuration filter (e.g. -Dfilter=), but that's not being wildcarded.

### Documentation
Still absent 😅 - so this is still WIP - but I think the functionality and test cases are basically sound and in a state to review.


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 jobs:
   # Build Website
   doc:
```
wrosenuance: Eep - sorry, this isn't supposed to change. I'll drop it when I rebase.

pom.xml
```diff
@@ -629,5 +629,27 @@
         <defaultGoal>generate-resources asciidoctor:process-asciidoc</defaultGoal>
       </build>
     </profile>
+    <profile>
```
wrosenuance: This doesn't have to come in either, I was just using it to tidy.

src/main/asciidoc/inc/build/_overview.adoc
```diff
@@ -13,7 +13,7 @@ Alternatively an external Dockerfile template or Docker archive can be used. Thi
 
 * *contextDir* specifies docker build context if an external dockerfile is located outside of Docker build context. If not specified, Dockerfile's parent directory is used as build context.
 * *dockerFile* specifies a specific Dockerfile path. The Docker build context directory is set to `contextDir` if given. If not the directory by default is the directory in which the Dockerfile is stored.
-* *dockerArchive* specifies a previously saved image archive to load directly. Such a tar archive can be created with `docker save`. If a `dockerArchive` is provided, no `dockerFile` or `dockerFileDir` must be given.
```
wrosenuance: this is from PR #1207

src/main/asciidoc/inc/build/_configuration.adoc
```diff
@@ -77,6 +77,18 @@ A provided `<from>` takes precedence over the name given here. This tag is usefu
 | *imagePullPolicy*
 | Specific pull policy for the base image. This overwrites any global pull policy. See the globale configuration option <<image-pull-policy, imagePullPolicy>> for the possible values and the default.
 
+| *loadNamePattern*
```
wrosenuance: this is from PR #1207

src/main/asciidoc/inc/_docker-save.adoc
```diff
@@ -10,6 +10,31 @@ If the option `saveFile` is not set, the file name is calculated automatically:
 
 Please note that the exported image contains all image layers and can be quite large (also, it takes a bit to export the image).
 
+.Controlling image compression
```
wrosenuance: this is from PR #1207

src/main/asciidoc/inc/_docker-save.adoc
```diff
@@ -27,6 +52,10 @@ Please note that the exported image contains all image layers and can be quite l
 | The filename to save.
 | `docker.save.file` or `docker.file` or `file`
 
+| *saveClassifier*
```
wrosenuance: this is from PR #1207

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -175,6 +175,9 @@ when a `docker.from` or a `docker.fromExt` is set.
 | *docker.labels.LABEL*
 | Sets a label which works similarly like setting environment variables.
 
+| *docker.loadNamePattern*
```
wrosenuance: this is from PR #1207

src/main/java/io/fabric8/maven/docker/SaveMojo.java
```diff
@@ -39,6 +39,9 @@
 	@Parameter(property = "docker.skip.save", defaultValue = "false")
 	private boolean skipSave;
 
+	@Parameter(property = "docker.save.classifier")
```
wrosenuance: this is from PR #1207

wrosenuance:
I realise now this is based on a bunch of unmerged changes from other PRs, so the best thing to review is just the [final commit](https://github.com/fabric8io/docker-maven-plugin/pull/1215/commits/9cbc11f93ef2d50de1a72016fc0948c6524a83bb).

rhuss:
Thanks for the PR ! Let's postpone for the next release, as I want to cut a release today. 

Could you please fix the conflicts ? I will do a review then afterwards.

wrosenuance:
I rebased onto the updated master - it's much easier to follow now.

Somehow I seem to have changed the CI comments and the ones from the main project don't run anymore, just the ones I tried to configure for branches in my fork. I hope they are still running somewhere for you.

wrosenuance:
Hi @rhuss, I have been tied up for a couple of weeks, but I would like to get this PR ready to merge.

I think the main thing that is lacking is documentation. Did you have any other feedback on the proposal?

rhuss:
@wrosenuance sorry, same for me. I only can work in bursts on this plugin. I will have a look when flying back home. Sorry for the delay.

wrosenuance:
Hi @rhuss, I saw your comment on the other PR about being away for a few weeks and I thought I should try to get this in before you head out!

I wrote up the documentation, and revised some of the behavior to hopefully make more sense.

The main change is that where the stopNamePattern and removeNamePattern on the mojos was previously used as a fallback if the image did not define one, they are now used to perform a pattern-based stop or remove independently of any defined images.

This is because I think there are two likely use cases for pattern-based stop/remove:
 * remove everything - uses the mojo-level configuration in an execution that defines no images
 * remove instances related to a particular image - uses the image-level configuration

In the scenario where you want to stop/remove using per-image patterns, I can't see it is much help to have a mojo-level fallback.

rhuss:
thanks for all your patience (and apologies for being late again). I try to squeeze that PR in today, so let's get that done. I would then make also a new release before going to vacations ;-)

Issue #1237, opened 2019-07-02
Bug in fix #1148


rpecheur:
When data are extracted from local ~/.docker/config.json, "docker.buildArg" is not removed from property name.

It give:
"Build args set {docker.buildArg.http_proxy=http://127.0.0.1:3001}"
Instead of:
"Build args set {http_proxy=http://127.0.0.1:3001}"

In commit ad5b3222fafc598456de74dc42ed4b3b24e97709:
BuildService.java: Line 217:
buildArgs.put(argPrefix + entry.getValue(), defaultProxyObj.get(entry.getKey()).getAsString());
-> Should remove argPrefix.

Thank you

rohanKanojia:
You mean #1148 ?

rpecheur:
Yes

rohanKanojia:
Can you raise a PR to fix it then? If you have spotted the bug?

rpecheur:
It s detailed in my post.
One element to remove i think.

rohanKanojia:
Yeah, I get it. But I'm asking you to raise PR if you have time :-)

rpecheur:
Done #1238

rhuss:
#1238 is merged. Thanks for your patience and sorry for the delay.

PR #1238, opened 2019-07-02
Fix Issue [#1237] - Bug in properties names extracted from docker con…

rpecheur:
Bug in properties names extracted from docker config json file


rohanKanojia:
@rpecheur : Could you please add a line to changelog also?

rpecheur:
Done

PR #1244, opened 2019-07-13
Relax targetDir requirements

tobiasstadler:
It should be possible that targetDir starts with an environment variable (e.g. to copy a war to the deployment dir of an app server)


rhuss:
> It should be possible that targetDir starts with an environment variable (e.g. to copy a war to the deployment dir of an app server)

Could you please elaborate on your use case ? IIUR this PR should allow for the target directoruy for COPY operation an environment variable so that you get entries in your Dockerfile like

```
COPY target/myapp.war ${TOMCAT_HOME}/webapps/
```

Who would set this environment variable `${TOMCAT_HOME}` which is evaluated during Docker build time, not when the container is running ?

tobiasstadler:
In my case I do have a per built docker image with `ENV WILDFLY_DEPLOYMENT_DIR=/srv/wildfly/standalone/deployments/` and I want to copy the war built by maven to that dir.

Right now I am using `/${WILDFLY_DEPLOYMENT_DIR}` as the target dir, but I would be nicer if I could remove the / at the beginning.

rhuss:
@tobiasstadler sorry, I still did not fully get it ;-) Would you mind to share you concrete plugin configuration ? Thanks !

tobiasstadler:
Here is a simplified version of what I want to achieve:
```xml
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
    <groupId>com.github.tobiasstadler</groupId>
    <artifactId>docker-maven</artifactId>
    <version>1.0-SNAPSHOT</version>
    <packaging>war</packaging>
    <properties>
        <failOnMissingWebXml>false</failOnMissingWebXml>
    </properties>
    <build>
        <finalName>${project.artifactId}</finalName>
        <plugins>
            <plugin>
                <groupId>io.fabric8</groupId>
                <version>0.30.0</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>build</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <images>
                        <image>
                            <name>tobiasstadler/docker-maven</name>
                            <build>
                                <from>airhacks/wildfly</from>
                                <assembly>
                                    <descriptorRef>artifact</descriptorRef>
                                    <targetDir>${DEPLOYMENT_DIR}</targetDir>
                                </assembly>
                            </build>
                        </image>
                    </images>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```

DEPLOYMENT_DIR is "exposed" by the airhacks/wildfly image.

rhuss:
I see. Using a variable like this in the config might be a bit dangerous because Maven would replace this when this is set in a property. Only when it's unset it's copied over literally.

PR #1245, opened 2019-07-13
Increase interoperability with docker-java

tobiasstadler:
docker-java users registry.* instead docker.* for registry auth config. Provide a fall back to registry.* when docker.* auth config is not found in order to increase interoperability to docker-java


rhuss:
Thanks for the PR ! Could you please help me to understand the use case for this request ? As d-m-p doesn't use docker-java I can't see where docker-java comes into play. Are you using another plugin which uses docker-java internally (which one ?) and you don't want to duplicate the properties ?

tobiasstadler:
Yes, I am also  using Arquillian and Testcontainers in my project (both use docker-java). And it would be nice to set the properties only once.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -207,10 +207,12 @@ private AuthConfig createStandardAuthConfig(boolean isPush, Map authConfigMap, S
             }
 
             // Check for openshift authentication either from the plugin config or from system props
-            ret = getAuthConfigFromOpenShiftConfig(lookupMode, authConfigMap);
-            if (ret != null) {
-                log.debug("AuthConfig: OpenShift credentials");
-                return ret;
+            if (lookupMode != LookupMode.REGISTRY) {
```
rhuss: I would skip that as it doesn't harm to use the openshift lookup with a registry. prefix, too. Its probably more confusing to have this special case here.

src/test/java/io/fabric8/maven/docker/util/AuthConfigFactoryTest.java
```diff
@@ -112,6 +112,35 @@ public void testSystemProperty() throws Exception {
...
+            verifyAuthConfig(config,"roland","secret","roland@jolokia.org");
+        } finally {
+            System.clearProperty("docker.username");
+            System.clearProperty("docker.password");
+            System.clearProperty("docker.email");
+        }
+    }
+
+    @Test
+    public void testRegistrySystemProperty() throws Exception {
```
rhuss: can we please avoid copy & pasting the test ? Just extract a common method which is called once with "docker" and once with "registry" from each individual test.

Helpful would be also a test that ensures that "docker" has precedence over "registry" to keep the default behaviour.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
tobiasstadler: The reason why I excluded `LookupMode.REGISTRY` is:
When the config contains `useOpenShiftAuth=true` and the system properties contain `docker.useOpenShiftAuth=false` `createStandardAuthConfig` returns a (valid) configuration (see `io.fabric8.maven.docker.util.AuthConfigFactoryTest#testOpenShiftConfigFromSystemPropsNegative` which fails if I remove the if)

PR #1251, opened 2019-08-08
Add option to keep trying to remove containers, even if one fails

peterjeschke:
This PR adds an option to continue removing container and networks if one fails. A possible scenario is that the stop goal might try to remove multiple containers and one does not exist anymore (for whatever reason). This change will allow (with a new configuration option) to continue removing containers. The build will still fail though, to make the user aware of the problem.


peterjeschke:
@rhuss I removed the configuration option and updated the branch to the latest commit on master

PR #1263, opened 2019-09-06
Prefer HOME environment variable

twz123:
The Java system property is filled by a syscall to determine the user's home directory. This is not matching the typical implementation of the docker and kubectl binaries, which prefer the HOME OS environment variable over the syscall.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,6 @@
 # ChangeLog
 
-* 0.34-SNAPSHOT
+* **0.34-SNAPSHOT**
   - Building 'spring-boot-with-jib' sample fails with NoSuchMethodError ([1384](https://github.com/fabric8io/docker-maven-plugin/issues/1384))
   - Loading Image tarball into docker daemon fails in JIB mode ([1385](https://github.com/fabric8io/docker-maven-plugin/issues/1385))
   - `assembly.inline` is removed when external properties are enabled ([1082](https://github.com/fabric8io/docker-maven-plugin/issues/1082))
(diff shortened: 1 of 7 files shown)


twz123:
I discovered this during the execution of a CI build, where the HOME variable was changed inside the build and the login to the Docker registry was made by `docker login`. The dmp wasn't picking up the login credentials in that case. My workaround was to also define `JAVA_TOOL_OPTIONS="-Duser.home=...`. Took me quite a while to figure out the problem. That's why I thought it might be a good idea to actually file a PR here.

twz123:
Updated the changelog.

rhuss:
I think this PR is valid, as a user can change HOME but not user.home.

I wonder whether we need update/add some documentation, too ?

rhuss:
`user.home` is not only used in`DockerFileUtil` but also in e.g `VolumeBindUtil` and also in `DockerConnectionDetector`. While I'm positive to this change I think we should also update all other places where user.home is used as the location of the home directory.

@twz123 anything you would like to tackle, too ? (sorry for the long delay, the intervals when i'm able to work on dmp get longer and longer ...)

twz123:
,@rhuss sure, I'll have a look, although this might take some time on my side, too 🙈

rhuss:
> ,@rhuss sure, I'll have a look, although this might take some time on my side, too 🙈

thanks ! and no worries, we are operating on a slow time scale anyway (sorry, i feel guilty a bit here that everything is so slow here, but i'm working on a solution here)

rohanKanojia:
@twz123 : polite ping, Are you still working on this?

rhuss:
@rohanKanojia let's fix that on our own, fix shouldn't be hard. @twz123 hope this is ok for you when we take over the last steps.

PR #1279, opened 2019-10-14
Fix Boolean comparison for healthy option

fabienbancharel:
It seems there is a regression on healthy option since 0.25.0, and the change from "boolean" to "Boolean".

It seems that "wait.getHealthy() == Boolean.TRUE" is always false, introduced in 0.25.0.
Comparison should be by value, not by reference.

Maven version : 3.5.4
JDK : 8

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -6,6 +6,7 @@
   - Allow merging of image configurations using <imagesMap> ([#360](https://github.com/fabric8io/docker-maven-plugin/issues/360))
   - Treat bridged and default network mode the same #1234
   - Fix NPE when cacheFrom is missing from config #1274
+  - Fix healthy option regression introduced in 0.25.0 #1279
 
 * **0.31.0** (2019-08-10)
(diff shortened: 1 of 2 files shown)


rohanKanojia:
@fabienbancharel : Thanks a lot for PR :+1: Could you please add a line to CHANGELOG regarding this change? We're good to merge this.

fabienbancharel:
@rohanKanojia : Do you mean adding a line to "doc/changelog.md", and doing an other PR ?
In a new "paragraph" "0.32-SNAPSHOT" ?
Thanks.

rohanKanojia:
Yeah, no need to create new or. You can just do a `git commit --amend` and
force push

सोम, 14 अक्तू॰ 2019, 18:59 को fabienbancharel <notifications@github.com> ने
लिखा:

fabienbancharel:
I followed this, before you answered :
"Add more commits by pushing to the master branch on xxx/docker-maven-plugin."

Do you want me to change something or is it ok ?

rohanKanojia:
:+1:

PR #1286, opened 2019-11-07
Use top-level authentication for extended auth

sebastiankirsch:
When relying on [Extended Authentication](https://dmp.fabric8.io/#extended-authentication) for ECR, we unfortunately have to configure the `push` and `pull` sections, even though they are the same. The problem seems to be that the AWS session token has to be put into the `<auth>` element, whereas the general section only knows an `<authToken>` element. This change would allow Extended Authentication to work with the general configuration.

Maybe the proper way forward would be to deprecate the `authToken` property and introduce `auth` - or vice versa.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/RegistryAuthConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/RegistryAuthConfiguration.java
@@ -45,6 +45,7 @@ public class RegistryAuthConfiguration implements Serializable {
         }
         if (StringUtils.isNotBlank(authToken)) {
             authMap.put(AuthConfigFactory.AUTH_AUTHTOKEN, authToken);
+            authMap.put("auth", authToken);
         }
         if (StringUtils.isNotBlank(email)) {
             authMap.put(AuthConfigFactory.AUTH_EMAIL,email);


rhuss:
Sorry, I do not fully understand the issue. Do you say that pull and push uses a different configuration format than what ?

Could you please elaborate a bit when `auth` and when `authToken` should be used ? Is there canonical documentation on the AWS side about the extended authentication ? (sorry, not deep in that code as this was a contribution, too, and I don't have much experience with AWS)

sebastiankirsch:
@rhuss We have to configure dmp like this:

    <plugin>
        <groupId>io.fabric8</groupId>
        <configuration>
            <authConfig>
                <pull>
                    <auth>${AWS_SESSION_TOKEN}</auth>
                    <username>${AWS_ACCESS_KEY_ID}</username>
                    <password>${AWS_SECRET_ACCESS_KEY}</password>
                </pull>
                <push>
                    <auth>${AWS_SESSION_TOKEN}</auth>
                    <username>${AWS_ACCESS_KEY_ID}</username>
                    <password>${AWS_SECRET_ACCESS_KEY}</password>
                </push>
            </authConfig>
        </configuration>
    </plugin>

instead of this

    <plugin>
        <groupId>io.fabric8</groupId>
        <configuration>
            <authConfig>
                <authToken>${AWS_SESSION_TOKEN}</authToken>
                <username>${AWS_ACCESS_KEY_ID}</username>
                <password>${AWS_SECRET_ACCESS_KEY}</password>
            </authConfig>
        </configuration>
    </plugin>

Same goes for configuration via System properties. Now this is awkward at least for the AWS-related stuff; but my understanding is that the `<auth>` and `<authToken>` elements are used for other authentication mechanisms as well, and are likely to have the same issue.

sebastiankirsch:
"Extended Authentication" (not sure where the term comes from) means to [fetch an authorization token](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetAuthorizationToken.html) to access AWS ECR, AWS's docker repositories.
In order to do that, one has to authenticate the necessary API call with AWS. If used in conjunction with [temporary credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html) - which is the recommended way to authenticate with AWS for security, you have to provide not only username & password, but also a "session token", hence the third criteria.

rhuss:
Ok, I see. I think we can fix this more generally by deprecating the top-level "authToken" in favor of an "auth" there, too. I created a PR #1296 would be awesome if you could verify whether it fits your bill.

sebastiankirsch:
#1296 looks good to me, just added some minor comments.
It certainly supersedes this PR!

rhuss:
Perfect, so lets get #1296 merged and I'm closing this one.

Thanks!

sebastiankirsch:
Thank you!

PR #1235, opened 2019-07-01
Add capability to get ECR credentials from IAM Task role from contain…

yadavnikhil:
Add capability to get ECR credentials from IAM Task role from container task metadata endpoint for ECS running with Fargate deployment type.
    For ECS with EC2 instance with will query instance metadata service if ECS_AWSVPC_BLOCK_IMDS is "false". It will query task metadata endpoint in instance metadata service is not accessible (blocked).

This fixes #1233

PR #1295, opened 2019-11-20
Support both ECS & Fargate tasks

sebastiankirsch:
The feature to pick up [AWS credentials for Fargate tasks](https://docs.aws.amazon.com/AmazonECS/latest/userguide/task-metadata-endpoint-fargate.html) was introduced with #1235 
The same mechanism applies to [ECS-based tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v2.html), but the composition of the URI is either slightly different or the underlying implementation more sensitive to the correctness of the URL: the current implementation of the plugin calculates `http://169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`, whereas `http://169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` would be appropriate for ECS-based tasks.

This PR extends the feature to work for both scenarios, and adds unit tests for both of them.
Instead of adding another framework like e.g. [WireMock](http://wiremock.org/), I opted to use the HTTP server shipped with [Apache HttpCore](https://hc.apache.org/httpcomponents-core-ga/), which is already in the dependency list; I hope that's ok.

I would kindly ask @yadavnikhil to test this feature before integrating/releasing it; since I have no real-world scenario with which I could validate this PR on Fargate.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
+++ b/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
@@ -10,7 +10,6 @@ import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
--- a/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
+++ b/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
@@ -248,7 +248,7 @@ public class AuthConfigFactory {
             }
 
             try {
-                ret = getAuthConfigFromTaskRole(settings);
+                ret = getAuthConfigFromTaskRole();
             } catch (ConnectTimeoutException ex) {
                 log.debug("Connection timeout while retrieving ECS meta-data, likely not an ECS instance (%s)",
                         ex.getMessage());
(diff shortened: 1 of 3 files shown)


src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -369,6 +386,21 @@ private AuthConfig getAuthConfigFromECSTaskRole() throws IOException {
         }
     }
 
+    /** This is present to support testing */
+    private Map<String, Object> getMetadataSettings(Settings settings) {
+        Server server = settings.getServer("junit.ecs-meta");
```
rhuss: Could we avoid to refer to tests from the code-to-test ? 

I really appreciate the effort put into testing (although its more an integration then a unit-test, but I'm not dogmatic here except if the test duration increase significantly).

I suggest that instead of going over the settings, just use an environment variable like `ECS_AUTH_SERVICE=<host>:<port>` which is just set prior to the test, and, if not set, uses the ECS default values. That way you avoid a coupling to the test (and provide kind of an easy to use general feature, which we might expose to others, but wouldn't comment it yet.)

sebastiankirsch: Sure can do that.
In that case I'd suggest to integrate [system-rules](https://github.com/stefanbirkner/system-rules) in order to manipulate the [Environment Variables](https://stefanbirkner.github.io/system-rules/#EnvironmentVariables)

sebastiankirsch: I agree that this is more of an integration test; but this code is really hard to test without extracting methods for the sake of testing that than have to be mocked/overridden (the part where the service is called).
I can do that as well; not my personal preference but it's your project :wink:

rhuss: > In that case I'd suggest to integrate [system-rules](https://github.com/stefanbirkner/system-rules) in order to manipulate the [Environment Variables](https://stefanbirkner.github.io/system-rules/#EnvironmentVariables)

Sounds good.

> I agree that this is more of an integration test; but this code is really hard to test without extracting methods for the sake of testing that than have to be mocked/overridden (the part where the service is called).

No, I'm fine with using an HTTP server here, no worries.

sebastiankirsch: Great, I'll update the PR then...

sebastiankirsch:
@rhuss fixed the test setup by considering the `ECS_METADATA_ENDPOINT` env variable.
I wouldn't mention that in the docs since from AWS' side it's absolutely fixed and thus the variable is introduced for testing only.

I additionally extracted a method b/c the response from ECS metatdata is the same as the one from EC2 metadata.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -1,17 +1,16 @@
...
 import java.io.InputStreamReader;
 import java.io.Reader;
 import com.google.gson.JsonObject;
 
-import java.io.IOException;
 import java.lang.reflect.Method;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
```
rhuss: Sonarcloud says this is unused.

src/test/java/io/fabric8/maven/docker/util/AuthConfigFactoryTest.java
```diff
@@ -532,13 +578,45 @@ private Server create(String id, String user, String password, String email) {
...
+                httpServer.getInetAddress().getHostAddress()+":" + httpServer.getLocalPort());
+        environmentVariables.set("AWS_CONTAINER_CREDENTIALS_RELATIVE_URI", containerCredentialsUri);
+    }
+
+    private void verifyAuthConfig(AuthConfig config, String username, String password, String email, String auth) {
+        assertNotNull(config);
         JsonObject params = gsonBuilder.create().fromJson(new String(Base64.decodeBase64(config.toHeaderValue().getBytes())), JsonObject.class);
-        assertEquals(username,params.get("username").getAsString());
-        assertEquals(password,params.get("password").getAsString());
+        assertEquals(username, params.get("username").getAsString());
```
rhuss: we can switch over to the AuthCofig constants if the other PR is merged.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
sebastiankirsch: Uhm, I don't see that line anymore?

sebastiankirsch: Sorry, wrong branch. Done

src/test/java/io/fabric8/maven/docker/util/AuthConfigFactoryTest.java
sebastiankirsch: Done-

rhuss:
Thanks ! Let's get that in and if there any remaining issue fix that later.

PR #1296, opened 2019-11-21
fix(auth): Deprecate "authToken" in favor of "auth"

rhuss:
This allows extended ecr authentication to share the same token for pull and push requests.

See also #1286.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/RegistryAuthConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/RegistryAuthConfiguration.java
@@ -48,11 +48,14 @@ public class RegistryAuthConfiguration implements Serializable {
             authMap.put(AuthConfig.AUTH_PASSWORD, password);
         }
 
-        // deprecated "authToken"
+        if (StringUtils.isNotBlank(authToken) && StringUtils.isNotBlank(auth)) {
+            throw new IllegalStateException("For a registry configuration either 'auth' or 'authToken' (deprecated) can be specified but not both. Use only 'auth' and remove 'authToken' in the registry configuration");
+        }
+
         if (StringUtils.isNotBlank(authToken)) {
             authMap.put(AuthConfig.AUTH_AUTH, authToken);
         }
-        // overwrite with "auth"
+
         if (StringUtils.isNotBlank(auth)) {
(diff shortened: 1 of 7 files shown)


src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -371,21 +364,30 @@ private AuthConfig getAuthConfigFromECSTaskRole() throws IOException {
...
         }
     }
 
+    private String getAuthProperty(Properties props, LookupMode lookupMode) {
+        String authProp = props.getProperty(lookupMode.asSysProperty(AuthConfig.AUTH_AUTH));
+        if (authProp != null) {
+            return authProp;
+        }
+        // Fallback is deprecated AUTH_AUTHTOKEN property
+        return props.getProperty(lookupMode.asSysProperty("authToken"));
```
sebastiankirsch: I think you kept the `AUTH_AUTHTOKEN` constant to be referenced here?

src/main/java/io/fabric8/maven/docker/config/RegistryAuthConfiguration.java
```diff
@@ -38,16 +42,23 @@ public Map toMap() {
...
+            authMap.put(AuthConfig.AUTH_PASSWORD, password);
         }
+
+        // deprecated "authToken"
         if (StringUtils.isNotBlank(authToken)) {
-            authMap.put(AuthConfigFactory.AUTH_AUTHTOKEN, authToken);
+            authMap.put(AuthConfig.AUTH_AUTH, authToken);
         }
+        // overwrite with "auth"
+        if (StringUtils.isNotBlank(auth)) {
```
sebastiankirsch: This will certainly make do, but I suggest to throw a _configuration exception_ if both auth & authToken are set.
This may seem disruptive, yet: if someone specified both, they obviously worked on that part just now and would thus be grateful to know that there's configuration they can remove.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
rhuss: Well, the idea was to remove the constant so that its not reused and the usage of the deprecated value is confined to a single space. So I remove the constant definition instead.

PR #1297, opened 2019-11-21
Fix #1197: Unable to build dockerFile without pushing it to docker server (DockerInDocker)

### LINKED ISSUE #1197: Unable to build dockerFile without pushing it to docker server (DockerInDocker)
Hi,

I use the docker-maven-plugin to generate a docker image from a Gitlab CI job. To achieve this I run maven inside a docker image. This docker image needs to access docker to build the image.

**To avoid Docker-In-Docker (security) problems, I would like to use Kaniko** as described here: https://docs.gitlab.com/ee/ci/docker/using_kaniko.html

With this approach, **I expect the docker-maven-plugin to create the DockerFile, but not to call the docker server to build the image**. In fact, a second Gitlab job dedicated to Kaniko will build the image after.

The plugin configuration does not support this feature: there is no parameter dedicated to build the DockerFile without building the image itself.
I successfully achieved it patching the class named **DockerAccessWithHcClient**, creating a new protocol named "stub" as an alternative to unix/npipe to bypass docker server calls (see the attached file : [DockerAccessWithHcClient.zip](https://github.com/fabric8io/docker-maven-plugin/files/3059674/DockerAccessWithHcClient.zip)).

It would be great if **this feature could be integrated to the plugin**. I would greatly appreciate to discuss on this here.

Thanks,
Ben.

rohanKanojia:
Fix #1197 

with `docker.skip.build.image` flag we can skip pushing created docker archive to docker daemon.
Peek 2019-11-21 20-29

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/BuildMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/BuildMojo.java
@@ -42,6 +42,12 @@ public class BuildMojo extends AbstractBuildSupportMojo {
     protected String name;
 
     /**
+     * Skip Sending created tarball to docker daemon
+     */
+    @Parameter(property = "docker.buildArchiveOnly", defaultValue = "false")
+    protected String buildArchiveOnly;
+
+    /**
      * Skip building tags
      */
     @Parameter(property = "docker.skip.tag", defaultValue = "false")
@@ -71,7 +77,10 @@ public class BuildMojo extends AbstractBuildSupportMojo {
         ImagePullManager pullManager = getImagePullManager(determinePullPolicy(imageConfig.getBuildConfiguration()), autoPull);
         BuildService buildService = hub.getBuildService();
 
-        buildService.buildImage(imageConfig, pullManager, buildContext);
+        File buildArchiveFile = buildService.buildArchive(imageConfig, buildContext, resolveBuildArchiveParameter());
+        if (Boolean.FALSE.equals(shallBuildArchiveOnly())) {
+            buildService.buildImage(imageConfig, pullManager, buildContext, buildArchiveFile);
+        }
         if (!skipTag) {
             buildService.tagImage(imageConfig.getName(), imageConfig);
         }
@@ -83,6 +92,27 @@ public class BuildMojo extends AbstractBuildSupportMojo {
     protected Date getReferenceDate() {
         return new Date();
     }
+
+    private String resolveBuildArchiveParameter() {
+        if (buildArchiveOnly != null && !buildArchiveOnly.isEmpty()) {
+            if (!(buildArchiveOnly.equalsIgnoreCase("false") ||
+                buildArchiveOnly.equalsIgnoreCase("true"))) {
+                return buildArchiveOnly;
+            }
+        }
+        return null;
+    }
+
+    private boolean shallBuildArchiveOnly() {
+        if (buildArchiveOnly != null && !buildArchiveOnly.isEmpty()) {
+            if (buildArchiveOnly.equalsIgnoreCase("false") ||
+                    buildArchiveOnly.equalsIgnoreCase("true")) {
+                return Boolean.parseBoolean(buildArchiveOnly);
+            }
+        }
+        return false;
+    }
+
     private String determinePullPolicy(BuildImageConfiguration buildConfig) {
         return buildConfig != null && buildConfig.getImagePullPolicy() != null ? buildConfig.getImagePullPolicy() : imagePullPolicy;
     }
--- a/src/main/java/io/fabric8/maven/docker/service/BuildService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildService.java
@@ -3,6 +3,7 @@ package io.fabric8.maven.docker.service;
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.file.Files;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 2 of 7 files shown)


rohanKanojia:
However, when I try to load image using this tarball. I get some error:
```
~/work/repos/dmp-demo-project : $ docker load --input ./target/docker/test-db/tmp/docker-build.tar                                                            
open /var/lib/docker/tmp/docker-import-295740741/repositories: no such file or directory  
```

Is this desired behavior?

rhuss:
> Is this desired behavior?

Yes, I think so. It's really that this tar is picked up e.g. by another build mechansim like Kaniko to build the real image. However, I think it would be important to implement `-Ddocker.saveBuildArchive=target/docker-build.tar` as well to copy over the tar to a different location so that it can be picked up from there without the need to dive into the target directory. Should be either an absolute path or, if relative, should be relative to the project directory.

rhuss:
Actually thinking again, I think that a single argument should be sufficient:

```
buildArchiveOnly (property: docker.buildArchiveOnly) for creating only the archive with Dockerfile which needs to be post-processed to create an image. The value should point to the path where to create the archive. If no value is given, the archive is created below `target/docker` as usual but no image is created.
```

wdyt ?

rohanKanojia:
@rhuss : okay, let me do the changes as you asked.

rohanKanojia:
Here is what I interpreted:

- Remove `docker.skip.build.image` 
- Add a new flag `docker.buildArchiveOnly` which would contain the path to where tarball should be saved/copied. I have one doubt here;  `If no value is given, the archive is created below target/docker as usual but no image is created`. But I thought no value would be the case when we want to POST the tarball to docker daemon. Am I missing something here?

rhuss:
I hope we can detect if the option is given but without value. I'm not 100% if this is possible. Ideally we should also interprete the special values of `true` and `false`.

So my suggestion would be:

* `-DdockerBuildArchiveOnly=/path/to/archive`  will create the build tar with name `/path/to/archive` and then stop without actually building
* `-DdockerBuildArchiveOnly`, `-DdockerBuildArchiveOnly=true` will just don't build the image (but don't copy over the archive to another place, just leaving it in target.
* `-DdockerBuildArchiveOnly=false` just build as usual. This is the default but might be useful if you want to override this field in sub-module pom.

@rohanKanojia wdyt ? I think this will nicely cover this feature with a single config option but still staying flexible and intuitive.

rohanKanojia:
cool :+1: , let me implement this in next update

rohanKanojia:
Here is how it looks now:
- Using `docker.buildArchiveOnly=/new/path`
Peek 2019-11-22 02-11

marcsaegesser:
I hadn't noticed that the discussion had moved over here. Sorry for delay responding.

I cloned this repo, built this branch and did some testing. So far things look good. I had to add ```-Ddocker.skip.tag=true``` when using ```-Ddocker.buildArchiveOnly``` or the build failed when trying to tag a non-existent image. Should ```docker.buildArchiveOnly``` imply ```docker.skip.tag```?

I've updated one of my CI/CD pipelines to use this version of the plugin and then used Kaniko to create the Docker image in an environment that had no Docker daemon.

src/main/java/io/fabric8/maven/docker/service/BuildService.java
```diff
@@ -60,14 +61,55 @@
...
+     * Create docker archive for building image
+     *
+     * @param imageConfiguration image configuration
+     * @param buildContext docker build context
+     * @param buildArchiveOnlyFlagOrBuildArchiveLocation build archive only flag, it can have values TRUE or FALSE and also
+     *                               it can hold path to archive where it might get copied over
+     * @return tarball for docker image
+     * @throws MojoExecutionException in case any exception comes during building tarball
+     */
+    public File buildArchive(ImageConfiguration imageConfiguration, BuildContext buildContext, String buildArchiveOnlyFlagOrBuildArchiveLocation)
```
rhuss: Can we make the last argument just a `archivePath` and if not-null, store the archive there ? The caller should then prepare that parameter based on the options given. That way this method becomes much more reusable and decoupled from its actual, Maven specific usage.

src/main/java/io/fabric8/maven/docker/service/BuildService.java
```diff
@@ -60,14 +61,55 @@
...
+        if (buildConfig.getDockerArchive() != null) {
+            return buildConfig.getAbsoluteDockerTarPath(params);
+        }
+        long time = System.currentTimeMillis();
+
+        File dockerArchive = archiveService.createArchive(imageName, buildConfig, params, log);
+        log.info("%s: Created %s in %s", imageConfiguration.getDescription(), dockerArchive.getName(), EnvUtil.formatDurationTill(time));
+
+        // Copy created tarball to directory if specified
+        if (buildArchiveOnlyFlagOrBuildArchiveLocation != null && !buildArchiveOnlyFlagOrBuildArchiveLocation.isEmpty()) {
```
rhuss: Can we extract that to a method ? Also, remember, when its empty its the same as false

src/main/java/io/fabric8/maven/docker/BuildMojo.java
```diff
@@ -71,7 +77,10 @@ protected void buildAndTag(ServiceHub hub, ImageConfiguration imageConfig)
         ImagePullManager pullManager = getImagePullManager(determinePullPolicy(imageConfig.getBuildConfiguration()), autoPull);
         BuildService buildService = hub.getBuildService();
 
-        buildService.buildImage(imageConfig, pullManager, buildContext);
+        File buildArchiveFile = buildService.buildArchive(imageConfig, buildContext, buildArchiveOnly);
```
rhuss: as mentioned below, the option should be already resolved here to the path (if any) for calling `buildArchive`

PR #1301, opened 2019-11-22
Fix #544: Avoiding chown to reduce the image size

### LINKED ISSUE #544: Avoiding chown to reduce the image size
The issue #175 introduced an extra layer for the chown that is doubling the content size of the chown ed folder (for an image containing the jdk+jboss it could be huge).
Is it possible to avoid this extra layer using a single CMD (without the COPY)?

By doing a curl and chown in the same RUN command, this drastically reduces the image size, but in this case I have to use an external Dockerfile and I'd prefer being fully maven-style (using the assembly).

Any idea how to fix this?

rohanKanojia:
Fix #544

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
@@ -195,32 +195,32 @@ public class DockerFileBuilder {
 
     private void addCopy(StringBuilder b) {
         if (assemblyUser != null) {
-            String tmpDir = createTempDir();
-            addCopyEntries(b, tmpDir);
+            String[] userParts = assemblyUser.split(":");
 
-            String[] userParts = StringUtils.split(assemblyUser, ":");
-            String userArg = userParts.length > 1 ? userParts[0] + ":" + userParts[1] : userParts[0];
-            String chmod = "chown -R " + userArg + " " + tmpDir + " && cp -rp " + tmpDir + "/* / && rm -rf " + tmpDir;
-            if (userParts.length > 2) {
-                DockerFileKeyword.USER.addTo(b, "root");
-                DockerFileKeyword.RUN.addTo(b, chmod);
-                DockerFileKeyword.USER.addTo(b, userParts[2]);
-            } else {
-                DockerFileKeyword.RUN.addTo(b, chmod);
+            for (CopyEntry entry : copyEntries) {
+                if (userParts.length > 2) {
+                    DockerFileKeyword.USER.addTo(b, "root");
+                }
+                addCopyEntries(b, "", (userParts.length > 1 ?
+                        userParts[0] + ":" + userParts[1] :
+                        userParts[0]));
+                if (userParts.length > 2) {
+                    DockerFileKeyword.USER.addTo(b, userParts[2]);
+                }
             }
         } else {
-            addCopyEntries(b, "");
+            addCopyEntries(b, "", null);
         }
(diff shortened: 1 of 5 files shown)


src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
```diff
@@ -195,28 +195,20 @@ private static void buildOption(StringBuilder b, DockerFileOption option, Object
...
-                DockerFileKeyword.RUN.addTo(b, chmod);
-                DockerFileKeyword.USER.addTo(b, userParts[2]);
-            } else {
-                DockerFileKeyword.RUN.addTo(b, chmod);
+            String[] userParts = assemblyUser.split(":");
+
+            for (CopyEntry entry : copyEntries) {
+                String dest = (basedir.equals("/") ? "" : basedir) + "/" + entry.destination;
+                DockerFileKeyword.ADD.addTo(b, " --chown="
+                        + (userParts.length > 2 ?
```
rhuss: Really > 2 ? Not > 1 ?

rhuss: I'm asking because I don't see `userPart[2]` used anywhere.

rhuss: I think, for backwards compatibility we shoudl use `userParts[2]` to switch to that user after the `ADD`, like in the original code. (but we can mark that as deprecated, as its not really useful anymore IMO)

rohanKanojia: Sorry, I might have missed this. Let me remove this and add a deprecation note to doc on next update

src/main/java/io/fabric8/maven/docker/assembly/DockerFileBuilder.java
```diff
@@ -195,28 +195,26 @@ private static void buildOption(StringBuilder b, DockerFileOption option, Object
...
-            } else {
-                DockerFileKeyword.RUN.addTo(b, chmod);
+            String[] userParts = assemblyUser.split(":");
+
+            for (CopyEntry entry : copyEntries) {
+                String dest = (basedir.equals("/") ? "" : basedir) + "/" + entry.destination;
+                if (userParts.length > 2) {
+                    DockerFileKeyword.USER.addTo(b, "root");
+                }
+                DockerFileKeyword.ADD.addTo(b, " --chown="
```
rhuss: Can't we just keep the `COPY` semnatics (instead of using `ADD` which works slightly different). I.e. I would add the extra command for chown to `addCopyEntries` and add an extra argument `String ownerAndGroup` to `addCopyEntries()` which can be either `null` or `"user:group"`. That way we can preserve the `COPY` semantics.

i.e. use `COPY --chown` instead of `ADD --chown`.

rhuss: @rohanKanojia I'd like to add this to the release. Do you think there is a chance to get this in for the next release  (i.e. change from `ADD` to `COPY`) ?

rohanKanojia: ah, sorry. I was busy in last few weeks. Let me update this today

rhuss: Awesome, thanks !

PR #1310, opened 2019-12-12
Pick up AWS credentials from ENV

sebastiankirsch:
With various AWS authentication mechanisms already in place, one of the most basic ones is yet missing: configuring credentials via environment variables.

See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -13,6 +13,7 @@
   - Allow killing and removing all spawned containers (#1182)
   - Deprecated "authToken" for ECR authentication in favor of "auth" (#1286)
   - Allow overriding of existing image in creation of temporary one with same tag before push ([#838](https://github.com/fabric8io/docker-maven-plugin/issues/838))
+  - Pick up AWS credentials from ENV variables (#1310)
 
 * **0.31.0** (2019-08-10)
   - Fix test cases on Windows ([#1220](https://github.com/fabric8io/docker-maven-plugin/issues/1220))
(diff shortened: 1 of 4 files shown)


sebastiankirsch:
I've adapted the changelog and also the documentation of the Extended Authentication.
Thanks for the hint!

rohanKanojia:
@sebastiankirsch : Np, Let's wait for Roland to review.

sebastiankirsch:
@rhuss I gave this some more thought and realized that there are also [Named Profiles](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) that people might use to authenticate against AWS.
As an alternative to this PR and all the other AWS authentication mechanisms in the `AuthConfigFactory`, one could include the `com.amazonaws:aws-java-sdk-core` library and simply use the `DefaultAWSCredentialsProviderChain`. This would come with a noticeable dependency chain though...

rhuss:
Sorry for the delay, year-end-stress here. For using the AWS support directly: What's about making this optional ? E.g. when the plugin is configured manually to use this dependency (within a `<plugin><dependencies>` section), then the SDK support would be used ? We could just a lookup via reflection in config factory.

wdyt ?

sebastiankirsch:
> Sorry for the delay, year-end-stress here. For using the AWS support directly: What's about making this optional ? E.g. when the plugin is configured manually to use this dependency (within a `<plugin><dependencies>` section), then the SDK support would be used ? We could just a lookup via reflection in config factory.
> 
> wdyt ?

Good idea. It would be a breaking change though (assuming the "native" support would be dropped), not sure what your stance is on this.

rhuss:
> Good idea. It would be a breaking change though (assuming the "native" support would be dropped), not sure what your stance is on this.

I mean we could keep the current support and use it if request but not official SDK dependency has been declared. 

But let's get that PR merged first, I hope I find some room until the weekend (but year-end-stress here ....)

PR #1311, opened 2019-12-20
Use AWS SDK to fetch AWS credentials

sebastiankirsch:
As discussed in #1310, credentials for ECR would be fetched via AWS SDK.

The dependency is optional, yet it would only introduce this:
```
[INFO] +- com.amazonaws:aws-java-sdk-core:jar:1.11.696:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.6.7.3:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.6.0:compile
[INFO] |  |  \- com.fasterxml.jackson.core:jackson-core:jar:2.6.7:compile

```
i.e. outside of the SDK itself, only Jackson.  
So: would you consider making this a full dependency, which would allow to remove 170loc in `AuthConfigFactory` and some more in the according test? It would also allow to get rid of the `AwsSigner4` and `AwsSigner4Request` classes, since that functionality is found in the SDK as well.

No matter what, this will make Jackson available, and with GSON already in the mix, maybe you want to prevent using Jackson via checkstyle or SonarCloud or something? Especially as long as it's optional.

I'll try to come up with a meaningful test for this...


sebastiankirsch:
My tests indicate that `<optional>true</optional>` doesn't seem to have an effect. I know that this works as expected for "normal" projects; maybe it's different for plugins?
So, it's either a no-go for this PR, incorporate the AWS SDK fully, or try to find a work-around with sub-modules and dynamic class loading...

src/main/java/io/fabric8/maven/docker/util/aws/AwsSdkAuthConfigFactory.java
```diff
@@ -0,0 +1,41 @@
+package io.fabric8.maven.docker.util.aws;
+
+import com.amazonaws.SdkClientException;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSSessionCredentials;
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
```
rhuss: Actually I would go full in with reflection: 

* Avoid any direct import or references to AWS SDK classes
* Check with `Class.forName()` whether a certain class is avalialble. 
* Instantiate object and execute method only via `javax.reflection` means, with proper error handling or fallback (e.g. necessary when there might be different signatures for different versions of the SDK).  For now, I would focus on a single version of the SDK and document that.

sebastiankirsch: Got it. Let's see when I find the time for this. :wink:

sebastiankirsch: @rhuss got it. Only thing missing is documentation and referring to it in the log.

Can the documentation be safely included in this branch or will it go live before releasing?

rhuss: Documentation is always based on the latest tag. You can just add it to the PR and we can merge on master. During release it will be merged into the dmp.fabric8.io branch from where the documentation is generated by circleci.

Thanks for the update ! I'm just on my way to a long weekend, but will have a look as soon as I'm back.

src/test/java/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.java
```diff
@@ -0,0 +1,23 @@
+package com.amazonaws.auth;
+
+import static java.lang.System.getenv;
+
+/**
+ * Shameless copy of the original for testing {@link io.fabric8.maven.docker.util.aws.AwsSdkAuthConfigFactory}.
+ * Based on <tt>com.amazonaws:aws-java-sdk-core:1.11.707</tt>.
```
rhuss: Are there any licensing issues ? Under which license the aws-java-sdk-core  is licensed (so that it allows code copying with contribution ?). Don't wanna get into struggle because of licensing (was there already :)

Is there any chance that we could implement this without reusing SDK code ?

src/main/java/io/fabric8/maven/docker/util/aws/AwsSdkAuthConfigFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+            Class<?> credentialsClass = Class.forName("com.amazonaws.auth.AWSCredentials");
+            return new AuthConfig(
+                    (String) credentialsClass.getMethod("getAWSAccessKeyId").invoke(credentials),
+                    (String) credentialsClass.getMethod("getAWSSecretKey").invoke(credentials),
+                    "none",
+                    sessionToken
+            );
+        } catch (Throwable t) {
+            log.debug("Failed to fetch AWS credentials: %s", t);
```
rhuss: I would switch to a `warn` here as it already a Amazon SDK has been detected (so the intention is clear), but this failed. We can still move on with more authentication but I would expose this more prominently so that we eventually could fix that. E.g. this could happen when the SDK evolves ...

src/main/asciidoc/inc/_authentication.adoc
```diff
@@ -187,7 +187,37 @@ In case you're using temporary security credentials provided by the AWS Security
...
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-java-sdk-core</artifactId>
+                <version>1.11.707</version>
+            </dependency>
+        </dependencies>
+    </plugin>
+</plugins>
+----
+
+This allows you to make use of all link:https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html[the options the default credential provider chain provides].
```
rhuss:
```suggestion
This extra dependency allows the usage of all link:https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html[options] that the AWS default credential provider chain provides.
```

src/test/java/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.java
rhuss: I know its only a test class, but still this could pose some issue for some folks. The good thing is that we are talking here about a tool which is not part the of any delivered application that builds with it.

sebastiankirsch: The AWS SDK is licensed under [Apache License 2.0](https://github.com/aws/aws-sdk-java/blob/master/LICENSE.txt). Afaik that's one of the most liberal licenses you could ask for.

I could remove this, but then there wouldn't be a test at all. What do you think?

rhuss: Ok, APL is ok. I had once a case where a GPL code snippet sneaked in (in another project). Was quite some hazzle to get this out again. I'm fine with it now (I will add some extra comment here)

sebastiankirsch: Yeah, licensing can be a pain...

sebastiankirsch:
@rhuss You're too fast for me: https://github.com/sebastiankirsch/docker-maven-plugin/commit/f5a514b9993e3b3a50ca2c4a51879ae4dd555ee2

rhuss:
> @rhuss You're too fast for me: [sebastiankirsch@f5a514b](https://github.com/sebastiankirsch/docker-maven-plugin/commit/f5a514b9993e3b3a50ca2c4a51879ae4dd555ee2)

sorry, this doesn't happen often these days ;-) I added your commit to master, as I already cut a new 0.33.0 release. So will make it into v0.34.0 then ...

PR #1314, opened 2020-01-08
fix: Error in boolean check which prevents logging sometimes

rhuss:
I.e. when enabled is a Boolean but not identical to Boolean.TRUE


rhuss:
Fixes @1291

PR #1318, opened 2020-01-13
Fix #1317: update ECR AuthorizationToken URL

### LINKED ISSUE #1317: Timeout on ECR AuthorizationToken when building on private EC2 instance
### Description

Trying to run a build on an AWS instance configured with an **EC2 Role on a private subnet** (no Public IP nor NAT gateway) fails with a timeout connecting to ecr.REGIONID.amazonaws.com (ex: ecr.eu-west-1.amazonaws.com)

### Info

* d-m-p version : 0.32
* Maven version (`mvn -v`) : 
```
Apache Maven 3.6.0
Maven home: /usr/share/maven
Java version: 1.8.0_232, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en, platform encoding: UTF-8
OS name: "linux", version: "4.15.0-1056-aws", arch: "amd64", family: "unix"
```
* Docker version : 19.03.5, build 633a0ea838

* How to reproduce :

1. Start an EC2 instance with:
  a) *Subnet* with access to no NAT gateway and *Auto-assign Public IP* disabled
  b) *IAM role* set 

2. SSH to the instance (from VPN or other intermediate server with Internet access and access to the instance)

3. Run **mvn -X docker:build** on the following pom.xml

```pom.xml
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <groupId>com.joaori.test</groupId>
    <artifactId>dkr-test</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <build>
        <plugins>
            <plugin>
                <groupId>io.fabric8</groupId>
                <version>0.32.0</version>
                <configuration>
                    <verbose>true</verbose>
                    <images>
                        <image>
                            <alias>${project.artifactId}</alias>
                            <build>
                                <from>123123123123.dkr.ecr.eu-west-1.amazonaws.com/foo:bar</from>
                            </build>
                        </image>
                    </images>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```

Command will hang for a while and then return a **Connection timed out**

```output
$ mvn -X docker:build
...
[INFO] --- docker-maven-plugin:0.32.0:build (default-cli) @ dkr-test ---
...
[DEBUG] DOCKER> System environment not set for variable AWS_ACCESS_KEY_ID, no AWS credentials found
[DEBUG] DOCKER> No user and password set for ECR, checking EC2 instance role
[DEBUG] DOCKER> Found instance role wwa-svc-jenkins-slave, getting temporary security credentials
[DEBUG] DOCKER> Received temporary access key ASIARWPR...
[DEBUG] DOCKER> AuthConfig: credentials from EC2 instance role
[DEBUG] DOCKER> registry = 123123123123.dkr.ecr.eu-west-1.amazonaws.com, isValid= true
[DEBUG] DOCKER> Get ECR AuthorizationToken from ecr.eu-west-1.amazonaws.com
[ERROR] DOCKER> Connect to ecr.eu-west-1.amazonaws.com:443 [ecr.eu-west-1.amazonaws.com/52.95.118.96] failed: Connection timed out (Connection timed out) [Connect to ecr.eu-west-1.amazonaws.com:443 [ecr.eu-west-1.amazonaws.com/52.95.118.96] failed: Connection timed out (Connection timed out)]
[INFO] BUILD FAILURE
[INFO] Total time:  02:12 min
[INFO] Finished at: 2020-01-13T16:15:15Z
[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.32.0:build (default-cli) on project dkr-test: Connect to ecr.eu-west-1.amazonaws.com:443 [ecr.eu-west-1.amazonaws.com/52.95.118.96] failed: Connection timed out (Connection timed out) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:docker-maven-plugin:0.32.0:build (default-cli) on project dkr-test: Connect to ecr.eu-west-1.amazonaws.com:443 [ecr.eu-west-1.amazonaws.com/52.95.118.96] failed: Connection timed out (Connection timed out)
```

### Solution

ECR was made accessible through [VPC Endpoints](https://docs.aws.amazon.com/AmazonECR/latest/userguide/vpc-endpoints.html#ecr-setting-up-vpc-create) about a year ago and the endpoint URL seems to have changed around that time to **api.ecr.REGIONID.amazonaws.com**.
[EcrExtendedAuth](https://github.com/fabric8io/docker-maven-plugin/blob/e1d842019cb52bc394555e4bbebd5b7f54ac5cfb/src/main/java/io/fabric8/maven/docker/access/ecr/EcrExtendedAuth.java#L127) needs to be updated accordingly.

joaori:


joaori:
I don't think there's any backward incompatibility.

From what I can tell, AWS seems to be serving ECR API calls in both **ecr** and **api.ecr**.

The [ECR Endpoints and Quotas](https://docs.aws.amazon.com/general/latest/gr/ecr.html) still references (old) **ecr** endpoint.

However, seems that both [python](https://github.com/boto/botocore/blob/master/botocore/data/endpoints.json) and [java](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/resources/com/amazonaws/partitions/endpoints.json#L181) SDK's are already using **api.ecr**. And [VPC Endpoints](https://docs.aws.amazon.com/AmazonECR/latest/userguide/vpc-endpoints.html) also mention it.


d-m-p will only fails on AWS if the following conditions are met:
1. It has an IAM Role defined
2. EC2 instance is on a private only subnet

Condition 1 will make **AuthConfig** to always be defined based on instance's AWS credentials.
This is also quite annoying as it won't allow for fallback to **~/.docker/config.json** credentials (assuming I'd want to deal with repo authentication externally and skip extended authentication).

Condition 2 forces having VPC Endpoints (api.ecr.. for API calls and dkr.ecr.. for push / pulls) created on the VPC so that endpoints can resolve to private IPs.

Setting up a NAT and / or proxy would still make it work I guess. Probably the reason why it hasn't been reported before.

sebastiankirsch:
I can confirm that the ECR API endpoint was updated to said value.

Since AWS itself is well-aware that there are older clients out there (the change to the SDK they provide isn't even a year old: https://github.com/aws/aws-sdk-java/commit/895452ffeb44ed3250e9eaeebbf10dbd6ed91911), I don't expect an upgrade/backward-compatibility issue there.

@joaori did anything not work for you, or why did you put in the effort of adapting this?

joaori:
I've been using the plugin for quite a while now (kudos to you all!) using a similar setup as stated. With the only exception that the instance has internet access through a (whitelisted) proxy server.

So it's using VPC endpoints to allow private access to ECR but still had to rely on the proxy for the d-m-p plugin to be able to auth using the (old) ecr endpoint.

I tried configuring it to simply bypass the plugin's auth behaviour but wasn't successful due to the issue explained above: the instance has an IAM role, so **~/.docker/config.json** fallback is never considered.
(execution output for **mvn -X deploy:build** on #1317 description)

Ended up patching the code with the new api.ecr endpoint :)
I've been using the patched plugin and seems to be working as expected so far.

rhuss:
Ok, convinced :-) Let's update the end point then.

sebastiankirsch:
I assume one could use `com.amazonaws:aws-java-sdk-ecr` to fetch the ECR authentication token instead of relying on the custom-written code incorporated in the plugin.
Similar to what I did in #1311 - would probably solve this and other "niche" use-cases...

PR #1323, opened 2020-01-25
Fix #1242: docker.container.<alias>.ip property is no longer set

### LINKED ISSUE #1242: docker.container.<alias>.ip property is no longer set
### Description

We recently attempted to upgrade from d-m-p 0.27.2 to 0.30.0, but noticed a regression in our build. It seems that the `docker.container.<alias>.ip` is no longer set?

I suspect this is caused by https://github.com/fabric8io/docker-maven-plugin/pull/1130 in 0.28.0. This PR changed the code location where the container properties are exposed. Perhaps they are exposed before the container IP has been assigned?

### Info

* d-m-p version : 0.30.0
* Maven version (`mvn -v`) : 3.6.0
* Docker version : 1.12.6

rohanKanojia:
Fix #1242 

Adding a `log.info(project.getProperties().toString()` in `StartMojo` seems to validate that properties are now getting set after this change.
```
~/work/repos/dmp-demo-project : $ mvn docker:start
[INFO] Scanning for projects...
[INFO] 
[INFO] -----------------------< io.fabric8:dockerfile >------------------------
[INFO] Building dmp-sample-dockerfile 0.30-SNAPSHOT
[INFO] --------------------------------[ war ]---------------------------------
[INFO] 
[INFO] --- docker-maven-plugin:0.33-SNAPSHOT:start (default-cli) @ dockerfile ---
[INFO] DOCKER> [test-db:latest] "rohankumar": Start container 77677e638894
[INFO] DOCKER> [test-db:latest] "rohankumar": Pausing for 5000 ms
[INFO] DOCKER> {docker.host.address=localhost, base=jetty, host.port=32772, project.artifactId=dockerfile, file=welcome.txt, docker.container.rohankumar.id=77677e638894, maven.compiler.target=1.8, docker.container.rohankumar.net.bridge.ip=172.17.0.4, docker.container.rohankumar.ip=172.17.0.4, maven.compiler.source=1.8}
[INFO] BUILD SUCCESS
[INFO] Total time:  7.099 s
[INFO] Finished at: 2020-01-25T19:29:42+05:30
~/work/repos/dmp-demo-project : $ 

```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java
+++ b/src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java
@@ -6,7 +6,6 @@ import java.util.Date;
 import java.util.Map;
 import java.util.Properties;
 
-import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.codehaus.plexus.util.StringUtils;
 
 import io.fabric8.maven.docker.access.DockerAccessException;
@@ -42,19 +41,19 @@ public class StartContainerExecutor {
 
     private StartContainerExecutor(){}
 
-    public ImmutablePair<String, Properties> startContainers() throws IOException, ExecException {
+    public String startContainers() throws IOException, ExecException {
         final Properties projProperties = projectProperties;
 
         final String containerId = hub.getRunService().createAndStartContainer(imageConfig, portMapping, gavLabel, projProperties, basedir, containerNamePattern, buildDate);
 
         showLogsIfRequested(containerId);
-        Properties exposedProperties = exposeContainerProps(containerId);
+        projProperties.putAll(queryContainerProperties(containerId));
         waitAndPostExec(containerId, projProperties);
 
-        return new ImmutablePair<>(containerId, exposedProperties);
+        return containerId;
     }
 
-    private Properties exposeContainerProps(String containerId)
+    public Properties queryContainerProperties(String containerId)
         throws DockerAccessException {
         String propKey = getExposedPropertyKeyPart();
         Properties exposedProperties = new Properties();
--- a/src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java
+++ b/src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java
@@ -41,13 +41,14 @@ public class StartContainerExecutor {
 
     private StartContainerExecutor(){}
 
-    public String startContainers() throws IOException, ExecException {
+    public String startContainer() throws IOException, ExecException {
         final Properties projProperties = projectProperties;
 
         final String containerId = hub.getRunService().createAndStartContainer(imageConfig, portMapping, gavLabel, projProperties, basedir, containerNamePattern, buildDate);
 
         showLogsIfRequested(containerId);
-        projProperties.putAll(queryContainerProperties(containerId));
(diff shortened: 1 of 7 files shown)


src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java
```diff
@@ -41,39 +42,41 @@
 
     private StartContainerExecutor(){}
 
-    public String startContainers() throws IOException, ExecException {
+    public AbstractMap.SimpleEntry<String, Properties> startContainers() throws IOException, ExecException {
```
rhuss, line 45: I think we already include Apache Commons Lang, so we could use https://commons.apache.org/proper/commons-lang/javadocs/api-3.9/org/apache/commons/lang3/tuple/ImmutablePair.html here, which better declares the intent imo

rhuss (commented):
Thanks ! looks good to me, i have a minor suggestion (see below)

src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java
```diff
@@ -41,39 +42,41 @@
...
         throws DockerAccessException {
         String propKey = getExposedPropertyKeyPart();
+        Properties exposedProperties = new Properties();
 
         if (StringUtils.isNotEmpty(exposeContainerProps) && StringUtils.isNotEmpty(propKey)) {
             Container container = hub.getQueryService().getMandatoryContainer(containerId);
 
             String prefix = addDot(exposeContainerProps) + addDot(propKey);
-            projectProperties.put(prefix + "id", containerId);
+            exposedProperties.put(prefix + "id", containerId);
```
rhuss, line 66: what happens to `projectProperties` in this class. Is this needed anymore ? If not maybe its better to remove it then ?

rohanKanojia: I think it's still used, it's being passed to `createAndStartContainer`:
https://github.com/fabric8io/docker-maven-plugin/blob/493aa8be14c642fc148c0a0eed717945ef9ef40f/src/main/java/io/fabric8/maven/docker/service/helper/StartContainerExecutor.java#L47

It seems to be consumed by `ContainerCreateConfig#environment()` :
https://github.com/fabric8io/docker-maven-plugin/blob/493aa8be14c642fc148c0a0eed717945ef9ef40f/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java#L75-L77

rhuss:
@rohanKanojia I checked the fix now and I don't believe it adds any change to the property handling as it was before. Instead I suspect that your test above would run as well with the latest 0.32.0 release.

I cleaned up the pr a bit and also improved the robustness of the tests a bit.

wrt/ to #1242 : We really need a reproducer of this issue so that we can fix this properly. In general I'm pretty sure that the IP property is still working as expected (verified it with several runs). @rohanKanojia could you please just retry the test that you mentioned in the description ?

@jpraet would it be possible to provide an example where the ip resolution does not work for you ?

rohanKanojia:
Seems like you're right. The test in the description seems to be working with 0.32.0 also. I'm not sure how. I remember at the time I was testing the initial fix it wasn't getting set(or maybe I didn't test it properly).

rhuss:
ok, so let me merge this improvements and start a fresh in case the original issue is still persistent.

PR #1336, opened 2020-04-22
Update RegistryService.java

### LINKED ISSUE #1337: pulling with multiple accounts from the same registry does not work
The authentication mechanism using the maven settings.xml doesn't work as [described](https://maven.fabric8.io/#authentication) 
We have a maven build that needs to pull 2 docker images which are maintained by separate organisations but are both  located in the gitlab docker registry.
The documentation says "If you have multiple accounts for the same registry, the second user can be specified as part of the ID", this means that the following setting.xml should work

`...
<servers>
  <server>
    <id>registry.gitlab.com/company-a</id>
    <username>user</username>
    <password>psw</password>
  </server>
  <server>
    <id>registry.gitlab.com/company-b</id>
    <username>user</username>
    <password>psw</password>
  </server>
</servers>
...`
which then should allow for the following plugin configuration
`...
<plugin>
	    <groupId>io.fabric8</groupId>
	    <configuration>
          <images>
              <image>
            		<name>registry.gitlab.com/company-a/app:1.0.0</name>	
              </image>
              <image>
            		<name>registry.gitlab.com/company-b/anotherapp:2.0.0</name>	
              </image>
          </images>
      </configuration>
</plugin>
...`

However the above doesn't work, the plugin fails to obtain credentials and is unable to pull
I can only get this to work by having in the settings.xml
`...
<server>
  <id>registry.gitlab.com</id>
  <username>user</username>
  <password>psw</password>
</server>
...`
but this results in being able to pull only one of the images as the credentials don't work for the second image.

I found the following issue #174 which is related to this issue but only talks about pushing images

I have traced this in the code and found that when pulling io.fabric8.maven.docker.util.AuthConfigFactory.createAuthConfig is being called from the io.fabric8.maven.docker.service.RegistryService with the user parameter set to null. If this user parameter is passed then the behaviour is as expected.  

I have submitted a PR that fixes this issue
#1336

Best Regards,

Johan

johanra:
Fix #1337 take into account the imageName.getUser() when determining credentials for the pull

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/RegistryService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/RegistryService.java
@@ -98,7 +98,7 @@ public class RegistryService {
             imageName.getRegistry(),
             registryConfig.getRegistry());
         docker.pullImage(imageName.getFullName(),
-                         createAuthConfig(false, null, actualRegistry, registryConfig), actualRegistry);
+                         createAuthConfig(false, imageName.getUser(), actualRegistry, registryConfig), actualRegistry);
         log.info("Pulled %s in %s", imageName.getFullName(), EnvUtil.formatDurationTill(time));
         pullManager.pulled(image);
 


rohanKanojia:
@johanra: Could you please provide some input on what this PR is trying to fix?  Some test validating desired behavior would be nice

johanra:
@rohanKanojia the issues is described in issue #1337

rhuss:
Thanks ! Looks good to me and sorry for the delay ...

PR #1347, opened 2020-06-12
bug fix for Volume binding bug in windows #1338

### LINKED ISSUE #1338: Volume binding bug in windows
### Description
I think volume binding in windows has a bug.
Consider this snippet
```
<volumes>
  <bind>
        <volume>${settings.localRepository}/../:/var/maven/.m2</volume>
   </bind>
</volumes>
```
on a windows pc it can resolve to C:\Users\giovanni\.m2\repository/../
If you run the plugin with this volume configuration the container does not start and report File not found
If you use an explicit Windows path value like this one
```
<volumes>
  <bind>
        <volume>C:\Users\giovanni\.m2\:/var/maven/.m2</volume>
   </bind>
</volumes>
```
the container starts correctly
Looking at the code it seems that in class VolumeBindingUtil

`public static String resolveRelativeVolumeBinding(File baseDir, String bindingString) `

the code applies
 
`resolvedFile.getCanonicalFile().getAbsolutePath()`

only to relative paths

On windows the canonicalization operation it's needed even on absolute paths since it translates 
C:\Users\giovanni\.m2\repository/../ to C:\Users\giovanni\.m2

### Info

* d-m-p version : all 
* Maven version (`mvn -v`) : 3.6.1

I don't think the problem is bound to a specific maven version

giovannicuccu:
Hi this is a pull request for issue #1338 
I fixed only the specific issue I was having because when I tried to make the code more generic some tests were failing. The current code considers the initial rel of 'rel:/path/to/container/dir' as an absolute path which is not, when I tried to align the code considering rel as a relative some tests were failing and I do not have enough knowledge about this project in order to make a wider fix.

Fix #1338

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/VolumeBindingUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/VolumeBindingUtil.java
@@ -1,6 +1,7 @@
 package io.fabric8.maven.docker.util;
 
 import io.fabric8.maven.docker.config.RunVolumeConfiguration;
+import org.apache.commons.lang3.SystemUtils;
 
 import java.io.File;
 import java.io.IOException;
@@ -53,6 +54,11 @@ public class VolumeBindingUtil {
      * A dot representing the current working directory
      */
     private static final String DOT = ".";
+
+    /**
+     * A dot representing the parent directory
+     */
+    private static final String DOUBLE_DOT = "..";
 
     /**
      * A tilde representing the current user's home directory
@@ -165,6 +171,14 @@ public class VolumeBindingUtil {
         // - datavolume:/var/lib/mysql
 
         String[] pathParts = bindingString.split(":");
+        if (SystemUtils.IS_OS_WINDOWS &&  WINDOWS_DRIVE_PATTERN.matcher(bindingString).matches() && pathParts.length>2) {
+            String[] winPathParts = new String[pathParts.length-1];
+            winPathParts[0]=pathParts[0]+":"+pathParts[1];
+            for (int i=1;i<winPathParts.length;i++) {
+                winPathParts[i]=pathParts[i+1];
+            }
+            pathParts=winPathParts;
+        }
         String localPath = pathParts[0];
 
         if (isRelativePath(localPath)) {
@@ -177,6 +191,13 @@ public class VolumeBindingUtil {
                 }
                 resolvedFile = resolveAbsolutely(localPath, baseDir.getAbsolutePath());
             }
+            try {
+                localPath = resolvedFile.getCanonicalFile().getAbsolutePath();
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to canonicalize '" + resolvedFile + "'");
+            }
+        } else if (SystemUtils.IS_OS_WINDOWS && (localPath.contains(DOUBLE_DOT))) {
+            File resolvedFile=new File(localPath);
             try {
                 localPath = resolvedFile.getCanonicalFile().getAbsolutePath();
             } catch (IOException e) {
--- a/src/test/java/io/fabric8/maven/docker/util/VolumeBindingUtilTest.java
+++ b/src/test/java/io/fabric8/maven/docker/util/VolumeBindingUtilTest.java
@@ -1,6 +1,7 @@
 package io.fabric8.maven.docker.util;
 
 import io.fabric8.maven.docker.config.RunVolumeConfiguration;
+import org.junit.Assume;
 import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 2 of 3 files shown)


src/test/java/io/fabric8/maven/docker/util/VolumeBindingUtilTest.java
```diff
@@ -302,4 +302,18 @@ public void testIsRelativePathForWindows() {
         assertFalse(isRelativePath("C:\\"));                               // C:\
         assertFalse(isRelativePath("\\"));                                 // \
     }
+
+    /**
+     * Insures that a host volume binding string that contains a windows path with .. is correctly canonicalized
+     */
+    @Test
+    public void testResolveAbsoluteWindowsVolumePath() {
+        String volumeString = format(BIND_STRING_FMT, "C:\\dir/subdir/../", CONTAINER_PATH);
```
rohanKanojia, line 311: your test is passing on windows but not on Linux. I think we should refactor this to pick different volume strings depending upon underlying OS

rhuss (commented):
I think there is a typo with DOUBLE_DOT

src/main/java/io/fabric8/maven/docker/util/VolumeBindingUtil.java
```diff
@@ -164,6 +170,14 @@ public static String resolveRelativeVolumeBinding(File baseDir, String bindingSt
         // - datavolume:/var/lib/mysql
 
         String[] pathParts = bindingString.split(":");
+        if (SystemUtils.IS_OS_WINDOWS &&  WINDOWS_DRIVE_PATTERN.matcher(bindingString).matches() && pathParts.length>2) {
```
rhuss, line 173: `> 2` or maybe `>= 2` ? (question do expect more 2 colons or only a single colon ?)

src/main/java/io/fabric8/maven/docker/util/VolumeBindingUtil.java
```diff
@@ -181,6 +195,13 @@ public static String resolveRelativeVolumeBinding(File baseDir, String bindingSt
             } catch (IOException e) {
                 throw new RuntimeException("Unable to canonicalize '" + resolvedFile + "'");
             }
+        } else if (SystemUtils.IS_OS_WINDOWS && (localPath.contains(DOUBLE_DOT))) {
```
rhuss, line 198: I think you can always resolve the absolut path, regardless wether its on Window/Linux or whether it contains .. or not. In the worst case its a no-operation, but it would be easier to read.

src/test/java/io/fabric8/maven/docker/util/VolumeBindingUtilTest.java
rhuss: I wonder whether there is annotation for JUnit to run tests on specific platforms only. If this is possible we can just create two tests, one for windows and one for unix like operating systems.

giovannicuccu: As far as I know the annotation for running Test on a specific platform (@EnabledOnOs) is Junit 5 specific. I added 
`Assume.assumeTrue(System.getProperty("os.name").toLowerCase().startsWith("win")); `
in the test code

src/main/java/io/fabric8/maven/docker/util/VolumeBindingUtil.java
```diff
@@ -53,6 +54,11 @@
      */
     private static final String DOT = ".";
 
+    /**
+     * A dot representing the parent directory
+     */
+    private static final String DOUBLE_DOT = ".";
```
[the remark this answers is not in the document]
giovannicuccu: It was a typo the real value shoud be "..". 
Fixed.
the test passes because I'm using String.contains that is also true for "."

src/main/java/io/fabric8/maven/docker/util/VolumeBindingUtil.java
giovannicuccu: My first try was exactly this, but doing so make cause some test failures and I don't know this project enough to understand how to fix the additional failures

giovannicuccu:
I pushed a new commit con my branch, let me know if you see the new version

src/test/java/io/fabric8/maven/docker/util/VolumeBindingUtilTest.java
rohanKanojia: I think for now this is okay too. Let's try to get this merged. We can refactor this later in case we migrate to Junit5 in future

rohanKanojia (approved):

PR #1349, opened 2020-06-13
Handle authentication tokens returned from credential helpers.

### LINKED ISSUE #1348: Authentication tokens from credential helpers incorrectly used as passwords
### Description

The documentation for ```docker login``` at https://github.com/docker/docker-ce/blob/master/components/cli/docs/reference/commandline/login.md says:

> If the secret being stored is an identity token, the Username should be set to ```<token>```

This means that when a credential helper response is returned with the username equal to ```<token>```, the AuthConfig constructed should use the identity token constructor instead of the username/password constructor.

This is similar to #1249, except there the credentials were stored in ```config.json```, while here they are returned from a credential helper.

* d-m-p version : 0.33-SNAPSHOT
* Maven version (`mvn -v`) : 3.6.3 (and others, I expect)

When using an Azure Container Registry and a credential helper, run ```az acr login -n <acr-name>``` to retrieve an authentication token and store in the credential helper.

Then, run a ```docker-maven-plugin:push``` that needs to use the credentials. It will fail with message like ```Unable to push 'example-image:latest' to registry 'acrname.azurecr.io' : unauthorized: aad access token with sp failed unknown error  [unauthorized: aad access token with sp failed unknown error ]```.

wrosenuance:
This PR fixes #1348 by checking for the special username ```<token>``` that the docker login command uses to indicate the password is actually an authentication token.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,7 +1,9 @@
 # ChangeLog
 
-* **0.33.0** (2020-01-21)
+* **0.33-SNAPSHOT**
   - Handle authentication tokens returned from credential helpers ([#1348](https://github.com/fabric8io/docker-maven-plugin/issues/1348))
+
+* **0.33.0** (2020-01-21)
   - Update to jnr-unixsocket 0.25 to solve concurrency issues ([#552](https://github.com/fabric8io/docker-maven-plugin/issues/552))
   - Udate ECR AuthorizationToken URL to new endpoint ([#1317](https://github.com/fabric8io/docker-maven-plugin/issues/1317))
   - Allow including `com.amazonaws:aws-java-sdk-core` as plugin dependency to pick up various forms of AWS credentials with which to authenticate at AWS ECR ([#1311](https://github.com/fabric8io/docker-maven-plugin/issues/1311))
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,7 @@
 # ChangeLog
 
 * **0.33.0** (2020-01-21)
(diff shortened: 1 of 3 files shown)


doc/changelog.md
```diff
@@ -1,6 +1,7 @@
 # ChangeLog
 
 * **0.33.0** (2020-01-21)
+  - Handle authentication tokens returned from credential helpers ([#1348](https://github.com/fabric8io/docker-maven-plugin/issues/1348))
```
rhuss, line 4: Please move that up to a **0.33-SNAPSHOT** section as this will become the changes for the next release. 0.33.0 is already released.

wrosenuance: Done!

wrosenuance:
> Maybe you can add a CredentialHelpeClientTest, make toAuthConfig() package visible and just test this method, so that you don't have to bother with executing external commands ?

Sure! I pushed another commit, is this what you were looking for?

carlitorendon:
@wrosenuance / @rhuss any ETA on the review / this making it to the release?

wrosenuance:
@carlitorendon I am not sure - I think (I hope) it's ready to go, I'm waiting for @rhuss to approve/merge though.

carlitorendon:
Yeah, I also looked at the changes and they look good (though obviously I'm not deeply familiar with the plugin 😀). If @rhuss is unavailable maybe one of the other maintainers can take it up.

rhuss:
Sorry for the delay, busy times ;-( I will be on PTO now, so let's get that merged as the test has been added.

carlitorendon:
Thanks a lot @rhuss !

PR #1355, opened 2020-06-24
Fix for Issue 1354 - Verify that REPO_TAGS can be processed as a JsonArray

robinr618:
REPO_TAGS member may be null (JsonNull), so we need to make sure it is an instance of JsonArray


rhuss:
Thanks ! looks good to me

PR #1361, opened 2020-07-28
1025: migrate from Joda Time to java.time

denisa:
use java.time.ZonedDateTime directly instead of custom Timestamp class.

This is part of migrating to java 8


denisa:
Let me know if you want a new PR with a single commit.

rhuss:
Thanks ! Looks good, let's get it merged now

Issue #1243, opened 2019-07-04
image build skipped should not be pushed by default


ebaotech-oss:
Hi,

Currently, when one of the image build is skipped (example below), push will still try to push all two images, is it possible for push to honor image.build.skip property?

<images>
    <image>
      <build>
        <name>image1</name>
        <skip>${maybe.skipped.1}</skip>
     </build>
   </image>
    <image>
      <build>
        <name>image2</name>
        <skip>${maybe.skipped.2}</skip>
     </build>
   </image>
<images>

Thanks

rhuss:
I agree that there should be a possibility to skip the push for certain images. However, I would introduce a `skipPush` field (like `skipTag`), to deliberately specify that intent. This is mostly to not break backwards compatibility.

Unfortunately I don't have any spare time at the moment, so if you are interested in that feature, feel free to submit a pull request and I will help you to integrate it.

PR #1368, opened 2020-08-27
Add skipPush option to build image configuration (#1243)

mattnelson:
Add skipPush option to build image configuration from #1243

Was getting test failures locally with `AuthConfigFactoryTest` due to my `~/.kube/config` setup, wasn't sure what needed to be done to get those to pass.

Example of the output from my build which consumed these proposed plugin changes.
```
[INFO] --- docker-maven-plugin:0.33-SNAPSHOT:push (push-image) @ my-module ---
[INFO] DOCKER> [my-registry/my-image:image1] : Skipped pushing
[INFO] DOCKER> Credentials helper reply for "docker-credential-desktop" is 0.6.0
[INFO] DOCKER> The push refers to repository [my-registry/my-image]
[INFO] DOCKER> image2: digest: sha256:abcd size: 1234
[INFO] DOCKER> Temporary image tag skipped. Target image 'my-registry/my-image:image2' already has registry set or no registry is available
[INFO] DOCKER> Pushed my-registry/my-image:image2 in 18 seconds
```


src/main/java/io/fabric8/maven/docker/service/RegistryService.java
```diff
@@ -45,6 +45,11 @@ public void pushImages(Collection<ImageConfiguration> imageConfigs,
                            int retries, RegistryConfig registryConfig, boolean skipTag) throws DockerAccessException, MojoExecutionException {
         for (ImageConfiguration imageConfig : imageConfigs) {
             BuildImageConfiguration buildConfig = imageConfig.getBuildConfiguration();
+            if (buildConfig.skipPush()) {
```
rhuss, line 48: I know there is a test missing in `RegistryServiceTest` for pushImages, but it would be cool if you could add one to test this change, too (then the Sonar check would be happy, too). It shouldn't be hard to add, as it should work the same as the pull test, too.

rhuss (commented):
Looks good, but I would add a test for pushImages, too in RegistryServiceTest if possible.

If you dont have time for that, we can merge this PR nevertheless

Thanks a lot for your contribution !

rhuss:
Ah, and could you please also add an enty to the changelog.md ? thanks !

mattnelson:
> Looks good, but I would add a test for pushImages, too in RegistryServiceTest if possible.

> Ah, and could you please also add an enty to the changelog.md ? thanks !

both done in 2dba1fb

rhuss (approved):
Cool, thanks a ton !

PR #1373, opened 2020-09-15
Fix #1372: Plugin now resolves ARGs provided in BuildImageConfiguration

### LINKED ISSUE #1372: ARG still not working properly in FROM
### Description
This was already discussed extensively since 2017 at #859 and partially fixed by #1299 with a new release today. I tested it, and there are several cases where it still does not work.

MWE pom.xml:
```
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">

    <groupId>com.example</groupId>
    <artifactId>dmp-test</artifactId>
    <version>1.0.0-SNAPSHOT</version>
    <packaging>pom</packaging>

    <properties>
        <java.version>1.8</java.version>
        <resource.delimiter>@</resource.delimiter>
        <maven.compiler.source>${java.version}</maven.compiler.source>
        <maven.compiler.target>${java.version}</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>

        <busybox.version>latest</busybox.version>
    </properties>

    <build>
        <plugins>
            <plugin>
                <groupId>io.fabric8</groupId>
                <version>0.34.0</version>
                <configuration>
                    <images>
                        <image>
                            <alias>${project.artifactId}</alias>
                            <build>
                                <contextDir>${project.basedir}</contextDir>
                                <args>
                                    <busyboxVersion>${busybox.version}</busyboxVersion>
                                    <busyboxVersionName>${busybox.version}</busyboxVersionName>
                                </args>
                            </build>
                        </image>
                    </images>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```

I want to use busyboxVersion, as supplied by Maven, in my Dockerfile, so I do this:
```
ARG busyboxVersion
FROM busybox:${busyboxVersion}
```

But upon building, I get this:
```
[INFO] DOCKER> [dmp-test:1.0.0-SNAPSHOT] "dmp-test": Created docker-build.tar in 83 milliseconds
[ERROR] DOCKER> Unable to check image [busybox:${busyboxVersion}] : {"message":"no such image: busybox:${busyboxVersion}: invalid reference format"} (Bad Request: 400) [{"message":"no such image: busybox:${busyboxVersion}: invalid reference format"} (Bad Request: 400)]
```

Apparently, the ARG is not resolved, although this is the correct format as per [the documentation](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact). So, I change it to this:
```
ARG busyboxVersion
FROM busybox:$busyboxVersion
```

And get this:
```
[INFO] DOCKER> [dmp-test:1.0.0-SNAPSHOT] "dmp-test": Created docker-build.tar in 122 milliseconds
[ERROR] DOCKER> Unable to check image [busybox:] : {"message":"no such image: busybox:: invalid reference format"} (Bad Request: 400) [{"message":"no such image: busybox:: invalid reference format"} (Bad Request: 400)]
```

The ARG is now resolved, but has no content. Hence, I test this:
```
ARG busyboxVersion
FROM busybox:latest
ARG busyboxVersionName
LABEL busyboxVersion=${busyboxVersion}
LABEL busyboxVersionName=${busyboxVersionName}
```

The image is now built (since the version is again hard-coded), but upon inspecting it in Docker, I get this:
```
"Labels": {
    "busyboxVersion": "",
    "busyboxVersionName": "latest"
}
```

Once again, the ARG was resolved to an empty string, while the second ARG was interpreted correctly. I further this test like so:
```
FROM busybox:latest
ARG busyboxVersion
ARG busyboxVersionName
LABEL busyboxVersion=${busyboxVersion}
LABEL busyboxVersionName=${busyboxVersionName}
```

Now inspecting the Docker image, everything is alright:
```
"Labels": {
    "busyboxVersion": "latest",
    "busyboxVersionName": "latest"
}
```

### Info

* d-m-p version : 0.34.0
* Maven version (`mvn -v`) :
```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: C:\Program Files\apache-maven-3.6.3\bin\..
Java version: 11.0.7, vendor: AdoptOpenJDK, runtime: C:\Program Files\AdoptOpenJDK\jdk-11.0.7.10-hotspot
Default locale: en_US, platform encoding: Cp1252
```
* Docker version : 19.03.12
* If it's a bug, how to reproduce : see above
* If it's a feature request, what is your use case : -
* Sample project : see above

rohanKanojia:
Fix #1372 

Refactored `DockerFileUtil.extractBaseImages` to accept a HashMap of
Build args which would be passed from BuildService in order to resolve
ARG values which are provided from plugin configuration

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
@@ -53,9 +53,9 @@ public class DockerFileUtil {
      * @param interpolator interpolator for replacing properties
      * @return LinkedList of base images name or empty collection if none is found.
      */
-    public static List<String> extractBaseImages(File dockerFile, FixedStringSearchInterpolator interpolator) throws IOException {
+    public static List<String> extractBaseImages(File dockerFile, FixedStringSearchInterpolator interpolator, Map<String, String> argsFromBuildConfig) throws IOException {
         List<String[]> fromLines = extractLines(dockerFile, "FROM", interpolator);
-        Map<String, String> args = extractArgs(dockerFile, interpolator);
+        Map<String, String> args = extractArgs(dockerFile, argsFromBuildConfig, interpolator);
         Set<String> result = new LinkedHashSet<>();
         Set<String> fromAlias = new HashSet<>();
         for (String[] fromLine :  fromLines) {
@@ -80,8 +80,8 @@ public class DockerFileUtil {
      * @param interpolator interpolator for replacement
      * @return HashMap of arguments or empty collection if none is found
      */
-    public static Map<String, String> extractArgs(File dockerfile, FixedStringSearchInterpolator interpolator) throws IOException {
-        return extractArgsFromLines(extractLines(dockerfile, "ARG", interpolator));
+    public static Map<String, String> extractArgs(File dockerfile, Map<String, String> argsFromBuildConfig, FixedStringSearchInterpolator interpolator) throws IOException {
+        return extractArgsFromLines(extractLines(dockerfile, "ARG", interpolator), argsFromBuildConfig);
     }
 
     /**
@@ -155,14 +155,23 @@ public class DockerFileUtil {
      * @param argLines list of string arrays containing lines with words
      * @return map of parsed arguments
      */
-    protected static Map<String, String> extractArgsFromLines(List<String[]> argLines) {
+    static Map<String, String> extractArgsFromLines(List<String[]> argLines, Map<String, String> argsFromBuildConfig) {
         Map<String, String> result = new HashMap<>();
         for (String[] argLine : argLines) {
             if (argLine.length > 1) {
-                updateMapWithArgValue(result, argLine[1]);
+                updateMapWithArgValue(result, argsFromBuildConfig, argLine[1]);
             }
         }
         return result;
+    }
+
+    static String resolveArgValueFromStrContainingArgKey(String argString, Map<String, String> args) {
+        if (argString.startsWith("$") && args.containsKey(argString.substring(1))) {
+            return args.get(argString.substring(1));
+        } else if (argString.startsWith("${") && argString.endsWith("}") && args.containsKey(argString.substring(2, argString.length() - 1))) {
+            return args.get(argString.substring(2, argString.length() - 1));
+        }
+        return null;
     }
 
     private static String resolveImageTagFromArgs(String imageTagString, Map<String, String> args) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -155,16 +155,25 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
...
+                updateMapWithArgValue(result, argsFromBuildConfig, argLine[1]);
             }
         }
         return result;
     }
 
+    static String resolveArgValueFromStrContainingArgKey(String argString, Map<String, String> args) {
+        if (argString.startsWith("$") && args.containsKey(argString.substring(1))) {
+            return args.get(argString.substring(1));
+        } else if (argString.startsWith("${") && argString.endsWith("}") && args.containsKey(argString.substring(2, argString.length() - 1))) {
```
rhuss, line 171: I think spaces are also allowed here like in `${ foobar }`. So for extracting the variable name it might be to user a regular expression. you then could also collapse the if statements into a single regular expression.

rhuss: I.e I think a REs like 

```
^\$([^{}]+)$

^\$\{\s*([^{}\s]+)\s*\}$
```

would do the trick (not verified though)

rhuss (commented):
Thanks @rohanKanojia ! I will have a look now, and let's get that fix in ASAP.

I'm planning a release this weekend to fix some of the more pressing issues.

rhuss:
@rohanKanojia let's get it merged now to fix the issue at hand, but would love to allow also variables provided in an extended way (as mentioned in the pr comments)

src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
rohanKanojia: I'm trying out an arg with spaces like this in this Dockerfile:
```Dockerfile
ARG busyboxVersion
FROM busybox:${ busyboxVersion }
```
When I trying to build it with docker, I'm getting an error:
```bash
docker : $ docker build -t test-dockerfile --build-arg busyboxVersion=latest .
Sending build context to Docker daemon 3.072 kB
Error response from daemon: Dockerfile parse error line 2: FROM requires either one or three arguments
```

PR #1376, opened 2020-09-19
Config to specify isolation technology for container

rahul-kulkarni73:


dermot-hardy:
Hi.  Is there a way to reference the plugin built from this branch?

rohanKanojia:
I think https://jitpack.io/ might help

rhuss (commented):
Thanks a lot for the PR ! Sound reasonable to me, we will consider this for one of the next releases (but please note that this plugin is on low maintenance mode, which means things might take a bit longer than normally).

Do you know since which API version the `Isolation` option is available for Docker ?

rahul-kulkarni73:
Kernel namespaces were introduced between kernel version 2.6.15 and 2.6.26. This means that since July 2008 (date of the 2.6.26 release ), namespace code has been exercised and scrutinized on a large number of production systems.
https://man7.org/linux/man-pages/man7/namespaces.7.html

Seems like docker allowed process isolation in this version 
https://github.com/docker/docker-ce/releases/tag/v18.09.1 (Jan 2019)
https://github.com/docker/engine/pull/81 

Found this docker doc dated 2016 https://www.docker.com/sites/default/files/WP_IntrotoContainerSecurity_08.19.2016.pdf.

rahul-kulkarni73:
@rhuss Any updates on possibility of getting this change in ?

rohanKanojia:
Sorry for late response. I think we should get this in for our upcoming release (`0.35.0`) Let's try to get this merged in coming week so that we can have it in our next release

rhuss (commented):
Looks good to me, and I guess that since this feature is added on request, it's not harmful when running against an older Docker that is not supporting that parameter (as in this case the user needs to take care whether to configure `Isolation` or not)

Please add a `doc/changelog.md` entry, then I think we are good to merge.

rohanKanojia (approved):

PR #1383, opened 2020-09-28
Follow up PR for https://github.com/fabric8io/docker-maven-plugin/pull/1373

rohanKanojia:
Follow up PR for #1373 

Use regex in order to resolve ARG value

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
@@ -43,6 +43,8 @@ import org.yaml.snakeyaml.Yaml;
  */
 public class DockerFileUtil {
 
+    private static final String ARG_PATTERN_REGEX = "\\$(?:\\{(.*)\\}|(.*))";
+
     private DockerFileUtil() {}
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -166,10 +166,14 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
     }
 
     static String resolveArgValueFromStrContainingArgKey(String argString, Map<String, String> args) {
-        if (argString.startsWith("$") && args.containsKey(argString.substring(1))) {
-            return args.get(argString.substring(1));
-        } else if (argString.startsWith("${") && argString.endsWith("}") && args.containsKey(argString.substring(2, argString.length() - 1))) {
-            return args.get(argString.substring(2, argString.length() - 1));
+        Pattern argPattern = Pattern.compile("\\$(?:\\{(.*)\\}|(.*))");
```
rhuss, line 169:
```suggestion
        Pattern argPattern = Pattern.compile("^\\$(?:\\{(.*)\\}|(.*))$");
```

Align with start and end of the string. Otherwise you would also match `bla$ad` or `${foo}bar`

rhuss: But please verify my assumption within the tests, too, please

src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java
```diff
@@ -166,10 +166,14 @@ public static FixedStringSearchInterpolator createInterpolator(MojoParameters pa
...
+        Matcher matcher = argPattern.matcher(argString);
+        if (matcher.find()) {
```
rhuss, line 171: Should you use `matches()` here as find also matches on substrings ? (if you `^` and `$` this should make nor difference, but I still think "match" is cleare)

rhuss (commented):
Thanks, looks good ! Some minor comments though ;)

rhuss (approved):
Thanks ! Looks good ...

Issue #535, opened 2016-08-08
for flat classpath should we try put dependencies in a separate directory by default?


jstrachan:
I saw this blog recently which was discussing why fat jars can be suboptimal:
http://product.hubspot.com/blog/the-fault-in-our-jars-why-we-stopped-building-fat-jars

each build requires re-downloading all jars again; even when no dependencies change.

I wonder if there's a way to promote base image sharing with flat classpath apps? e.g. if folks are doing CI/CD and doing lots of builds per day of the same microservice; most times its only the microservices's code changing; the dependencies will rarely change. So it'd be awesome if one docker layer was the dependencies; then another docker layer was the apps jar. Then this would promote docker caching.

I wonder if we could implement this by default by using a dependencies folder? e.g. for microservice foo we create an install like this...

```
/app
  foo.jar
  dependencies/
     jackson-core-2.7.4.jar
     ...
```

then the Dockerfile contains 2 ADD statements; one for the dependencies, one for the builds jar. 

Then hopefully docker will try cache the dependency layer as much as possible; which will help speed up docker builds, pushes, pulls

rhuss:
When we use flat classpath files (which really makes sense for speeding up deployments), then plainf caching should be already easily possible, if not already working. One would need to ensure the proper order of `ADD` commands used, so that dependencies come first. 

The alternative would be to operate with two images: One base image holding the dependencies and a sub-image holding only the application code. This would work even without relying on Docker caching to work. This approach could probably already done (since d-m-p supports already multiple images), although they will be built again for each run (but then also probably already fully cached when nothing changes). 

But working with one image is probably easier, so let's have a look how we add things today.

Also, in order to fully replace the previous hawt-app functionality something like #257 would be nice (although it could be realized with the dependency plugin, too, it makes sense add a simple classpath file generation from a classpath file, too).

jstrachan:
I watched this demo today; its a kinda argument for app servers (as in you only rebuild / deploy your WAR and they tend to be really tiny as if you stick to JEE then there's no dependencies in your WAR)
https://www.youtube.com/watch?v=aZH01HzUlww

its certainly a compelling argument; though personally its not quite compelling enough given the flexibility, simplicity & awesomeness of spring boot :)

So adding files individually sounds like it could make gazillions of images if a typical app has around 40 transitive dependencies and you have many images? Though if they are sorted in a good order; folks could share some base images across apps I guess (at the expense of docker image explosion).

[...]

PR #1387, opened 2020-10-05
Fix #554: Add support for multiple copy layers using multiple assemblies

### LINKED ISSUE #554: Support for multiple COPY layers
We have large folders inside our container and it would be nice if we could put them on different COPY layers. 
This could be archived for example with multiple assembly files, and for each assembly file create a different COPY statement in the dockerfile.

zer0keefie:
Fix #554 

This was largely motivated by the new Spring Boot support for building docker images. I wanted a similar result (separate layers for release dependencies, snapshot dependencies, and the project artifact) without having to switch maven plugins. It looked like the DockerFileBuilder already had enough support for what I needed, and I just had to add support for multiple assemblies.

I _think_ I hit all the points I was aiming for:
- Allow adding multiple <assembly> elements (in an <assemblies> element, because that seemed cleaner), and create a layer for each of those when generating the Dockerfile.
- Preserve the current semantics when a Dockerfile is supplied: require each <assembly> be referenced in a COPY command.
- Don't force a move to the <assemblies> format: <assembly> directly in <build> is still supported, and you can mix the two.
- Assemblies each have to have a unique name, to avoid collisions. I'm not positive that would happen, but I'd rather just avoid it.
- I think I have `docker:watch` working with multiple assemblies.
- As a result of that, JIB mode should work too.
- Updated the docs.

Because I had a specific use case, and I suspect it's pretty common, I added artifact descriptors for dependencies, release dependencies, and snapshot dependencies.

This covers #554, but I made no attempt to avoid the extra posts discussed in #535.


rhuss:
@zer0keefie Thanks a lot, that looks really awesome! Unfortunately, I'm quite loaded with work (hmm, probably don't have to tell that all the time as this probably always true ;-), and since this PR is not a small one, it might take a bit until I make my way through. At a first look, it looks like that all mandatory parts are included (like docs and tests). 

Going to jump on to this as quick as possible, but probably only on weekends. 

@rohanKanojia if you have some spare cycles it would be awesome if you could add some review comments, too.

rhuss:
Looks like that the latest rebase introduced a NPE in the e2e-tests.

rohanKanojia:
strange, I'll take a look into this possibly tomorrow.

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -237,24 +241,22 @@ void verifyGivenDockerfile(File dockerFile, BuildImageConfiguration buildConfig,
...
 
-        synchronized (trackArchiver) {
-            MappingTrackArchiver ta = (MappingTrackArchiver) trackArchiver;
-            ta.init(log, assemblyName);
+        synchronized (trackArchivers) {
+            trackArchivers.init(log, assemblyName);
             assembly.setId("tracker");
             assemblyArchiver.createArchive(assembly, assemblyName, "track", source, false, null);
-            return ta.getAssemblyFiles(mojoParams.getSession());
+            return trackArchivers.getAssemblyFiles(mojoParams.getSession(), assemblyName);
```
rohanKanojia, line 259: I think trackArchiver's destFile is no longer being set here after `assemblyArchiver.createArchive(assembly, assemblyName, "track", source, false, null);` call. This seems to be causing NullPointerException in MappingTrackArchiver#getAssemblyFiles() where `getDestFile()` is called. I think explicitly adding destfile should fix this problem:
```suggestion
            File trackArchiverDestFile = assemblyArchiver.createArchive(assembly, assemblyName, "track", source, false, null);
            trackArchivers.get(assemblyName).setDestFile(trackArchiverDestFile);

            return trackArchivers.getAssemblyFiles(mojoParams.getSession(), assemblyName);
```

rohanKanojia:
@zer0keefie : Thanks a lot for your contribution :heart: . This is a really nice feature. I think it would be nice if we had a quickstart for multiple assemblies mode. Could you please add a sample whenever you revisit this? Also, I would appreciate it if you could add an integration test for this so that someone doesn't accidentally break this in the future.

rohanKanojia:
I think there are a few Sonar Code Smells: https://sonarcloud.io/project/issues?id=io.fabric8%3Adocker-maven-plugin&pullRequest=1387&resolved=false&types=CODE_SMELL&q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

Would appreciate if you could address them too.

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
```diff
@@ -639,6 +666,24 @@ public String initAndValidate(Logger log) throws IllegalArgumentException {
             healthCheck.validate();
         }
 
+        List<AssemblyConfiguration> assemblyConfigurations = getAssemblyConfigurations();
+        if (!assemblyConfigurations.isEmpty()) {
+            Map<String, Integer> seenNames = new HashMap<>();
+            for (AssemblyConfiguration config : assemblyConfigurations) {
+                seenNames.compute(config.getName(), (k, v) -> v == null ? 1 : v + 1);
+            }
```
rohanKanojia, line 674: I see you're trying to find assembly names with the same name. Do you think we can use a `Set` here instead? [Set `add()`](https://docs.oracle.com/javase/6/docs/api/java/util/Set.html#add(E)) method returns false if the element already exists in the collection. I haven't tried this myself but this can be simplified like this:
```java
Set<String> assemblyNames = new HashSet<>();
for (AssemblyConfiguration assemblyConfiguration : assemblyConfigurations) {
    boolean wasElementAbsent = assemblyNames.add(assemblyConfiguration.getName());
    if (!wasElementAbsent) {
        log.error("Multiple assemblies use the name %s. Please assign each assembly a unique name.");
        throw new IllegalArgumentException("Assembly names must be unique");
    }
}
```

src/test/java/io/fabric8/maven/docker/config/BuildImageConfigurationTest.java
```diff
@@ -208,5 +209,19 @@ public void compression() {
...
+        BuildImageConfiguration config =
+                new BuildImageConfiguration.Builder().
+                        assemblies(Arrays.asList(assemblyConfigurationOne, assemblyConfigurationTwo)).build();
 
+        try {
+            config.initAndValidate(logger);
+        } catch (IllegalArgumentException expected) {
+            return;
+        }
+        fail("Should have failed.");
```
rohanKanojia, line 225: We can use the expected attribute of the `@Test` annotation to declare `IllegalArgumentException` to be thrown anywhere in the annotated test method like this: `@Test(expected = IllegalArgumentException.class)`

zer0keefie:
Thanks for all the feedback! I've incorporated the suggestions (and merged the latest commit from master), plus added a sample that uses multiple assemblies. I think I got most of the Sonar issues cleaned up, but I'll let the Sonar check tell me what I missed.

The Contributing guide recommended squashing the commits for a PR. I find it easier to keep track of what I changed if the incremental changes are separate commits, but I'll rebase and squash them together once everything looks good.

src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java
zer0keefie: For as long as I've been using Java, I never looked at the return value of Set.add(). This is much nicer.

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
zer0keefie: Huh, I must be missing something that was getting set up in the dependency injection, because I can't figure out how this was working before. It's definitely working with this change, though.

zer0keefie:
I think that covers everything. I've added the integration test, took care of most of the Sonar issues, and squashed all of the commits into a single one.

The two remaining code smells Sonar flagged are removing the deprecated code (for the single assembly at the top level) and replacing some unit tests with a single parameterized test. There didn't seem to be a good way to make those tests parameterized without having to rewrite a decent chunk of the tests or upgrading to JUnit 5.

rohanKanojia:
@zer0keefie : Hi, Sorry for the late reply. Thanks for taking the time to address my comments. It looks really good. I'll revisit your PR this weekend.

it/multi-assembly/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+        <artifactId>dmp-it-parent</artifactId>
+        <version>0.34-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>dmp-multi-assembly</artifactId>
+    <version>0.34-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <description>Docker Maven Plugin Example with Spring Boot with Multiple Layers</description>
```
rohanKanojia, line 17:
```suggestion
    <description>Docker Maven Plugin Integration Test with Spring Boot with Multiple Layers</description>
```

it/multi-assembly/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+                <version>${spring.version}</version>
+            </plugin>
+            <plugin>
+                <groupId>io.fabric8</groupId>
+                <artifactId>docker-maven-plugin</artifactId>
+                <version>0.34-SNAPSHOT</version>
```
rohanKanojia, line 93: I think you should either remove this(this would be picked from parent pom) or set as `$[project.version}`

samples/multi-assembly/pom.xml
```diff
@@ -0,0 +1,145 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rohanKanojia, line 1: Thanks for adding a sample :+1: . Could you please add a README.md file too regarding build instructions? Something like we have in JIB sample: https://github.com/fabric8io/docker-maven-plugin/tree/master/samples/spring-boot-with-jib

src/test/java/io/fabric8/maven/docker/assembly/DockerAssemblyManagerTest.java
```diff
@@ -2,22 +2,13 @@
...
-import io.fabric8.maven.docker.util.Logger;
-import io.fabric8.maven.docker.util.MojoParameters;
-import mockit.Expectations;
-import mockit.Injectable;
-import mockit.Mock;
-import mockit.MockUp;
-import mockit.Tested;
-import mockit.Verifications;
+import io.fabric8.maven.docker.util.*;
+import mockit.*;
```
rohanKanojia, line 11: nit, we usually avoid wildcard imports. I think this might be due to IDE configuration.

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -203,32 +201,47 @@ private void interpolateDockerfile(File dockerFile, BuildDirs params, FixedStrin
...
+    private String firstNonOptionArgument(String[] line) {
+        int i;
+        for (i = 1; i < line.length; i++) {
+            String component = line[i];
+            if (!component.startsWith("--")) {
+                return line[i];
+            }
+        }
+
+        return null;
```
rohanKanojia, line 223: nit, I think this can be simplified like this:
```java
return Arrays.stream(line)
        .skip(1)
        .filter(component -> !component.startsWith("--"))
        .findFirst()
        .orElse(null);
```

rohanKanojia (approved):
Thanks a lot :+1: .

it/multi-assembly/pom.xml
zer0keefie: Good catch. I removed it.

samples/multi-assembly/pom.xml
zer0keefie: You're welcome! I've added a README as requested.

src/test/java/io/fabric8/maven/docker/assembly/DockerAssemblyManagerTest.java
zer0keefie: I always forget to go turn that setting back off... I've replaced them with single imports.

samples/multi-assembly/pom.xml
rohanKanojia: Thanks a lot!

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
zer0keefie: That's much nicer; I'll do that instead.

rohanKanojia:
@zer0keefie : Thanks for addressing my comments. I've already approved your PR :heavy_check_mark: . Let's wait for Roland to review this(he is an expert on dmp); Hopefully it would be merged and available in next release

spmason:
Any update on this issue? I've been keeping an eye on it for a while now as it should help me to avoid pushing large images during dev testing

it/multi-assembly/src/main/java/io/fabric8/maven/sample/springboot/layers/Application.java
```diff
@@ -0,0 +1,29 @@
...
+                LOGGER.info("Hello from Spring Boot!");
+                SpringApplication.exit(applicationContext);
+            }
+        };
+    }
+
+    public static void main(String[] args) {
+        SpringApplication.run(Application.class);
+    }
+}
```
ohbus, line 29: Add an empty line

```suggestion
}

```

src/main/resources/assemblies/dependencies.xml
```diff
@@ -0,0 +1,24 @@
...
+        <fileSet>
+            <includes>
+                <include>classpath</include>
+            </includes>
+            <directory>${project.build.directory}/</directory>
+            <outputDirectory>.</outputDirectory>
+            <fileMode>0644</fileMode>
+        </fileSet>
+    </fileSets>
+</assembly>
```
ohbus, line 24: Add an empty line

```suggestion
</assembly>

```

samples/multi-assembly/src/main/java/io/fabric8/maven/sample/springboot/layers/Application.java
```diff
@@ -0,0 +1,29 @@
...
+                LOGGER.info("Hello from Spring Boot!");
+                SpringApplication.exit(applicationContext);
+            }
+        };
+    }
+
+    public static void main(String[] args) {
+        SpringApplication.run(Application.class);
+    }
+}
```
ohbus, line 29:
```suggestion
}

```

src/main/resources/assemblies/release-dependencies.xml
```diff
@@ -0,0 +1,16 @@
...
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <excludes>
+                <exclude>*:*:*:*-SNAPSHOT</exclude>
+            </excludes>
+            <scope>runtime</scope>
+            <fileMode>0644</fileMode>
+        </dependencySet>
+    </dependencySets>
+</assembly>
```
ohbus, line 16: Add an extra line

```suggestion
</assembly>

```

src/main/resources/assemblies/snapshot-dependencies.xml
```diff
@@ -0,0 +1,16 @@
...
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <includes>
+                <include>*:*:*:*-SNAPSHOT</include>
+            </includes>
+            <scope>runtime</scope>
+            <fileMode>0644</fileMode>
+        </dependencySet>
+    </dependencySets>
+</assembly>
```
ohbus, line 16: Add an extra line

```suggestion
</assembly>

```

src/test/java/io/fabric8/maven/docker/assembly/TrackArchiverCollectionTest.java
```diff
@@ -0,0 +1,77 @@
...
+        assertEquals(1, entries.size());
+        entry = entries.get(0);
+        assertEquals(tempFile2, entry.getSrcFile());
+        assertEquals(destination2, entry.getDestFile());
+
+        // Verify that updating a file in "deps" doesn't pollute "maven"
+        entries = files.getUpdatedEntriesAndRefresh();
+        assertEquals(0, entries.size());
+    }
+}
```
ohbus, line 77: Add an extra line

```suggestion
}

```

ohbus (approved):

zer0keefie:
I swear I have something wrong in my notification settings, so I just noticed there were some comments to address. I pulled in the suggested whitespace fixes.

rohanKanojia:
Hi, Sorry for late reply. I think we should get this in our upcoming release (0.35.0) let's try to get this merged by the end of coming week so that it's available in upcoming release

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
rhuss: Is this fix alread applied ? If not, would be cool to add it before merging.

rhuss (approved):
Thanks a lot for the pull request for this very valuable feature. Really like it.

Although I haven't dived into detail into the PR, the implementation looks straight forward and also thank you for taking care about all the extras around the code itself (samples, documentation, the property handler that is often forgotten ;-)

I suggest to merge this PR asap when the conflicts are resolved and the @rohanKanojia comments have been addressed. @rohanKanojia feel free to merge when you think its the right time for it.

src/main/asciidoc/inc/build/_configuration.adoc
```diff
@@ -8,6 +8,9 @@ of an image configuration. The following configuration options are supported:
 |===
 | Element | Description
 
+| <<config-image-build-assembly, *assemblies*>>
+| Specifies multiple assembly configurations as described in <<build-assembly,Build Assembly>>
+
```
rhuss, line 13: I think the enhanced assembly documentation is missing for the property style configuration. Could you please check and add it if missing there ?

rhuss:
ah yes, and an entry in `docs/changelog.md` would be very cool, too.

rhuss:
fyi, I resolved the conflicts as they seemed to be trivial to resolve.

src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
rohanKanojia: I checked and seems like the fix is applied. I'm resolving this comment.

src/main/asciidoc/inc/build/_configuration.adoc
rohanKanojia: Are you talking about configuring assembly via properties? I'm trying this but somehow this isn't working

```
mvn docker:build -Ddocker.imagePropertyConfiguration=override \
  -Ddocker.from=fabric8/java-centos-openjdk8-jdk:1.5.6 \
  -Ddocker.cmd="java -jar /app/dmp-sample-multi-assembly-0.34-SNAPSHOT.jar" \
  -Ddocker.ports.1=8080 \
  -Ddocker.assemblies.1.descriptorRef=dependencies \
  -Ddocker.assemblies.1.name=deps \
  -Ddocker.assemblies.1.targetDir=/app/lib \
  -Ddocker.assemblies.2.descriptorRef=artifact \
  -Ddocker.assemblies.2.targetDir=/app
```

rohanKanojia:
I want to merge this PR and release 0.35.0, I think we can address the assembly property configuration section later on(I think it's some misconfiguration from my side), XMl configuration seems to be working fine.

PR #1388, opened 2020-10-08
Add a skipPom parameter, skipping a project if packaging is pom

rolaca11:
The usecase for this feature:

There are multiple projects which use the same docker-maven-plugin configuration. A logical way to extract that config to one place is to use the `pluginManagement` tag.

However, this leaves the need to define the plugin in every project wanting to use that configuration.

A ready solution is to define the plugin in the parent pom, but than the build will fail, because the docker build goal cannot find the packaged jar archive.

Here comes this PR, which tells the docker build goal to skip if the currently built project is a pom.

### DIFF
--- a/src/main/asciidoc/inc/_global-configuration.adoc
+++ b/src/main/asciidoc/inc/_global-configuration.adoc
@@ -137,6 +137,9 @@ By default a progress meter is printed out on the console, which is omitted when
 | If set dont push any images even when `{plugin}:push` is called.
 | `docker.skip.push`
 
+| *skipPom*
+| If set to `true` this plugin will skip every projects, where `project.packaging` is set to `pom`. Property: `docker.skip.pom`
+
 | *skipRun*
 | If set dont create and start any containers with `{plugin}:start` or `{plugin}:run`
 | `docker.skip.run`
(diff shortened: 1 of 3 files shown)


src/main/asciidoc/inc/_global-configuration.adoc
```diff
@@ -137,6 +137,9 @@ By default a progress meter is printed out on the console, which is omitted when
 | If set dont push any images even when `{plugin}:push` is called.
 | `docker.skip.push`
 
+| *skipPom*
+| If set to `true` this plugin will skip every projects, where `project.packaging` is set to `true`. Property: `docker.skip.pom`
```
rohanKanojia, line 141: I think there is a typo here:
```suggestion
| If set to `true` this plugin will skip every projects, where `project.packaging` is set to `pom`. Property: `docker.skip.pom`
```

rohanKanojia:
@rolaca11: Thanks a lot for this :+1:

src/main/asciidoc/inc/_global-configuration.adoc
```diff
@@ -137,6 +137,9 @@ By default a progress meter is printed out on the console, which is omitted when
 | If set dont push any images even when `{plugin}:push` is called.
 | `docker.skip.push`
 
+| *skipPom*
+| If set to `true` this plugin will skip every projects, where `project.packaging` is set to `pom`. Property: `docker.skip.pom`
```
mabrarov, line 141: There should be 3 columns in each row:

```
| *skipPom*
| If set to `true` this plugin will skip every projects, where `project.packaging` is set to `pom`.
| `docker.skip.pom`
```

rohanKanojia: @mabrarov: Thanks a lot for noticing, Could you please create a PR to fix this ?

mabrarov: Here is it #1448

PR #1406, opened 2020-11-14
make build reproducible

hboutemy:
see https://maven.apache.org/guides/mini/guide-reproducible-builds.html

on next release, I'll independently check rebuild vs the reference binaries you'll have pushed to Central repository and publish the result on https://github.com/jvm-repo-rebuild/reproducible-central


rohanKanojia (approved):

hboutemy:
thank you @rohanKanojia 
can you associate the milestone, please?

PR #1413, opened 2020-12-03
Fix build on Windows

sabi0:
Fix OS specific filesystem assumptions in the tests


sabi0:
So much for
> We ♥︎♥︎ LOVE ♥︎♥︎ contributions ;-)

rohanKanojia:
@sabi0 : Hi, Sorry for late response.Thanks a lot for your PR :+1: . I think we should add some github action workflow for windows so that this doesn't happen again .

Issue #1168, opened 2019-02-28
Container name pattern ignored when stopping


jakub-bochenski:
### Description
I want to run concurrent builds on a single docker host (for CI).

I started two builds with `-Ddocker.containerNamePattern="$BUILD_TAG-%n-%i"` (e.g. `-Ddocker.containerNamePattern=build-1-%n-%i` and `-Ddocker.containerNamePattern=build-2-%n-%i`).

I expected the builds to ignore each other's containers, only stopping the ones that match their respective build pattern, but instead they stop each other's containers.

This results in errors because containers being stopped mid-use.
```
12:59:54 [ERROR] DOCKER> [acme-all:latest] "acme-all": Container stopped with exit code 137 unexpectedly after 35529 ms while waiting on healthcheck '"CMD-SHELL", "wget --server-response -q -O /dev/null http://localhost:8080/acme/status || exit 1"'
12:59:54 [ERROR] DOCKER> Error occurred during container startup, shutting down...
12:59:54 [ERROR] DOCKER> Unable to remove container [b097fd97dc80] : removal of container b097fd97dc80 is already in progress (Conflict: 409) [removal of container b097fd97dc80 is already in progress (Conflict: 409)]
```
While in the other build log there is
```
12:59:54 [INFO] DOCKER> [acme-all:latest]: Stop and removed container b097fd97dc80 after 0 ms
```

### Info

* d-m-p version : 0.27.2
* Maven version (`mvn -v`) : 3.5
* Sample project: https://github.com/jakub-bochenski/demp-concurrent-stop

jakub-bochenski:
I've created a sample project to demonstrate the issue: https://github.com/jakub-bochenski/demp-concurrent-stop

This is a major blocker for me, so I would appreciate at least some hints on how to work around this.

jakub-bochenski:
Looking at https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/util/ContainerNamingUtil.java#L66 there is no code to differentiate images created with different name patterns. 

Seems like the only thing that matters is GAV and the image used for the container. The name pattern is not taken into account to differentiate containers. This is not what the documentation suggests

> You can combine the placeholders in any combination and will be resolved during docker:start, docker:stop and docker:watch.

Would you accept a PR that would change it, so that only containers started with the same name pattern would be considered?

rhuss:
[...]

PR #1427, opened 2021-01-18
fix stop mojo by taking container name pattern into account

j3t:
Fixes fabric8io/docker-maven-plugin#1168


zbelzer:
This would be very useful for our team. Please consider its review.

ohbus (approved):

src/main/java/io/fabric8/maven/docker/util/ContainerNamingUtil.java
```diff
@@ -57,33 +57,37 @@ public static String formatContainerName(final ImageConfiguration image,
...
-                image.getAlias(),
-                buildTimestamp);
+        Map<String, FormatParameterReplacer.Lookup> lookups = new HashMap<>();
+        lookups.put("a", () -> image.getAlias());
+        lookups.put("n", () -> cleanImageName(image.getName()));
+        lookups.put("t", () -> "\\d{10,}");
+        lookups.put("i", () -> "\\d+");
 
-        return keepOnlyLastIndexedContainer(containers, partiallyApplied);
+        String appliedContainerNamePattern = new FormatParameterReplacer(lookups).replace(containerNamePattern);
```
rhuss, line 86: Is it correct that this changes the current behaviour in a non-backwards compatible way? (i.e. that all containers are stopped now by default, not the latest started)

If so we would need to be more careful to keep the old behaviour (at least for a certain deprecation period) and introduce this new behaviour either by a certain indication in the container name pattern or if this is not possible, with a new flag for `docker:stop`.

rhuss (commented):
In general I believe the use case that this PR is solving and that is described in the reference issue is valid, but I'm concerned about backwards compatibility which must be ensured.

Unfortunately I don't have enough time to dive deeper in this PR technically, but @rohanKanojia if you could give it tech review that would be cool.

rhuss:
btw, I'm also not sure why only the container name with the highest index is used (when the container pattern includes a `%i`. We might remove it, though.

rhuss:
Can we get #1410 in first and then check out how this fit to this PR ? (guess there might be a conflict)

rohanKanojia:
@j3t: Hello, Thanks for updating this PR. Somehow commit history seems to be messed up. Could you please take a look into it? Let's try to get it reviewed and merged soon.

j3t:
Hi @rohanKanojia, is there any chance this gets included in the next release? What is missing? How can I help?

rhuss:
Looks good to me.  Let's get that merged.

rhuss:
@j3t thanks a ton ! Sorry for the delay, I hope we can make a release asap.

rohanKanojia:
@j3t: I'm sorry for the late reply. I was on PTO. Do you need a new patch release with this fix? Or are you okay with waiting for the next minor release?

j3t:
> @j3t: I'm sorry for the late reply. I was on PTO. Do you need a new patch release with this fix? Or are you okay with waiting for the next minor release?

No worries, the next minor release is sufficient I guess.

Issue #1428, opened 2021-01-22
Exception when parsing Timestamp


jbennett2091:
### Description
Exception thrown parsing timestamp information for postgres

### Info
Exception in thread "Thread-6" java.time.format.DateTimeParseException: Text 'Error' could not be parsed at index 0
        at java.base/java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2046)
        at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1948)
        at java.base/java.time.ZonedDateTime.parse(ZonedDateTime.java:598)
        at java.base/java.time.ZonedDateTime.parse(ZonedDateTime.java:583)
        at io.fabric8.maven.docker.util.TimestampFactory.createTimestamp(TimestampFactory.java:46)
        at io.fabric8.maven.docker.access.log.LogRequestor.callLogCallback(LogRequestor.java:200)
        at io.fabric8.maven.docker.access.log.LogRequestor.readStreamFrame(LogRequestor.java:173)
        at io.fabric8.maven.docker.access.log.LogRequestor.parseResponse(LogRequestor.java:186)
        at io.fabric8.maven.docker.access.log.LogRequestor.run(LogRequestor.java:104)

* d-m-p version :  0.34.1
* Maven version (`mvn -v`) :  3.6.3
```

```
* Docker version :
* If it's a bug, how to reproduce :
* If it's a feature request, what is your use case :
* Sample project : *[GitHub Clone URL]*

simontaen:
Same Issue here:

`
 Exception in thread "Thread-7" java.time.format.DateTimeParseException: Text 'Error' could not be parsed at index 0
        at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
        at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
        at java.time.ZonedDateTime.parse(ZonedDateTime.java:597)
        at java.time.ZonedDateTime.parse(ZonedDateTime.java:582)
        at io.fabric8.maven.docker.util.TimestampFactory.createTimestamp(TimestampFactory.java:46)
        at io.fabric8.maven.docker.access.log.LogRequestor.callLogCallback(LogRequestor.java:200)
        at io.fabric8.maven.docker.access.log.LogRequestor.readStreamFrame(LogRequestor.java:173)
[...]

PR #1429, opened 2021-01-22
Issue #1428: exception when parsing timestamp

jbennett2091:
If parsing fails, use default timestamp instead


jbennett2091:
Fixes #1428

jbennett2091:
squashed and signed-off

ohbus (approved):

rhuss:
Thanks for the fix (and sorry for the delay) but I think we should also check, which the extraction of the timestamp fails in the first place. Happy to add a fallback (not sure about using the current timestamp though as there can be a considerable drift between the local time and the server time), but I suspect that the log line that causes the error matched the `LOG_LINE` regexp, but in an undesired way (i.e. the output of the log might have changed). 

Could you check what concrete log line caused the error? Maybe the regexp could be adapt to pick up the concrete value. (btw the "Error" part in the exception provided in the original issue also looks a bit suspicious)

jbennett2091:
Hi Roland-
I made a cursory attempt at the time to determine the log-line that generated the issue, but was never able to pin it down precisely - lots of spam and a messy business process.  I am no longer at this gig, so ability to recheck it at this point is not possible for me.  I do know the fix provided "worked for me" - seemed to turn a hard-exception into code-that-runs.
    On Sunday, March 7, 2021, 04:57:00 AM EST, Roland Huß <notifications@github.com> wrote:  
 
 


Thanks for the fix (and sorry for the delay) but I think we should also check, which the extraction of the timestamp fails in the first place. Happy to add a fallback (not sure about using the current timestamp though as there can be a considerable drift between the local time and the server time), but I suspect that the log line that causes the error matched the LOG_LINE regexp, but in an undesired way (i.e. the output of the log might have changed).

Could you check what concrete log line caused the error? Maybe the regexp could be adapt to pick up the concrete value. (btw the "Error" part in the exception provided in the original issue also looks a bit suspicious)

rhuss:
yeah, no worries. I think we can more or less apply this fix but also log out a warning with the full log line in case we hit this. So hopefully we can fine tune the regexp in retrospective then.

rhuss:
Sorry, there are conflict, not sure how to resolve those (it's very likely JMockit vs. Mockito ...)

PR #1447, opened 2021-03-25
Avoid race condition when creating port bindings

### LINKED ISSUE #1446: Race condition when starting container causes IndexOutOfBoundsException when binding ports
### Description
An intermittent issue can be seen when starting containers. It seems to be a race condition that is exacerbated when running maven with debug logging `-X`. Without the debug logging containers fail to start with  
```
[ERROR] DOCKER> Error occurred during container startup, shutting down...
```

The full stack trace of the exception is:
```
Caused by: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
    at jdk.internal.util.Preconditions.outOfBounds (Preconditions.java:64)
    at jdk.internal.util.Preconditions.outOfBoundsCheckIndex (Preconditions.java:70)
    at jdk.internal.util.Preconditions.checkIndex (Preconditions.java:248)
    at java.util.Objects.checkIndex (Objects.java:372)
    at java.util.ArrayList.get (ArrayList.java:459)
    at com.google.gson.JsonArray.get (JsonArray.java:194)
    at io.fabric8.maven.docker.model.ContainerDetails.createPortBindings (ContainerDetails.java:200)
    at io.fabric8.maven.docker.model.ContainerDetails.getPortBindings (ContainerDetails.java:139)
    at io.fabric8.maven.docker.service.RunService.updateMappedPortsAndAddresses (RunService.java:461)
    at io.fabric8.maven.docker.service.RunService.createAndStartContainer (RunService.java:156)
    at io.fabric8.maven.docker.service.helper.StartContainerExecutor.startContainer (StartContainerExecutor.java:47)
    at io.fabric8.maven.docker.StartMojo.lambda$startImage$0 (StartMojo.java:299)
```

When adding some logging to the maven plugin I can see that during the the `io.fabric8.maven.docker.model.ContainerDetails.createPortBindings` method call the ports will usually have some mappings like this:
```
Ports:{"5432/tcp":[{"HostIp":"0.0.0.0","HostPort":"54290"}],}
```
However sometimes the array of mapped ports will be empty, e.g.
```
Ports:{"5432/tcp":[]}
```

This causes the exception as the array is expected to not be empty and the first element retrieved.

I believe the call to the api in `io.fabric8.maven.docker.service.RunService.updateMappedPortsAndAddresses` is sometimes returning this empty data for the port if called too early.

The solution I would propose would be to retry the API call and the `createPortBindings` activity if an empty array is encountered.

### Info

* docker-maven-plugin version : 0.34.1
* Maven version (`mvn -v`) : 
```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /usr/local/Cellar/maven/3.6.3_1/libexec
Java version: 11.0.10, vendor: AdoptOpenJDK, runtime: /Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "10.14.6", arch: "x86_64", family: "mac"
```
* Docker version : 3.2.2
* If it's a bug, how to reproduce : The issue is intermittent so impossible to consistently reproduce. It appeared to be more easily reproducible when running `mvn verify` with `-X` flag.

ewencluley:
See https://github.com/fabric8io/docker-maven-plugin/issues/1446

Welcome feedback to my approach, to be frank I am not well versed with this code base and may have got some terminology wrong (especially in the unit test). If you can think of a better way to handle this I would love to hear it and happy to change the PR to implement.

On container startup, sometimes the port information retrieved from the
docker API is incomplete with port details being empty arrays, e.g.
Ports:{"2525/tcp":[]}. This causes an IndexOutOfBoundsException to be thrown
as the createPortBindings method assumes there will be at least one entry in
the mapping. In these cases we should retry the API call to get the more
complete port information.

Fixes #1446

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/RunService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/RunService.java
@@ -20,6 +20,7 @@ import static io.fabric8.maven.docker.util.VolumeBindingUtil.resolveRelativeVolu
  */
 
 import java.io.File;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -55,6 +56,7 @@ import io.fabric8.maven.docker.model.Container;
 import io.fabric8.maven.docker.model.ContainerDetails;
 import io.fabric8.maven.docker.model.ExecDetails;
 import io.fabric8.maven.docker.model.Network;
+import io.fabric8.maven.docker.model.PortBindingException;
 import io.fabric8.maven.docker.util.ContainerNamingUtil;
 import io.fabric8.maven.docker.util.EnvUtil;
 import io.fabric8.maven.docker.util.GavLabel;
@@ -62,6 +64,8 @@ import io.fabric8.maven.docker.util.Logger;
 import io.fabric8.maven.docker.util.StartOrderResolver;
 import io.fabric8.maven.docker.wait.WaitTimeoutException;
 import io.fabric8.maven.docker.wait.WaitUtil;
+import net.jodah.failsafe.Failsafe;
+import net.jodah.failsafe.RetryPolicy;
 
 
 /**
@@ -464,12 +468,23 @@ public class RunService {
     }
 
     private void updateMappedPortsAndAddresses(String containerId, PortMapping mappedPorts) throws DockerAccessException {
-        Container container = queryService.getMandatoryContainer(containerId);
-        if (container.isRunning()) {
-            mappedPorts.updateProperties(container.getPortBindings());
-        } else {
-            log.warn("Container %s is not running anymore, can not extract dynamic ports",containerId);
-        }
+        RetryPolicy<Void> retryPolicy = new RetryPolicy<Void>()
+                .withMaxAttempts(10)
+                .withDelay(Duration.ofMillis(250))
+                .handle(PortBindingException.class)
+                .onFailedAttempt(f -> log.debug("Failed to update mapped ports for container %s (attempt %d), retrying", 
+                        containerId, f.getAttemptCount()))
+                .onRetriesExceeded(f -> log.warn("Failed to update mapped ports for container %s after %d retries", 
+                        containerId, f.getAttemptCount()));
+        
+        Failsafe.with(retryPolicy).run(() -> {
+            Container container = queryService.getMandatoryContainer(containerId);
+            if (container.isRunning()) {
+                mappedPorts.updateProperties(container.getPortBindings());
+            } else {
+                log.warn("Container %s is not running anymore, can not extract dynamic ports", containerId);
+            }
+        });
     }
 
     private void shutdown(ContainerTracker.ContainerShutdownDescriptor descriptor, boolean keepContainer, boolean removeVolumes)
--- a/src/main/java/io/fabric8/maven/docker/service/RunService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/RunService.java
@@ -21,6 +21,7 @@ import static io.fabric8.maven.docker.util.VolumeBindingUtil.resolveRelativeVolu
 
 import java.io.File;
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -469,8 +470,8 @@ public class RunService {
 
     private void updateMappedPortsAndAddresses(String containerId, PortMapping mappedPorts) throws DockerAccessException {
         RetryPolicy<Void> retryPolicy = new RetryPolicy<Void>()
-                .withMaxAttempts(10)
-                .withDelay(Duration.ofMillis(250))
+                .withMaxAttempts(20)
+                .withBackoff(10, 100, ChronoUnit.MILLIS)
                 .handle(PortBindingException.class)
                 .onFailedAttempt(f -> log.debug("Failed to update mapped ports for container %s (attempt %d), retrying", 
                         containerId, f.getAttemptCount()))
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -13,6 +13,7 @@
   - Add a skipPom parameter, skipping a project if packaging is pom ([1388](https://github.com/fabric8io/docker-maven-plugin/pull/1388))
   - Add support for config to specify isolation technology for container ([1376](https://github.com/fabric8io/docker-maven-plugin/pull/1376))
   - Add Docker build cache friendly example utilizing Spring Boot Layered JAR and Maven Assembly Plugin ([1412](https://github.com/fabric8io/docker-maven-plugin/pull/1412))
+  - Retry port mapping to avoid race condition where complete port information may not be initially available in Docker Engine 20.10.5 ([1447](https://github.com/fabric8io/docker-maven-plugin/pull/1447))
   
 * **0.34.1** (2020-09-27)
   - Fix NPE with "skipPush" and no build configuration given ([#1381](https://github.com/fabric8io/docker-maven-plugin/issues/1381))
--- a/pom.xml
+++ b/pom.xml
@@ -183,6 +183,12 @@
       <artifactId>plexus-interpolation</artifactId>
       <version>1.24</version>
     </dependency>
+    
+    <dependency>
+      <groupId>net.jodah</groupId>
+      <artifactId>failsafe</artifactId>
+      <version>2.4.0</version>
+    </dependency>
 
 
     <!-- =============================================================================== -->
--- a/src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
+++ b/src/main/java/io/fabric8/maven/docker/model/ContainerDetails.java
@@ -8,6 +8,8 @@ import java.util.Map;
 import java.util.Set;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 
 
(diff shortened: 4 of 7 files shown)


src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -464,11 +467,23 @@ private void startContainer(ImageConfiguration imageConfig, String id, GavLabel
     }
 
     private void updateMappedPortsAndAddresses(String containerId, PortMapping mappedPorts) throws DockerAccessException {
-        Container container = queryService.getMandatoryContainer(containerId);
-        if (container.isRunning()) {
-            mappedPorts.updateProperties(container.getPortBindings());
-        } else {
-            log.warn("Container %s is not running anymore, can not extract dynamic ports",containerId);
+        for (int attempt = 0; attempt < 5; attempt++) {
```
rohanKanojia, line 470: Hello, Do you think we should add `Thread.currentThread().isInterrupted()` in this loop terminating condition here?

ewencluley: Good shout. Added

src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -464,11 +467,23 @@ private void startContainer(ImageConfiguration imageConfig, String id, GavLabel
...
+                }
+                return;
+            } catch (PortBindingException e) {
+                try {
+                    log.warn("Failed to get port bindings for container %s (attempt %d of %d)", containerId, attempt, UPDATE_MAPPED_PORTS_RETRIES);
+                    Thread.sleep(UPDATE_MAPPED_PORTS_RETRY_DELAY_MS);
+                } catch (InterruptedException ex) {
+                    Thread.currentThread().interrupt();
+                }
+            }
```
ewencluley, line 486: Im wondering if I should throw an exception here (possibly changing the retry loop to throw the `PortBindingException` if it was seen on the last attempt)- i.e. if it has not successfully updated the mapped port's properties I guess its best it fails loudly and indicates what has gone wrong instead of just moving on. Having said this I don't understand the process fully enough to determine what will happen if this step fails & it quietly moves on.

src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -71,7 +72,9 @@
  * @since 16/06/15
  */
 public class RunService {
-
+    private static final int UPDATE_MAPPED_PORTS_RETRIES = 5;
+    private static final int UPDATE_MAPPED_PORTS_RETRY_DELAY_MS = 500;
```
git9999999, line 76: i would recommend to go here to 250 and increase the UPDATE_MAPPED_PORTS_RETRIES

git9999999: i checked the logs with this parameters is see only one try per container. i will test it with a lower time
[WARNING] DOCKER> Failed to get port bindings for container 4d2b4ff6615c (attempt 0 of 5)
[WARNING] DOCKER> Failed to get port bindings for container 35efb753ef46 (attempt 0 of 5)

git9999999: private static final int UPDATE_MAPPED_PORTS_RETRIES = 50;
    private static final int UPDATE_MAPPED_PORTS_RETRY_DELAY_MS = 50;
    
    i still see only one attemt
    [WARNING] DOCKER> Failed to get port bindings for container 8c64d1409bef (attempt 0 of 50)

git9999999: private static final int UPDATE_MAPPED_PORTS_RETRIES = 500;
    private static final int UPDATE_MAPPED_PORTS_RETRY_DELAY_MS = 5;
    
    [WARNING] DOCKER> Failed to get port bindings for container 0fbb04f25568 (attempt 0 of 500)
    [WARNING] DOCKER> Failed to get port bindings for container 0fbb04f25568 (attempt 1 of 500)

note that a have a pretty new PC with a lot of power.

so finally i would recommend 10ms

src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -464,11 +467,23 @@ private void startContainer(ImageConfiguration imageConfig, String id, GavLabel
...
+                Container container = queryService.getMandatoryContainer(containerId);
+                if (container.isRunning()) {
+                    mappedPorts.updateProperties(container.getPortBindings());
+                } else {
+                    log.warn("Container %s is not running anymore, can not extract dynamic ports", containerId);
+                }
+                return;
+            } catch (PortBindingException e) {
+                try {
+                    log.warn("Failed to get port bindings for container %s (attempt %d of %d)", containerId, attempt, UPDATE_MAPPED_PORTS_RETRIES);
```
git9999999, line 481: i would change this to debug, but add at the end a error log if this fails.

src/main/java/io/fabric8/maven/docker/service/RunService.java
ewencluley: Yeah, i think the issue is it only prints out when it fails (also it is 0 indexed) so it may fail once and then pass on the 2nd attempt but will not print anything out. Will tweak this.

ewencluley: I think having the delay of 250ms and 10 attempts is reasonable- it really should not make a huge difference to container start times to wait 250ms between attempts, especially as it failing to get the port data from docker is an edge case and race condition. As container start times are usually in the order of seconds or tens of seconds.

src/main/java/io/fabric8/maven/docker/service/RunService.java
ewencluley: yeah, I think thats sensible. Have also added a debug log message when it succeeds too.

src/main/java/io/fabric8/maven/docker/service/RunService.java
git9999999: I would change this
`private static final int UPDATE_MAPPED_PORTS_RETRY_DELAY_MS = 500;`
to
`private static final int UPDATE_MAPPED_PORTS_RETRY_DELAY_MS = 10;`
and also adjust the UPDATE_MAPPED_PORTS_RETRIES to match again.

ewencluley:
I've updated this PR to use the failsafe library for the retries. I think it makes the code clearer than the loop with if conditions that I had. Hope the maintainers of this repo are OK with introducing this dependency.

src/main/java/io/fabric8/maven/docker/service/RunService.java
ewencluley: Im not convinced that hammering the docker api every 10ms for (potentially) up to 500 attempts is the best idea. By setting the delay to 250ms it will wait long enough that the port information should be available after 1 retry. 250ms should not be problematic to add to container start times.

Happy to be told otherwise by the maintainers of this codebase but IMO 250ms and 10 attempts should be sufficient.

git9999999:
As a Polling Framework i recommend to use https://github.com/awaitility/awaitility

her you can specify a **initial delay** and a **polling interval**. here an example

`        await()
                .pollDelay(0, SECONDS)
                .pollInterval(5, SECONDS)
                .atMost(60, SECONDS)
                  .....`

IMO: for me build time is a very precious thing. Alone in my project i start 8 Docker-Images. I think we should put some effort in keep this time as short as possible. as we have seen on my PC, after10ms the container is ready, so 240ms are lost.

ewencluley:
> As a Polling Framework i recommend to use https://github.com/awaitility/awaitility
> 
> her you can specify a **initial delay** and a **polling interval**. here an example
> 
> ` await() .pollDelay(0, SECONDS) .pollInterval(5, SECONDS) .atMost(60, SECONDS) .....`
> 
> IMO: for me build time is a very precious thing. Alone in my project i start 8 Docker-Images. I think we should put some effort in keep this time as short as possible. as we have seen on my PC, after10ms the container is ready, so 240ms are lost.

I have not used Awaitility before but from a cursory look it appears to be less versatile than Failsafe while achieving the same goal. The suggestion to use Awaitility seems like a personal preference rather than having identified any deficiency with using Failsafe. In particular Awaitility appears to have several drawbacks that, while it would be possible to use, in my mind make it less desirable in this instance.

- It does not have (as far as i can tell) a way of specifying a max number of attempts, only a timeout. This is fine for the functionality we want to achieve, however it makes writing tests harder (e.g. with a mock will return state that causes failures n times before returning state that will succeed.) It makes tests less predictable and we can see different results when debugging tests or if tests run faster/slower than expected.
- Using `untilAsserted` (which seems to be the only option for running a method until it does not throw an exception) on ultimate failure it throws a `ConditionTimeoutException` with no reference to the original exception that caused the ultimate failure. While this maybe ok but does obscure the cause of the failure.
- Logging each attempt would need done inside the condition (a runnable passed to the `untilAsserted` method) where there is no information available on the current attempt number. Failsafe provides nice `onAttempt`, `onRetry`, etc. hooks that do provide this information. I think in particular if we intend to reduce to retry interval and increase the number of attempts it will be useful from a debugging standpoint to be able to see the number of attempts done before success/failure.

Since you feel strongly enough about reducing the retry interval I am happy to do that unless there is any objection from the maintainers.

@rohanKanojia it would be good to have some input from maintainers on this PR whether it looks ok.

ewencluley:
> IMO: for me build time is a very precious thing. Alone in my project i start 8 Docker-Images. I think we should put some effort in keep this time as short as possible. as we have seen on my PC, after10ms the container is ready, so 240ms are lost.

Actually come to think of this a potential compromise could be to use a backoff policy so that initially it retries very frequently but if the machine is slower it will backoff to a longer interval. Have added that as a separate commit. This way it will start waiting 10ms between retries but double its interval every attempt up to 100ms.

rohanKanojia:
Hello, I don't have any problem with keeping failsafe :+1: . I'll give it a review again tonight and approve/merge if everything is fine.

rohanKanojia (approved):

rohanKanojia:
Could you please update [doc/changelog.md](https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/changelog.md) regarding this change? I think we can proceed with merge

ewencluley:
@rohanKanojia if you're happy with the changelog can you merge it? I don't have permission to merge I don't think.

ewencluley:
Although, not sure when the 0.35 release i likely to be, would it be worth making this a hotfix of (0.34.2) for this as it seems to affect the docker plugin

Issue #1455, opened 2021-04-19
Using <external> properties in image configuration disables Docker cache during build


brenuart:
Using an external configuration of type "properties" seems to disable the Docker cache.

To illustrate the issue, I took the "multi-assembly" sample available at https://github.com/fabric8io/docker-maven-plugin/tree/master/samples/multi-assembly. After a first execution to warm up the cache, the second `mvn clean package` gives the following output for the docker build stage:
```
[INFO] DOCKER> Step 1/5 : FROM fabric8/java-centos-openjdk8-jdk:1.5.6
[INFO] DOCKER> 
[INFO] DOCKER> ---> 34854f349ada
[INFO] DOCKER> Step 2/5 : EXPOSE 8080
[INFO] DOCKER> 
[INFO] DOCKER> ---> Using cache
[INFO] DOCKER> ---> 9a4ab427263d
[INFO] DOCKER> Step 3/5 : COPY deps /app/lib/
[INFO] DOCKER> 
[INFO] DOCKER> ---> Using cache
[INFO] DOCKER> ---> d5a69097ec70
[INFO] DOCKER> Step 4/5 : COPY maven /app/
[INFO] DOCKER> 
[INFO] DOCKER> ---> 05c4ea19f56d
[INFO] DOCKER> Step 5/5 : CMD java -jar /app/dmp-sample-multi-assembly-0.35.0.jar
[INFO] DOCKER> 
[INFO] DOCKER> ---> Running in 2ea4994a3863
[INFO] DOCKER> Removing intermediate container 2ea4994a3863
[INFO] DOCKER> ---> f9b7fb8d7949
[INFO] DOCKER> Successfully built f9b7fb8d7949
[INFO] DOCKER> Successfully tagged fabric8/dmp-sample-multi-assembly:latest
```
Notice the `Using cache` messages for step 2 & 3 that tell me the docker cache is effectively used.

Then I modified the image configuration slightly and added the following declaration in the <image> configuration:
```
...
<image>
    <name>${docker.user}/dmp-sample-multi-assembly</name>
    <external>
       <type>properties</type>
       <prefix>docker</prefix>
       <mode>skip</mode>
    </external>
...
```
Note that `<mode>` is set to `skip` which according to the documentation should be the same as not specifying any `<external>` at all (car. https://dmp.fabric8.io/#combining-property-config).

Executing `mvn clean package` with this new configuration gives:
```
[...]

PR #1456, opened 2021-04-20
Extract `cacheFrom` from original BuildImageConfiguration as a LIST instead of STRING

brenuart:
See #1455

### DIFF
--- a/src/main/asciidoc/inc/external/_property_configuration.adoc
+++ b/src/main/asciidoc/inc/external/_property_configuration.adoc
@@ -368,7 +368,7 @@ For example:
 <docker.tags.current>latest</docker.tags.current>
 ----
 
-The _idx_ suffixe defines the order of copying. Entries which _idx_ is an integer number are copied first in ascending order by the parsed value of _idx_. Entries which _idx_ is not an integer number are copied second in ascending alphabetical order by _idx_.
+The _idx_ suffix defines the order of copying. Entries which _idx_ is an integer number are copied first in ascending order by the parsed value of _idx_. Entries which _idx_ is not an integer number are copied second in ascending alphabetical order by _idx_.
 
 [[combining-property-config]]
 .Combining property and XML configuration
--- a/src/main/asciidoc/inc/external/_property_configuration.adoc
+++ b/src/main/asciidoc/inc/external/_property_configuration.adoc
@@ -68,16 +68,16 @@ when a `docker.from` or a `docker.fromExt` is set.
 | Don't use Docker's build cache. This can be overwritten by setting a system property `docker.noCache` when running Maven.
 
 | *docker.bind.idx*
-| Sets a list of paths to bind/expose in the container
+| Sets a list of paths to bind/expose in the container. See <<list-properties>>.
 
 | *docker.buildArg.VARIABLE*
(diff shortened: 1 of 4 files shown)


rohanKanojia:
I think you would need to update tests as well

brenuart:
Yeah, I forgot :-(
I have some issues to run them under Eclipse.

rohanKanojia:
Sorry I'm not an eclipse IDE user. I use IntelliJ

brenuart:
Hmmm... The documentation says that `cacheFrom` is "_**List** of image names to use as cache sources_".
When defined inline in the plugin configuration as an XML element, this should give something like:
```
<image>
  <build>
    <cacheFrom>
      <cache>image1:latest</cache>
      <cache>image2:latest</cache>
    </cacheFrom>
   ...
```

When defined as a property, we should probably do something like:
```
docker.cacheFrom.1 = image1:latest
docker.cacheFrom.2 = image2:latest
```
If I understand correctly, the documentation at https://dmp.fabric8.io/#property-configuration should refer use `docker.cacheFrom.idx` to refer to this property. Correct?

If I'm right - this explains why the test fails...

brenuart:
Any feedback about this PR ?

rohanKanojia:
Do you think we need to update documentation too https://dmp.fabric8.io/#property-configuration for cacheFrom?

brenuart:
I have updated the documentation. I also created a new section dedicated to List Properties instead of repeating how they work for every property of this type.

brenuart:
Done! Is everything ok now ?

src/main/asciidoc/inc/external/_property_configuration.adoc
```diff
@@ -344,6 +354,22 @@ configuration are both used for running container and building
...
+[source,xml]
+----
+<docker.ports.1>80<docker.ports.1>
+<docker.ports.2>8080<docker.ports.2>
+
+<docker.tags.jenkins>${BUILD_TIMESTAMP}</docker.tags.jenkins>
+<docker.tags.current>latest</docker.tags.current>
+----
+
+The _idx_ suffixe defines the order of copying. Entries which _idx_ is an integer number are copied first in ascending order by the parsed value of _idx_. Entries which _idx_ is not an integer number are copied second in ascending alphabetical order by _idx_.
```
rohanKanojia, line 371:
```suggestion
The _idx_ suffix defines the order of copying. Entries which _idx_ is an integer number are copied first in ascending order by the parsed value of _idx_. Entries which _idx_ is not an integer number are copied second in ascending alphabetical order by _idx_.
```

rohanKanojia (approved):

rohanKanojia:
@brenuart : Could you please add a line to [doc/changelog.md](https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/changelog.md) regarding this change? Are we changing cacheFrom behavior from old releases? If yes, Shall we add a note about that in doc/changelog?

brenuart:
Done. I added an entry in the `changelog.md` with a warning about the fact that `docker.cacheFrom` alone without an index suffix won't work anymore (as far as it ever worked correctly.

brenuart:
Do you want me to squash and rebase before you merge the PR ?

Issue #1457, opened 2021-04-20
Consider pulling the <cacheFrom> images automatically


brenuart:
It would be nice if the plugin could automatically pull the <cacheFrom> images so they are available locally (possibly according to a configured CachePullPolicy as for the base image).

This feature would be very convenient for those of us that delegate the entire image build process to Maven. Without this feature, one needs to explicitly attempt to _pull_ the cacheFrom images prior executing the maven build - which is not very handy, certainly in CI environments when using generic build/pipeline scripts (the images to cache from are indicated inside the pom and not known from outside).

What do you think ?

PR #1463, opened 2021-04-25
[1457] pull cacheFrom images during build

stefanos-kalantzis:
First PR on this project, so not sure about processes and conventions.

I wasn't sure if this feature should have its own configuration or not. For me it makes sense to always try to pull the `cacheFrom` images in a non-failing way, while adhering to `config.imagePullPolicy`.

---

fixes https://github.com/fabric8io/docker-maven-plugin/issues/1457

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/BuildService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildService.java
@@ -1,22 +1,8 @@
 package io.fabric8.maven.docker.service;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.Serializable;
-import java.nio.file.Files;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.regex.PatternSyntaxException;
-
-import org.apache.maven.plugin.MojoExecutionException;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
-
 import io.fabric8.maven.docker.access.BuildOptions;
 import io.fabric8.maven.docker.access.DockerAccess;
 import io.fabric8.maven.docker.access.DockerAccessException;
@@ -27,13 +13,15 @@ import io.fabric8.maven.docker.config.CleanupMode;
 import io.fabric8.maven.docker.config.ImageConfiguration;
 import io.fabric8.maven.docker.model.ImageArchiveManifest;
 import io.fabric8.maven.docker.model.ImageArchiveManifestEntry;
-import io.fabric8.maven.docker.util.DockerFileUtil;
-import io.fabric8.maven.docker.util.EnvUtil;
-import io.fabric8.maven.docker.util.ImageArchiveUtil;
-import io.fabric8.maven.docker.util.ImageName;
-import io.fabric8.maven.docker.util.Logger;
-import io.fabric8.maven.docker.util.MojoParameters;
-import io.fabric8.maven.docker.util.NamePatternUtil;
+import io.fabric8.maven.docker.util.*;
+import org.apache.maven.plugin.MojoExecutionException;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.file.Files;
+import java.util.*;
+import java.util.regex.PatternSyntaxException;
 
 public class BuildService {
 
@@ -66,6 +54,7 @@ public class BuildService {
 
         if (imagePullManager != null) {
             autoPullBaseImage(imageConfig, imagePullManager, buildContext);
+            autoPullCacheFromImage(imageConfig, imagePullManager, buildContext);
         }
 
         buildImage(imageConfig, buildContext.getMojoParameters(), checkForNocache(imageConfig), checkForSquash(imageConfig), addBuildArgs(buildContext), buildArchiveFile);
@@ -384,6 +373,20 @@ public class BuildService {
         for (String fromImage : fromImages) {
             if (fromImage != null && !DockerAssemblyManager.SCRATCH_IMAGE.equals(fromImage)) {
                 registryService.pullImageWithPolicy(fromImage, imagePullManager, buildContext.getRegistryConfig(), queryService.hasImage(fromImage));
+            }
+        }
+    }
+
+    private void autoPullCacheFromImage(ImageConfiguration imageConfig, ImagePullManager imagePullManager, BuildContext buildContext) throws MojoExecutionException {
+        if (imageConfig.getBuildConfiguration().getCacheFrom() == null) {
+            return;
+        }
+
+        for (String cacheFromImage : imageConfig.getBuildConfiguration().getCacheFrom()) {
+            try {
+                registryService.pullImageWithPolicy(cacheFromImage, imagePullManager, buildContext.getRegistryConfig(), queryService.hasImage(cacheFromImage));
+            } catch (DockerAccessException e) {
+                log.warn("Could not pull cacheFrom image: '%s'. Reason: %s", cacheFromImage , e.getMessage());
             }
         }
     }
--- a/src/main/java/io/fabric8/maven/docker/service/BuildService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildService.java
@@ -13,14 +13,25 @@ import io.fabric8.maven.docker.config.CleanupMode;
 import io.fabric8.maven.docker.config.ImageConfiguration;
 import io.fabric8.maven.docker.model.ImageArchiveManifest;
 import io.fabric8.maven.docker.model.ImageArchiveManifestEntry;
-import io.fabric8.maven.docker.util.*;
+import io.fabric8.maven.docker.util.DockerFileUtil;
+import io.fabric8.maven.docker.util.EnvUtil;
+import io.fabric8.maven.docker.util.ImageArchiveUtil;
+import io.fabric8.maven.docker.util.ImageName;
+import io.fabric8.maven.docker.util.Logger;
+import io.fabric8.maven.docker.util.MojoParameters;
+import io.fabric8.maven.docker.util.NamePatternUtil;
 import org.apache.maven.plugin.MojoExecutionException;
 
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.nio.file.Files;
-import java.util.*;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
 import java.util.regex.PatternSyntaxException;
 
 public class BuildService {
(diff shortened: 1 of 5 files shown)


stefanos-kalantzis:
Also, I tested this manually with docker deamon and works fine, but I have no idea about `jib`. I don't know what it is, or how to test it. I get an error that _"Dockerfile mode is not supported with JIB"_.

rohanKanojia:
No worries, Thanks a lot for your PR. @brenuart: Could you please help in review of this since you requested this feature?

src/main/java/io/fabric8/maven/docker/service/BuildService.java
```diff
@@ -388,6 +377,31 @@ private void autoPullBaseImage(ImageConfiguration imageConfig, ImagePullManager
         }
     }
 
+    private void autoPullCacheFromImage(ImageConfiguration imageConfig, ImagePullManager imagePullManager, BuildContext buildContext) throws DockerAccessException, MojoExecutionException {
+        if (imageConfig.getBuildConfiguration().getCacheFrom() == null) {
```
brenuart, line 381: `cacheFrom` is a List. In most cases the field is indeed `null` when the list is empty (not good practice IMHO), but don't you think you'd better to check for `getCacheFrom().isEmpty()` here as well (for robustness) ?

Same remark for `imageConfig.getBuildConfiguration()`. The method is invoked only when the buildConfig is not null but nothing in the method signature let you know that. Again, just for robustness, I would check if buildConfig!=null in the same if...

src/main/java/io/fabric8/maven/docker/service/BuildService.java
```diff
@@ -388,6 +377,31 @@ private void autoPullBaseImage(ImageConfiguration imageConfig, ImagePullManager
...
+        for (String cacheFromImage : imageConfig.getBuildConfiguration().getCacheFrom()) {
+            if (queryService.hasImage(cacheFromImage)) {
+                return;
+            }
+        }
+
+        for (String cacheFromImage : imageConfig.getBuildConfiguration().getCacheFrom()) {
+            try {
+                registryService.pullImageWithPolicy(cacheFromImage, imagePullManager, buildContext.getRegistryConfig(), false);
+                if (queryService.hasImage(cacheFromImage)) {
```
brenuart, line 394: As you described in the documentation, here you stop pulling images after the first success. Why so? I would rather attempt to pull all of them... In our setup we usually have at least two images to cache from, with obviously different content/layers. These are often the last release and last snapshot. Docker consider them in the order they are configured. Basically, what we do here is trying to first reuse a layer of a released image (happens during a release build) and fall back to the snapshot (for snapshot build)...
I can imagine other scenarios where the user configures multiple cacheFrom image for other reasons. It doesn't harm to pull them all... after all, this is what the user configured ;-)
See what I mean?

brenuart (changes requested):
Happy to see this PR so fast after submitting my feature request...
What is proposed here is similar to what we have done in our "private fork" but cleaner with less hacks ;-)
My biggest "concern" is about not pulling all the configured images. See comments below.

src/main/java/io/fabric8/maven/docker/service/BuildService.java
stefanos-kalantzis: - a check for `getCacheFrom().isEmpty()` is not needed, because it will never go inside the for-loop. The method will just return gracefully.
- The `imageConfig.getBuildConfiguration()` is never checked for nullness in this class already (5 occurences). I decided to skip this check to be consistent with the rest of the class.

src/main/java/io/fabric8/maven/docker/service/BuildService.java
stefanos-kalantzis: Actually the docker docs are very poor on this. There's no official information on how docker handles multiple `--cache-from` arguments (at least I couldn't find any).

My understanding is that docker prefers to do a single-pass over the image, and will use the first applicable cacheFrom image (not multiple).

Here are some other comments/discussions that back my point:
- https://github.com/moby/moby/issues/34715#issuecomment-425933774
- https://github.com/moby/moby/issues/26065#issuecomment-249046559

I found the last comment (`249046559`) referenced even in 2020 discussions, so I believe it's still valid, even though it's from 2016.

---

That being said, there's another scenario that I didn't think of when implementing this:

Let's say the user provides 2 images, `a` and `b`. `a` has no shared layers with the new build, but `b` has.

If we successfully pull the `a` image from the `cacheFrom` list, then `b` will not get pulled.

This is indeed inconsistent with the expectations and behavior of `docker build --cache-from`. Because docker would have found a shared layer with `b` and picked it as the cache-from image.

---

Even so, in my opinion this is quite an extreme case.

Personally I would prefer my CI pipeline to be a bit faster on average (i.e. stop after pulling the first successful image), rather than trying to cover the edge case of multiple cacheFrom images that have drifted apart (which should happen more rarely).

My proposal is to open a new issue that would request for a new config that controls this behavior.
Maybe something like `config.cacheFromPullPolicy`

src/main/java/io/fabric8/maven/docker/service/BuildService.java
brenuart: This is not strictly required indeed. Was just for code robustness in case of future refactoring... No problem for me with your approach.

src/main/java/io/fabric8/maven/docker/service/BuildService.java
brenuart: Docker documentation is indeed not clear on the subject. But experimentation (and practice) has shown that when building a new layer (as a result of a COPY for instance), Docker check the first cache-from image and reuse any matching layer from there. If none can be found, it tries with the second image.

> in my opinion this is quite an extreme case

In our case, a SNAPSHOT image reuses the layers from the previous RELEASE image (our libs are spread across different layers using distinct COPY operations a bit like what SpringBoot 2.3+ is doing). We currently have 4 layers:
- one with the release jars (Spring, SpringBoot, etc)
- a second with our "company" jars (likely to change more frequently than external dependencies included in the first layer)
- a third with SNAPSHOT dependencies
- and a fourth with the application code and resources

This means that as long as nothing except the application code changes, the new SNAPSHOT reuses the layers with libraries (Spring and SpringBoot for instance) from the RELEASE image. If project makes use of snapshot dependencies (which is often the case), then it is likely to reuse layers 1&2 from the last released image (tagged with `latest`).

In short, our configuration is therefore to cacheFrom:
- <project>:latest
- <project>:<current version>

As you notice, <current version> is something like `1.0.0-SNAPSHOT` while the project is SNAPSHOT and becomes `1.0.0` during a "release" build. With this strategy, the build always tries to reuse from both the last released image and the last snapshot (from which one a layer will be reused depends on the actual changes made in the project since then). For a "release" build, `1.0.0` does not exist yet (obviously) and the only source of cache becomes the last released image.

I don't see this scenario as an _extreme case_ ;-) Or may be we missed something and came up with an over complicated solution?

brenuart: Oh.. I forgot to mention that the cacheFrom is configured in the Maven POM itself. We could have passed this information from the CI (Jenkins) itself and use different values depending on the type of build being performed. But in our case the CI doesn't know the registry name of the images being build (and is not even aware that the Maven build will produce any).

stefanos-kalantzis: Docker will **_attempt_** to find a matching cacheFrom image from the given list, but once has found a single layer, then it will lock onto that image, and not try any of the other images anymore.

As I linked before, [this is a comment](https://github.com/moby/moby/issues/26065#issuecomment-249046559) from the [author](https://github.com/moby/moby/pull/26839/commits) of `cacheFrom` inside docker:
> Specifying multiple --cache-from images is bit problematic. If both images match there is no way(without doing multiple passes) to figure out what image to use. So we pick the first one(let user control the priority) but that may not be the longest chain we could have matched in the end. If we allow matching against one image for some commands and later switch to a different image that had a longer chain we risk in leaking some information between images as we only validate history and layers for cache. Currently I left it so that if we get a match we only use this target image for rest of the commands.

_(the code inside `cache.go` has not been modified by anyone ever since, so this docker behavior is still true :) )_ 

This PR is already an improvement over `master`. So we could merge it and then it will become available for many users :)
Then feel free to create follow-up issues / PRs if required.

As I implied before, I'm not against it, but I don't think it should be the default behavior. There should be a new config flag to control this.

brenuart: I see what you mean.

Let me just add that by stopping after the first successful pull, you imply that this is the one Docker should use. When they say "_we pick the first one_", my experience shows that they actually "_pick the first with a match_" - which may be the second (am I wrong on this?). By stopping after the first pull success, we don't give docker a chance to have a match on the second and take the risk to break the caching feature completely.

Personally, I would expect the following configuration:
```
<cacheFrom>image1</cacheFrom>
<cacheFrom>image2</cacheFrom>
```
to be functionally equivalent to:
```
docker pull image1
docker pull image2
docker build --cache-from image1 --cache-from image2 ...
```
without making any assumption about what Docker will actually be doing with the two images (which is unclear as per the various links you mentioned).

stefanos-kalantzis: Yes that's correct. That was my attempt to explain above with the `a` and `b` images example :) 

So, ok, I see that this will not get merged until I change this... :sweat: 

Even though I still believe that if you take the average of all builds over time, it will be slower like this; I'm willing to back down because I prefer to see this merged rather than not.

stefanos-kalantzis:
Updated PR to behave as @brenuart suggested. It will now pull all images provided by the user.

stefanos-kalantzis:
@rohanKanojia  @brenuart could someone approve this PR then? :)

brenuart:
+1 for me but don't have approval permissions...

src/main/java/io/fabric8/maven/docker/service/BuildService.java
```diff
@@ -27,13 +13,15 @@
...
-import io.fabric8.maven.docker.util.MojoParameters;
-import io.fabric8.maven.docker.util.NamePatternUtil;
+import io.fabric8.maven.docker.util.*;
+import org.apache.maven.plugin.MojoExecutionException;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.file.Files;
+import java.util.*;
```
rohanKanojia, line 23: nit, Could you please avoid using wildcard imports?

rohanKanojia (approved):

rohanKanojia:
@stefanos-kalantzis: Thanks for the PR. Could you please resolve conflicts whenever you revisit this? I tried updating your branch but I don't have access to push to your fork. I think we can proceed with merge after that

src/main/java/io/fabric8/maven/docker/service/BuildService.java
Kidlike: done :+1: (sorry I re-installed IntelliJ and forgot to change this setting)

stefanos-kalantzis:
@rohanKanojia I fixed the conflict, but now it looks like there's something wrong with sonarcloud credentials on circleci :(

rohanKanojia:
I had removed the credentials from circleci for sonar. Don't worry about it.

PR #1469, opened 2021-05-05
If docker.image.tag is set to a placeholder, we should replace it.

### LINKED ISSUE #1468: A placeholder in docker.image.tag isn't replaced by the final result when used during docker:build
### Description

If the property `docker.image.tag` is set to `%l` and used in the image name, `docker:build` will fail to replace `%l` in the image name.

### Info

*  See https://github.com/henri-tremblay/docker-bug to reproduce
*  Just do `mvn docker:build`

The cause is that FormatParameterReplacer will do a lookup. In the lookup we get the `docker.image.tag` property but return directly instead of trying to replace the `%l`. Mysteriously, we only do that for `docker.image.tag` and not `docker.image.repo` or anything else.

See the demo project above to reproduce but the layout is:

```xml
  <properties>
    <docker.image.tag>%l</docker.image.tag>
    <docker.image.name>%a</docker.image.name>
    <docker.image.repo>${docker.image.name}</docker.image.repo>
  </properties>
```

```xml
      <plugin>
        <groupId>io.fabric8</groupId>
        <version>0.35.0</version>
        <configuration>
          <images>
            <image>
              <name>${docker.image.repo}:${docker.image.tag}</name>
              <build>
                <from>adoptopenjdk:11-jre-hotspot</from>
              </build>
            </image>
          </images>
        </configuration>
      </plugin>
```

henri-tremblay:
Fix #1468 
This occurs when docker.image.name is constructed with `${docker.image.repo}:${docker.image.tag}`.

This is a fix to #1468

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -4,6 +4,7 @@
   - CI builds with Maven Wrapper to ensure that Maven Wrapper files and configuration are correct ([1450](https://github.com/fabric8io/docker-maven-plugin/issues/1450))
   - Using <external> properties in image configuration disables Docker cache during build ([1455](https://github.com/fabric8io/docker-maven-plugin/issues/1455))
   - Update documentation to clearly state that `docker.cacheFrom.idx` is a _list_ property and should always be used with a `idx` suffix. With this change, `docker.cacheFrom` (without _idx_) is not considered anymore.
+  - A placeholder in docker.image.tag isn't replaced by the final result when used during docker:build ([1468](https://github.com/fabric8io/docker-maven-plugin/issues/1468))
   
 * **0.35.0** (2021-04-04)
   - Building 'spring-boot-with-jib' sample fails with NoSuchMethodError ([1384](https://github.com/fabric8io/docker-maven-plugin/issues/1384))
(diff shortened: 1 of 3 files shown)


henri-tremblay:
**Note:** It is undocumented that `docker.image.tag` can override `%l`. So if I do `mvn docker:build -Ddocker.image.name=%a:%l -Ddocker.image.tag=nightly` I will end up with `myapp:nightly`

rohanKanojia:
@henri-tremblay: Thanks a lot for the PR. Could you please add a line to doc/changelog.md regarding this change? 

> Note: It is undocumented that docker.image.tag can override %l. So if I do mvn docker:build -Ddocker.image.name=%a:%l -Ddocker.image.tag=nightly I will end up with myapp:nightly

Shall we document this behavior in this PR or in separate PR?

henri-tremblay:
Done for the changelog. You should really generate the change log from the issues in the release. I'm doing something similar here: https://github.com/easymock/easymock/blob/master/generate-changelog.sh.

I will let you update the documentation in another PR. From my reading of ImageNameFormatter, two parameters are special. docker.image.user and docker.image.tag. Both can be passed in parameter to override the %g and %v %t %l.

rohanKanojia (approved):

PR #1472, opened 2021-05-11
Add an option to dump the logs to an output file

andreaTP:
This option is designed after:
https://maven.apache.org/plugins/maven-dependency-plugin/list-mojo.html#outputFile

Basically, it redirects the output logs to a file that can be processed by something else (another Mojo, external tools, etc.)

For context:
I'm working on Maven plugin that needs(in this case) to extract the pushed image `sha`.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -268,7 +268,12 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
                         access = dockerAccessFactory.createDockerAccess(dockerAccessContext);
                     }
                     ServiceHub serviceHub = serviceHubFactory.createServiceHub(project, session, access, log, logSpecFactory);
-                    executeInternal(serviceHub);
+                    try {
+                        executeInternal(serviceHub);
+                    } catch (Exception ex) {
+                        ex.printStackTrace();
+                        throw ex;
+                    }
                 } catch (IOException | ExecException exp) {
                     logException(exp);
                     throw new MojoExecutionException(log.errorMessage(exp.getMessage()), exp);
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -268,12 +268,7 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
                         access = dockerAccessFactory.createDockerAccess(dockerAccessContext);
                     }
                     ServiceHub serviceHub = serviceHubFactory.createServiceHub(project, session, access, log, logSpecFactory);
-                    try {
-                        executeInternal(serviceHub);
-                    } catch (Exception ex) {
-                        ex.printStackTrace();
-                        throw ex;
-                    }
+                    executeInternal(serviceHub);
                 } catch (IOException | ExecException exp) {
                     logException(exp);
                     throw new MojoExecutionException(log.errorMessage(exp.getMessage()), exp);
--- a/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
@@ -103,6 +103,10 @@ public abstract class AbstractDockerMojo extends AbstractMojo implements Context
     @Component
     protected DockerAccessFactory dockerAccessFactory;
 
+    // Redirect the plugin output to a file
+    @Parameter(property = "outputFile")
+    private String outputFile;
+
     @Parameter(property = "docker.autoPull")
     protected String autoPull;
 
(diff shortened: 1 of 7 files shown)


rohanKanojia:
There seem to be some test failures. Could you please address them?

andreaTP:
Hey @rohanKanojia good to see you here 🙂  !
I should have fixed the tests now.

andreaTP:
Sonar is failing with:

> Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.9.0.2155:sonar (default-cli) on project docker-maven-plugin: You're not authorized to run analysis. Please contact the project administrator.

probably because the PR is from a fork, not sure how to fix it, any guidance?

andreaTP:
@rohanKanojia have you had the chance to review this PR? Anything I can do to give more context or any additional explanation?
Sorry for asking, but I'm building quite some code/builds on top of this:
https://github.com/lightbend/cloudflow/pull/1049/files

The plugin itself is amazing and I really enjoy it!
Thanks a lot in advance!

rohanKanojia:
Sorry for delay, I'll review your PR this weekend.

andreaTP:
Truly appreciated @rohanKanojia !
Yes, I would need a release of the plugin, I'm building on top of it:

https://github.com/lightbend/cloudflow/pull/1049/files#diff-5c8dd2c16e0716fc062f33c1e549d561efc917f75fb9f52268cb7059d7358609R36

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -257,7 +268,12 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                         access = dockerAccessFactory.createDockerAccess(dockerAccessContext);
                     }
                     ServiceHub serviceHub = serviceHubFactory.createServiceHub(project, session, access, log, logSpecFactory);
-                    executeInternal(serviceHub);
+                    try {
+                        executeInternal(serviceHub);
+                    } catch (Exception ex) {
```
rohanKanojia, line 273: Is this additional try-catch block necessary?

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -59,44 +62,78 @@ public AnsiLogger(Log log, boolean useColor, String verbose) {
...
     }
 
     /** {@inheritDoc} */
     public void error(String message, Object ... params) {
-        log.error(colored(message, COLOR_ERROR, true, params));
+        if (outputFile != null && log.isErrorEnabled()) {
+            pw.println(format(message, params));
+        } else {
+            log.error(colored(message, COLOR_ERROR, true, params));
+        }
```
rohanKanojia, line 136: Do you think maybe we should move this if-else block to a method? Maybe something like this:
```java
private void logOrPrintToFile(Predicate<Log> logPredicate, Consumer<Log> logConsumer, String message, Object ... params) {
    if (outputFile != null && logPredicate.test(log)) {
        pw.println(format(message, params));
    } else {
        logConsumer.accept(log);
    }
}
```

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
andreaTP: removed

src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
andreaTP: I have still to practice some more functional Java 👍 🙂 
I like it! thanks for the suggestion!

andreaTP:
@rohanKanojia ready for next review round 🙂

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
```diff
@@ -238,7 +242,14 @@
     public void execute() throws MojoExecutionException, MojoFailureException {
         if (!skip) {
             boolean ansiRestore = Ansi.isEnabled();
-            log = new AnsiLogger(getLog(), useColorForLogging(), verbose, !settings.getInteractiveMode(), getLogPrefix());
+            File output = null;
+            if (outputFile != null) {
+                output = new File(outputFile);
+                if (output.exists()) {
+                    output.delete();
+                }
```
rohanKanojia, line 250: I think we might not need to check whether outputFile exists and delete it. I think the default behavior of PrintWriter is to overwrite the file. What do you think?

rohanKanojia:
Could you please add a line to doc/changelog regarding this feature as well? It looks good to be merged :+1:

andreaTP:
@rohanKanojia added an entry to the changelog and rebased on latest master, thanks for bearing with me!

src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java
rohanKanojia: ohk, Thanks for clarification

rohanKanojia (approved):

andreaTP:
Thanks! Do you have any ETA for the release?

rohanKanojia:
I will try to cut a release this weekend. Could you please test against master SNAPSHOT build to check everything is working as expected with your project?

andreaTP:
I can confirm that everything works as expected in my project thanks again 👍

PR #1488, opened 2021-08-24
inconsistent names of the `entryPoint` configuration parameter

energister:
maven 3.6.3 error:
> Unable to parse configuration of mojo io.fabric8:docker-maven-plugin:0.37.0:start for parameter entryPoint: Cannot find 'entryPoint' in class io.fabric8.maven.docker.config.RunImageConfiguration

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -4,6 +4,9 @@
   - Allow replacement in tags. Added a new replacement `%T` which always adds a timestamp. ([#1491](https://github.com/fabric8io/docker-maven-plugin/pull/1491))
   - Only push the `latest` tag if no other tags where specified in docker mode. This can break your build, if you rely on the automatic `latest` tag. ([#1496](https://github.com/fabric8io/docker-maven-plugin/pull/1496))
   - Only push the `latest` tag if no other tags where specified in jib mode. This can break your build, if you rely on the automatic `latest` tag. ([#1498](https://github.com/fabric8io/docker-maven-plugin/pull/1498))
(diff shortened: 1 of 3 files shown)


rohanKanojia:
I guess this is due to entrypoint field present in both BuildImageConfiguration `<build>` and RunImageConfiguration `<run>`. It's named differently at both places:

BuildImageConfiguration:
https://github.com/fabric8io/docker-maven-plugin/blob/7d23d73321442877166d6407913ab7d451d3b30c/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java#L148

RunImageConfiguration:
https://github.com/fabric8io/docker-maven-plugin/blob/7d23d73321442877166d6407913ab7d451d3b30c/src/main/java/io/fabric8/maven/docker/config/RunImageConfiguration.java#L52

energister:
So should I rename the property in `RunImageConfiguration` instead of changing documentation to make settings names consistent?

rhuss:
+1 for making the naming of the fields consistent in the code and name it `entryPoint` but also keep the `entrypoint` but deprecate it (like we did here https://github.com/fabric8io/docker-maven-plugin/blob/7d23d73321442877166d6407913ab7d451d3b30c/src/main/java/io/fabric8/maven/docker/config/BuildImageConfiguration.java#L119-L124 for `noCache`, a similar issue)

For the document I would only use `entryPoint` though, but keeping `entrypoint` helps for backwards compatibility. Don't forget to mention it in the changelog, too.

thanks !

rohanKanojia (approved):

PR #1513, opened 2021-12-20
Cleanup dangling images

rkhmelichek:
There is a `cleanup` build configuration option that will by default 'cleanup dangling (untagged) images after each build'.
I would propose extending that to also cleanup dangling images as a result of:
tagging
auto-pulling of base images
auto-pulling of cache-from images

We use docker-maven-plugin on a CI server and auto-removing dangling images keeps the server from filling up with old image versions.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,7 @@
 # ChangeLog
 
 * **0.39-SNAPSHOT** :
+  - Cleanup dangling images as a result of image tagging, auto-pulling a base image, or auto-pulling a cacheFrom image ([#1513](https://github.com/fabric8io/docker-maven-plugin/pull/1513)) @rkhmelichek
 
 * **0.38.1** (2021-12-18):
   - Update to jnr-unixsocket 0.38.14 to solve UnsatisfiedLinkError on Apple M1 ([#1257](https://github.com/fabric8io/docker-maven-plugin/pull/1507)) @henningn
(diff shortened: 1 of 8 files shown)


rohanKanojia:
Thanks for your PR :+1: . I think we should update docs as well regarding this

rohanKanojia:
@rkhmelichek : polite ping, still working on this patch?

rkhmelichek:
Hi @rohanKanojia,

Apologies for the delay, I'm still working on refining this.
One issue I've been thinking about is how in the documentation, the `cleanup` config option is described as `Cleanup dangling (untagged) images after each build (including any containers created from them).`
Before my changes, only images built by the docker-maven-plugin were being cleaned up.
And so it is safe to clean the previously built image, it will indeed become dangling.

However, the way the dangling image is cleaned up (`docker.removeImage(oldImageId, true);`) does not guarantee that the image is dangling (e.g. no other tags) - because the second arg is setting `force=true`.
Note: the `force` argument is described as `Remove the image even if it is being used by stopped containers or has other tags`

So, in the case of my changes, we're cleaning up in additional scenarios, like in case an updated base image was pulled.
It's possible that we might delete an image that has other tags.
I'm thinking when cleaning up dangling images in those cases, we should be more conservative by passing `force=false` to Docker. And additionally querying the old image for any tags - if it is still tagged, then we skip the remove step, since it's technically not dangling.

What do you think about this?

rkhmelichek:
I pushed the changes for the more conservative approach, to make sure only truly dangling images are cleaned up as a result of tagging, or auto-pulling base, or cacheFrom images.
Also updated the documentation.

rohanKanojia:
I think you're right. Thanks for looking into this.

rohanKanojia (approved):

PR #1522, opened 2022-01-27
Fix logic bug to remove dangling images

rkhmelichek:
Hi @rohanKanojia,
As a follow up to #1513, I've been testing a private build of the changes on our CI machines, and discovered a logic bug.
The condition to remove the image only if there are no other associated tags/repos should be reversed.

Additionally I'm looking to address one other scenario of dangling images.
That happens when pushing tags.
For example, the following code could result in a dangling image that will never be cleaned up, so I'm thinking of extending the logic to clean it up here as well.
```java
        if (alreadyHasImage) {
            log.warn("Target image '%s' already exists. Tagging of '%s' will replace existing image",
                targetImage, fullName);
        }
```

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,8 +1,9 @@
 # ChangeLog
 
 * **0.39-SNAPSHOT** :
-  - `skipPom` is ignored by "push" goal ([1482](https://github.com/fabric8io/docker-maven-plugin/issues/1482))
+  - `skipPom` is ignored by "push" goal ([1482](https://github.com/fabric8io/docker-maven-plugin/issues/1482)) @rohanKanojia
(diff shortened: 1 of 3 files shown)


rohanKanojia:
hmm makes sense. Could you please add a test for this too?

rkhmelichek:
hi @rohanKanojia, sure I'll work on adding some test cases too.

rohanKanojia:
@rkhmelichek : I plan to cut 0.39.0 probably today or tomorrow. Do you think we should include this PR in that?

rohanKanojia (approved):

rkhmelichek:
@rohanKanojia, yes thanks for merging!
I will open a new pull request for handling dangling images as a result of pushing tags (and work on tests too).

PR #1529, opened 2022-02-23
[#1528] determineFinalArgValue respect default value if key exists but value is null

### LINKED ISSUE #1528: Dockerfile ARG default throws NPE if ARG is null
### Description
It seems like if you're building a Dockerfile with `ARG` and defining a default like `ARG TEST=test` and pass `<TEST></TEST>` to the plugin, this will fail with NullpointerException in https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/util/DockerFileUtil.java#L268

### Info

* docker-maven-plugin version : 0.39.0

* pom.xml configuration
```
			<plugin>
				<groupId>io.fabric8</groupId>
				<version>0.39.0</version>

				<configuration>
					<images>
						<image>
							<name>example-image</name>
							<build>
								<contextDir>example</contextDir>
								<args>
									<TEST></TEST>
								</args>
							</build>
						</image>
					</images>
				</configuration>
			</plugin>

```

* Example Dockerfile:
```
FROM alpine
ARG TEST=something
```

The example is obviously not what is happening in the end (with that you could just ask me to remove the `arg` if I'm not setting it, but rather the problem that I have another variable in there which potentially is not set)

twendelmuth:
Fix #1528

See #1528 - will return the default value when `argMap.getValue() == null`

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,7 @@
 # ChangeLog
 
 * **0.40-SNAPSHOT** :
+  - determineFinalArgValue respect default value if key exists but value is null ([1528](https://github.com/fabric8io/docker-maven-plugin/issues/1528)) @twendelmuth
 
 * **0.39.0** (2022-02-06):
   - `skipPom` is ignored by "push" goal ([1482](https://github.com/fabric8io/docker-maven-plugin/issues/1482)) @rohanKanojia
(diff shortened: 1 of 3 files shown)


rohanKanojia (approved):

rohanKanojia:
@twendelmuth : Thanks for your PR :+1: . Would you be needing a patch release with this bugfix?

twendelmuth:
@rohanKanojia I'd prefer it - unless you're planning to release 0.40.0 anyway soon 👍

rohanKanojia:
@twendelmuth : I've cut 0.39.1 today https://repo1.maven.org/maven2/io/fabric8/docker-maven-plugin/0.39.1/ . Would appreciate if you could try it out whenever you get time.

PR #1531, opened 2022-02-27
Support for docker run --sysctl parameters

### LINKED ISSUE #1530: Support for docker run --sysctl parameters
I need to configure the net.ipv4.tcp_keepalive_time sysctl parameter, but it doesn't look like this is currently supported?

https://docs.docker.com/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime

jpraet:
Fixes #1530


rohanKanojia:
Thanks for your PR :+1: . Looks good.

Would be nice to add some doc about this new parameter.

rohanKanojia (approved):

PR #1533, opened 2022-03-01
Prevent concurrent access to secDispatcher during password decryption

joserebelo:
secDispatcher is not thread-safe, as it keeps some internal state when decrypting the password. This would make multi-thread maven builds sometimes.

With this, I believe that we could probably even annotate the mojos as threadSafe - at least I haven't found any further issues so far - but I also don't have a very deep knowledge of the code base.


rhuss (approved):
Thanks! Looks good to me.

PR #1552, opened 2022-05-05
Add builderName configuration for buildx multi-architecture

### LINKED ISSUE #1551: Multi arch support: x509: certificate signed by unknown authority
### Description
With https://github.com/fabric8io/docker-maven-plugin/pull/1544 multi-arch support was introduced to the docker-maven-plugin. Today I tried to build a simple image for different architectures (arm64 and amd64). My base image is from an internal registry which uses a selfsigned certificate. Due to this the plugin is unable to lookup the base image with following error:

```
[INFO] DOCKER> #4 [linux/arm64 internal] load metadata for privateregistry/internal/.../springboot-jdk11:latest
[INFO] DOCKER> #4 ERROR: failed to do request: Head "https://privateregistry/v2/internal/.../springboot-jdk11/manifests/latest": x509: certificate signed by unknown authority
```
I found this tutorial how to setup the buildx builder to use/mount own ca-certificate into the build container:
https://github.com/docker/buildx/blob/master/docs/guides/custom-registry-config.md

However it seems that the docker-maven-plugin uses always a temporary builder 
`--builder dmp_privateregistry_my-artifact-name`.

What would be the best solution the solve the certificate issue? 

- Provide a possibility to override the builder? (builder-config is local which requires everyone to config this by himself which is probably not a good idea)
- Provide a possibility to configure a builder which will be used by the plugin?
- Are there other ways to solve this issue which I'm not aware yet?

chonton:
Closes #1551


Chr3is:
@chonton being able to pass the name of an existing builder is the first step. 
Maybe we should provide the possibility to create and configure a new builder like this in the future:
```
docker buildx create  \
  --name mybuilder \
  --driver docker-container \
  --config /etc/buildkitd.toml
```
with this we would be able to be independent from the workstation where the project (and docker-plugin) is executed because we can deliver the builder-configuration within the project (and other config stuff like certificates).

I think at the moment the build would break if the project is checked out and executed on a machine where the builder with the specific name is not available.

chonton:
Added builder config file per suggestion.

rohanKanojia (approved):

PR #1577, opened 2022-07-01
When using buildx, do not force build of native platform

### LINKED ISSUE #1572: Native platform image also forcely built with buildx config [native arch forcely added]
### Description
When we used buildx config and set non-native platform only, like
`                                                                <buildx>
                                                                        <platforms>
                                                                        </platforms>
                                                                </buildx>`
for as example linux/arm64 (M1 Pro)

Plugin will as second step also build image for the native platform due to
https://github.com/fabric8io/docker-maven-plugin/blob/f42c906e5641e0c35bab747f2344948c5da22fce/src/main/java/io/fabric8/maven/docker/service/BuildXService.java#L102
and will fail in our case as PARENT image not availbale for linux/arm64 (openjdk17-alpine)
[INFO] DOCKER> error: failed to solve: openjdk:17-jdk-alpine: no match for platform in manifest sha256:4b6abae565492dbe9e7a894137c966a7485154238902f2f25e9dbd9784383d81: not found

In general we understand rationality for such behavoiur, but it's hard-coded, which make it very hard to use on M1 pro only for package phase for CI/CD.
Can we make it configurable (enabled by default, but with opportunity to disable)

As dirty workaround - we modify Dockerfile, and forcely set FROM --platform=linux/amd64 - but this is not the fancy fix, and destroy all idea of multiarch builds

### Info

* docker-maven-plugin version : 0.40.1
* Maven version (`mvn -v`) : 3.8.6
```

```
* Docker version : 20.10.16
* If it's a bug, how to reproduce : any buildx config when only non-native platform listed

chonton:
closes #1572

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
@@ -98,8 +98,15 @@ public class BuildXService {
     }
 
     private void buildMultiPlatform(List<String> buildX, String builderName, BuildDirs buildDirs, ImageConfiguration imageConfig, File buildArchive) throws MojoExecutionException {
-        buildX(buildX, builderName, buildDirs, imageConfig, imageConfig.getBuildConfiguration().getBuildX().getPlatforms(), buildArchive, null);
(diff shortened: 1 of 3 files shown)


chonton:
@rohanKanojia please review

rohanKanojia (approved):

PR #1579, opened 2022-07-02
Remove buildx cache.  Do not delete builder instances after goal.  Us…

### LINKED ISSUE #1576: Confused about buildx behavior
Context: I am migrating from [spotify/dockerfile-maven](https://github.com/spotify/dockerfile-maven) to this plugin. The recent introduction for buildx support made us tackle this long overdue migration.

Use case: build an image from `Dockerfile` in `package` phase, tag it with two tags, push image with one of the tags to internal Docker registry in `deploy` phase.

_If there's a better way to ask for community support please let me know. It's not ideal to create an issue, I know._

### Description

The way I interpreted the documentation the following configuration should implement my use case.

```xml
<properties>
    <docker.image-name>hello/multiarch</docker.image-name>
    <docker.docker-file>${project.basedir}/Dockerfile</docker.docker-file>
...
</properties>

<build>
    <plugins>
        <plugin>
            <groupId>io.fabric8</groupId>
            <version>0.40.1</version>
            <!-- [package] build Docker image with two tags -->
            <!--           * master                                            -->
            <!--           * $project.version                             -->
            <!-- [deploy] publish the image with the $project.version tag -->
            <configuration>
                <images>
                    <image>
                        <name>${docker.image-name}:${project.version}</name>
                        <registry>docker.io</registry>
                        <build>
                            <dockerFile>${docker.docker-file}</dockerFile>
                            <args>
                                <DOCKER_IMAGE_VERSION>${project.version}</DOCKER_IMAGE_VERSION>
                            </args>
                            <!-- tags *in addition* to potential tag included in the <image><name> -->
                            <tags>
                                <tag>master</tag>
                            </tags>
                            <buildx>
                                <platforms>
                                    <platform>linux/arm64</platform>
                                </platforms>
                            </buildx>
                        </build>
                    </image>
                </images>
            </configuration>
            <executions>
                <execution>
                    <id>build-image</id>
                    <goals>
                        <goal>build</goal>
                    </goals>
                    <phase>package</phase>
                </execution>
                <execution>
                    <id>push-image-version</id>
                    <goals>
                        <goal>push</goal>
                    </goals>
                    <phase>deploy</phase>
                    <configuration>
                        <!-- even though we built an image with two tags, only push one of them -->
                        <filter>${docker.image-name}:${project.version}</filter>
                    </configuration>
                </execution>
            </executions>
        </plugin>
...
    </plugins>
</build>
```

### Issues

Issue 0: the Docker push fails but there are separate issues file for this already e.g. #1574.

Issue 1: since I defined two executions, Docker builds the whole image _twice_. Hence, the Maven `push` goal includes the `build` goal. The effective command is something like `docker ... buildx build ... --push ...`. In fact, the only difference between the two executions is the inclusion of `--push` in the later. They're identical otherwise.

Issue 2: the layers of the parent image are pulled down for every Docker build. I expected them to be cached by default between builds (unless they change of course).

What are the remedies for issues 1 & 2?

### Info

```
* docker-maven-plugin version : 0.40.1
* Maven version (`mvn -v`) : 3.8.5
```

```
* Docker version :
* If it's a bug, how to reproduce : 20.10.16
```

chonton:
…e builder instance to cache artifacts

closes #1576


chonton:
@rohanKanojia, please review
@marcelstoer, this is proposal 1 mentioned in issue #1576. Does this branch solve the caching problem you experienced?

marcelstoer:
Thanks, the proposal is sane. I checked out your branch and `mvn install`ed it. 

The first `mvn deploy` in my multi-arch test project was fine apart from the "insufficient_scope: authorization failed" failure at the end which was my bad. However

- Based on the changes in this PR I was expecting to see a `maven` buildx builder to survive the Maven build. `docker buildx ls` says this be not the case.
- Subsequent `mvn deploy` (re-)create the `maven` buildx builder but fail shortly after with `[INFO] DOCKER> error: failed to solve: failed to read dockerfile: open /tmp/buildkit-mount1275820182/Dockerfile: no such file or directory`. I don't know how to fix it as I don't see where this path is coming from.

chonton:
A builder is nothing more than a set of configurations stored in config.json in the config directory.  In order to list the builder using the cli, you must set `--config` to the same value that dmp uses (the dockerStateDir).  By default, this directory is under the project target directory and will be deleted with `mvn clean`.  Make sure the target directory is not deleted or set dockerStateDir to another location.

derkaiserreich:
I also ran this version to build our platform and can confirm, that the cached versions from the build goal are now used in the push goal. 

However wouldn't it in general make more sense to push the images directly, instead of trying to build them again in the push goal, even if the cached images are now used?

marcelstoer:
> wouldn't it in general make more sense to push the images directly

You get this behavior if you only have a single `<execution>` that uses the `push` goal. Building twice happens if you use the build-test-push approach (three executions).

chonton:
Currently, you will need both goals.  This is consistent with the behavior of non-buildx builds: build, test, push.  We will consider a single build/push goal if there is high demand.  How much time do you find the separate goals add to your build?  Is this a significant hit?

derkaiserreich:
Now with the caching being enabled, the impact of the separate goals isn't that significant for our CI pipeline. For local development we can work around this and build the images without buildx, so for our usecase I don't see the immediate demand for a single goal.

However I can see other projects having their build time growing significantly when building multiple images for multiple platforms, so I can see the benefits of just one single, combined goal. 

I really appreciate the work you're doing here, so thank you for that!

rohanKanojia (approved):

PR #1585, opened 2022-07-21
Fix #1512: Plugin doesn't abort building an image in case Podman is u…

### LINKED ISSUE #1512: Plugin doesn't abort building an image in case Podman is used and the Dockerfile can't be processed
### Description

Short:

Podman's response for POST /build lacks of header "Content-Type: application/json".

Long:

When triggering a build, docker-maven-plugin sends a POST /build request and checks the response.
The response is 200 OK despite the build result, so docker-maven-plugin checks for errors in the response's content.
It does this by checking the headers for "Content-Type: application/json". If given, it checks if the response contains an "error" object. If present, docker-maven-plugin raises an exception to stop the Maven build.

The code in HcChunkedResponseHandlerWrapper is:

```
    public Object handleResponse(HttpResponse response) throws IOException {
        try (InputStream stream = response.getEntity().getContent()) {
            // Parse text as json
            if (isJson(response)) {
                EntityStreamReaderUtil.processJsonStream(handler, stream);
            }
        }
        return null;
    }

    private static boolean isJson(HttpResponse response) {
        return Stream.of(response.getAllHeaders())
                .filter(h -> h.getName().equalsIgnoreCase("Content-Type"))
                .anyMatch(h -> h.getValue().toLowerCase().startsWith("application/json"));
    }
```

And the handler BuildJsonResponseHandler is:

```
    public void process(JsonObject json) throws DockerAccessException {
        if (json.has("error")) {
            String msg = json.get("error").getAsString();
            String detailMsg = "";
            ...
            throw new DockerAccessException(...);
        ...
```

In case Podman is used, no "Content-Type: application/json" is given, nor any "Content-Type" at all. docker-maven-plugin won't parse the response-content for errors, and won't raise an exception to stop the Maven build.

A workaround is needed.

### Versions

Maven:

    $ mvn -v
    Apache Maven 3.6.3
    Maven home: /usr/share/maven
    Java version: 1.8.0_312, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
    OS name: "linux", version: "5.11.0-43-generic", arch: "amd64", family: "unix"

Plugin:

    <plugin>
        <groupId>io.fabric8</groupId>
        <version>0.38.1</version>
    </plugin>

Podman:

podman version 3.4.2

jh-cd:
Fix #1512

In the previous version of this file the following if() was as follows. The methode isJson() checks by header (not by body):
```
     if(isJson(response)) {
          EntityStreamReaderUtil.processJsonStream(handler, stream);
     }
```

In case the Podman daemon is used, the POST /build response is JSON and the HTTP status code is 200
as expected, but there is no header "Content-Type = application/json" nor "Content-Type" at all. Seen in
Podman 3.4.2. But the docker-maven-plugin relies on that JSON-body. In BuildJsonResponseHandler.process():
```
     if (json.has("error")) {
    ...
    throw new DockerAccessException();
```
 If no error is detected, the Maven-build goes on despite there was a problem building the
image!
The following if() first checks for the application/json Content-Type. If no Content-Type is set,
 it tries to detect if the body is JSON. If so, the handler is called.


manusa:
Relates to:
 - #1404 
 - https://github.com/eclipse/jkube/pull/451

manusa (approved):
Logic seems fine.

Maybe you'd want to introduce a complete test case with JSON response that verifies the scenario described by the issue.

Issue #1598, opened 2022-09-05
Docker pulling progress only shown after pull has completed and not in real-time like previous versions


causalnet:
### Description

Since version 0.40.2, when pulling an image the progress bar is only shown after the pull has completed.  For large downloads / slow connections, this lack of progress display can be an issue as the user won't know whether the download has frozen or if Docker is doing work.  This didn't used to be the case with earlier versions of the plugin where the download progress would be displayed in real time from Docker.

e.g. 
```
Pulling from library/postgres
#####
```

It looks like this change (https://github.com/fabric8io/docker-maven-plugin/commit/d8befdbcb989ee48afa3e3f8805c9c8d6d5aa463) is what caused the real-time progress break.  In `HcChunkedResponseHandlerWrapper.handleResponse()` the entire body of the response from Docker is read into a `ByteArrayInputStream` now instead of live-streamed.  So for a response that is sending back progress of a pull, this method now only returns after the pull has finished.  Previously the response was live-streamed into `EntityStreamReaderUtil.processJsonStream()`.  Looks like this was done because the body is now scanned whether it is actually JSON, where previously the headers were fully relied on.

Maybe it would be possible to still live-stream instead of buffering if the headers show the response is JSON, and only buffer the response if there is no content type header?  This would allow live-progress to still work for most users.

### Info

* docker-maven-plugin version : 0.40.2
* Maven version (`mvn -v`) : 3.8.6
* Docker version : 18.09.0, also tested with 20.10.17
* If it's a bug, how to reproduce : do anything that pulls an image, e.g. docker:start when the image is not yet pulled

PR #1599, opened 2022-09-05
Avoid buffering and stream the response if the headers indicate it is JSON

causalnet:
Fix https://github.com/fabric8io/docker-maven-plugin/issues/1598

Instead of always buffering the response, stream the response to the handler if the headers indicate it is definitely JSON.  Buffer like before if there are no content-type headers and the body needs to be checked.

This will allow pull progress to be displayed in real time for all situations where there is a JSON content type, which would be for the majority of cases.


rohanKanojia:
@causalnet : Thanks for your PR :+1: . Is it possible to add a test to validate that fix is working as expected?

causalnet:
@rohanKanojia there are some tests that cover existing functionality already in `HcChunkedResponseHandlerWrapperTest` which covers functionality of JSON and non-JSON headers and bodies.

As for testing whether the result is streamed or buffered, I'm trying to think of a good way to do this.  I could change `EntityStreamReaderUtil.processJsonStream()` to be an instance method instead of a static, then use Mockito to spy calls and detect whether the stream it was passed was the original entity's stream or a new buffered `ByteArrayInputStream`.  Would you be OK with me changing `EntityStreamReaderUtil` to do this?

rohanKanojia:
Umm, can we use mockStatic instead?

causalnet:
Didn't know about mockStatic, that worked great.

PR #1360, opened 2020-07-06
fixed issue: #1359

alan-czajkowski:
fixes #1359

PR #1613, opened 2022-10-20
Add flag for convert tcp to http

nodece:
https://github.com/fabric8io/docker-maven-plugin/pull/1360 is a good idea, but this idea affects too many people, so this PR can help people switch to the HTTP protocol from the TCP protocol, the default to HTTPS protocol.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
+++ b/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
@@ -43,7 +43,13 @@ public class EnvUtil {
 
     // Convert docker host URL to an HTTP(s) URL
     public static String convertTcpToHttpUrl(String connect) {
-        String protocol = connect.contains(":" + DOCKER_HTTP_PORT) ? "http:" : "https:";
+        String toHTTP = System.getenv("DOCKER_TCP_TO_HTTP");
+        String protocol;
+        if ("1".equals(toHTTP) || "true".equals(toHTTP)) {
+            protocol = "http:";
+        } else {
+            protocol = connect.contains(":" + DOCKER_HTTP_PORT) ? "http:" : "https:";
+        }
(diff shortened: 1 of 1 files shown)


src/main/java/io/fabric8/maven/docker/util/EnvUtil.java
```diff
@@ -43,7 +43,13 @@ private EnvUtil() {}
 
     // Convert docker host URL to an HTTP(s) URL
     public static String convertTcpToHttpUrl(String connect) {
-        String protocol = connect.contains(":" + DOCKER_HTTP_PORT) ? "http:" : "https:";
+        String toHTTP = System.getenv("DOCKER_TCP_TO_HTTP");
```
rohanKanojia, line 46: Is this `DOCKER_TCP_TO_HTTP` a docker-maven-plugin configuration option or is it already defined in docker?

rohanKanojia: If it's former, we need to update our documentation as well.

nodece: `DOCKER_TCP_TO_HTTP` belongs to the docker-maven-plugin.

nodece: > If it's the former, we need to update our documentation as well.

Good catch!

rohanKanojia: @trevorgetty : Is it possible for someone from your team to pick this up? You can cherry-pick this commit and create a new PR if possible.

rohanKanojia: @trevorgetty : That would be really appreciated! I think we can start by creating a new PR out of this and working on top of it.

PR #1288, opened 2019-11-09
Fix #1030: docker:build does not support 'network' option

rohanKanojia:
Fix #1030

PR #1636, opened 2023-01-29
Allow overriding build network from properties

ktulinger:
The configuration of
```
<images>
  <image>
      <name>xx/%a:%l</name>
      <alias>xxxx</alias>
      <build>
          <dockerFile>${project.basedir}/Dockerfile</dockerFile>
          <network>host</network>
      </build>

      <external>
          <type>properties</type>
          <prefix>docker</prefix>
          <mode>override</mode>
      </external>
  </image>
</images>
```

simply does not propagate the network option further. When omitting "external" it works as expected. The network field is not resolved when external configuration resolution takes place.

Network field was introduced in https://github.com/fabric8io/docker-maven-plugin/pull/1288

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,7 @@
 # ChangeLog
 * **0.41-SNAPSHOT** :
   - multi-arch build should use provided repository ([1597](https://github.com/fabric8io/docker-maven-plugin/issues/1597)) @merikan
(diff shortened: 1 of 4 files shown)


ktulinger:
I think I filled the documentation too, please check once you have time

rohanKanojia:
@ktulinger : Thanks,  could you please add a line to changelog and a test if possible?

rohanKanojia (approved):

PR #1659, opened 2023-03-20
#1658 Fallback to default docker driver to speed-up build time in buildx mode - useBuildxOnlyOnPush param

### LINKED ISSUE #1658: Fallback to default docker driver to speed-up build time in buildx mode
### Description

While using buildx mode on M1 Mac, I've noticed that "Sending tarball" phase takes extremely long (tens of seconds), some good explanation is here - https://github.com/docker/buildx/issues/107#issuecomment-1307929258.
This makes the plugin unusable in some cases.

...
As far as I see, the plugin builds the image matching host architecture in the build phase, skipping other architectures and using buildx to load the image to local registry (`--load` flag).
It looks like buildx (more specifically `docker-container` driver)  is not optimized for local loading.

My proposal is always to use legacy non-buildx way in build phase and use buildx only on push phase even if `<buildx>` is present in maven configuration . WDYT?

### Info

* docker-maven-plugin version: `0.41.0`
* Maven version (`mvn -v`) : `3.8.6`
* Docker version: `23.0.1`

tfactor2:
Resolves #1658 by adding the `useBuildxOnlyOnPush` parameter,
 
```xml
<configuration>
  <images>
    <image>
      <name>example-image</name>
      <build>
        <buildx>
          <useBuildxOnlyOnPush>true</useBuildxOnlyOnPush>
          <platforms>
            ....
```
Adjusted the docu, see PR.

### DIFF
--- a/src/main/asciidoc/inc/build/_buildx.adoc
+++ b/src/main/asciidoc/inc/build/_buildx.adoc
@@ -4,6 +4,7 @@
 Buildx is enabled when there is a non-empty `<platform>` element inside the  `<buildx>` configuration.  Only the native platform
 is built and saved in the local image cache by the `build` goal.  All specified platforms are built and pushed into the remote
 repository by the `push` goal. This behavior is to prevent non-native images from tainting the local image cache.
+Alternatively, `useBuildxOnlyOnPush` can be set. This will force the plugin to use buildx only on push making local builds faster.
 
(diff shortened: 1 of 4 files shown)


chonton:
As mentioned in my comment on #1658, you can accomplish this using profiles and configuration.

tfactor2:
Thanks @chonton, the desired configuration is described in Examples at https://dmp.fabric8.io/#build-buildx as well.

PR #1704, opened 2023-08-21
Add support setting driverOpts for buildx

Hellspam:
I recently had a use case where I wanted to change the builder image name to use an image from our local Artifactory instance.

To do that with the builder, I needed to add support for driverOpts to set  `--driver-opt image=MY_IMAGE`. Tested locally, seems to work fine.

Regarding testing - I did not see any tests currently verifying the buildx create.

https://docs.docker.com/engine/reference/commandline/buildx_create/#driver-opt


rohanKanojia:
I think you should be able to add an integration test project in `it/` folder with maven executions.

rohanKanojia:
@Hellspam : polite ping, are you still working on this?

rohanKanojia (approved):

PR #1707, opened 2023-09-26
Add new build parameter: "defaultExclusion"

thuva9872:
This commit introduces a new build parameter, "defaultExclusion," within the build properties of the plugin. The current behavior of the plugin is to build a Docker archive with default exclusion rules, which exclude hidden files. However, the Docker CLI includes hidden files by default.

With the addition of the "defaultExclusion" parameter, users can now control whether hidden files should be excluded during the Docker archive build process. This parameter provides flexibility and allows users to align the plugin's behavior with their specific requirements.

Fixes: https://github.com/fabric8io/docker-maven-plugin/issues/1708

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
+++ b/src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
@@ -176,7 +176,7 @@ public class DockerAssemblyManager {
                         // Interpolated Dockerfile is already added as it was created into the output directory when
                         // using dir dir mode
                         excludeDockerfile(fileSet, dockerFile);
-
+                        fileSet.setUsingDefaultExcludes(buildConfig.useDefaultExcludes());
                         // If the content is added as archive, then we need to add the Dockerfile from the builddir
                         // directly to docker.tar (as the output builddir is not picked up in archive mode)
                         if (isArchive(assemblyConfigurations)) {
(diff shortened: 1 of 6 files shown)


src/main/java/io/fabric8/maven/docker/assembly/DockerAssemblyManager.java
```diff
@@ -176,7 +176,7 @@ public TarArchiver customize(TarArchiver archiver) throws IOException {
                         // Interpolated Dockerfile is already added as it was created into the output directory when
                         // using dir dir mode
                         excludeDockerfile(fileSet, dockerFile);
-
+                        fileSet.setUsingDefaultExcludes(buildConfig.defaultExclusion());
```
rhuss, line 179: We might consider to set this always to "false", as IMO this is the sensible behaviour (and usually it also doesn't harm except that it might increase the performance slightly).

rhuss:
Thanks for the PR! We might even want to fix it without config changes, please see the discussion on the issue.

rhuss:
Thanks a lot, looks good to me. One last little thing to add to the documentation, then we can merge it.

thuva9872:
> Thanks a lot, looks good to me. One last little thing to add to the documentation, then we can merge it.

I already added the new configuration in the build configuration documentation. Do I need to add anything else?

rhuss:
Please add the configuration property to `src/main/asciidoc/inc/external/_property_configuration.adoc`, too (mind the alphabetical order of the properties, too)

rohanKanojia (approved):

Issue #1713, opened 2023-10-16
Volume binding bug in windows


shi00:
I think volume binding in windows has a bug.
Consider this snippet:
```
<volumes>
  <bind>
        <volume>d:\a\b:c:\test</volume>
   </bind>
</volumes>
```
I am using Docker Desktop to run a Windows Container on Windows 10. When starting the container, I hope to perform a mount operation, but unfortunately, the operation failed.
When I checked the log information I found the following error:
`**I/O Error [Unable to create container for [builder-base-windows:1.0.0] : {"message":"invalid volume specification: '/d/a/b:c:/test'"} (Internal Server Error: 500)]**`

It is correct for me to start the container through the command line. The startup command is as follows:
**docker run -it -v D:\a\b:c:\test  builder-base-windows:1.0.0 cmd.exe**

Looking at the code it seems that in class VolumeBindingUtil
public static String resolveRelativeVolumeBinding(File baseDir, String bindingString)
Maybe there is something wrong with this code.

docker-maven-plugin version : 0.43.4
Maven version (`mvn -v`) :
Apache Maven 3.8.3 (ff8e977a158738155dc465c6a97ffaf31982d739)
Maven home: D:\apache-maven
Java version: 21, vendor: Azul Systems, Inc., runtime: D:\zulu-21
Default locale: zh_CN, platform encoding: UTF-8
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"

bergerst:
@shi00 Can you test if my fix in #1720 works?

1. Clone my fork and switch to the fix-windows-volume-mounts branch
2. Run `mvn install -DskipTests -Dsonar.skip=true`
3. Change the docker-maven-plugin version in your own project's pom.xml to 0.44-SNAPSHOT
4. Trigger the plugin execution

bergerst:
Since I didn't get a response, I set up a Github workflow to test it and the binding works: https://github.com/bergerst/docker-plugin-windows-showcase/tree/windows-container

You can only bind directories into Windows-based containers according to https://github.com/moby/moby/issues/30555.

bergerst:
@rohanKanojia Can be closed now

shi00:
Hi guy, sorry for the late reply. I have completed the test using the fork you provided, and the result is ok. Thank you for your help.

rohanKanojia:
@shi00 : Thanks for your feedback. I'll cut a new release within a few weeks.

PR #1720, opened 2023-11-07
Fix Windows volume bindings for Rancher, Podman and Windows-based containers

### LINKED ISSUE #1719: Windows volume bindings do not work correctly with Rancher Desktop or Podman
### Description
The docker-maven-plugin currently turns all Windows file paths (C:\Users\example/test.txt) into /c/Users/example/test.txt in [EnvUtils.fixupPath](https://github.com/fabric8io/docker-maven-plugin/blob/f228f644e6ec945f9bdea037980db16a9bab28f7/src/main/java/io/fabric8/maven/docker/util/EnvUtil.java#L347).

This leads to problems when using Rancher Desktop or Podman instead of Docker Desktop. Since both of these tools only run in the WSL, the correct path in the WSL image would be /mnt/c/Users/example/test.txt . The /mnt/ part is currently missing.

However, both Rancher Desktop and Podman can already correct Windows paths into WSL paths if you call the Docker API with Windows file paths. So for example, `docker run -d --name nginx -v C:\Users\example\test.txt:/tmp/test.txt nginx:latest` works.

Rancher Desktop and Podman have different behavior right now with the latest docker-maven-plugin release:
- Rancher Desktop creates an empty directory under /c/Users/example/test.txt
- Podman fails with `make cli opts(): making volume mountpoint for volume /c/Users/example/test.txt: mkdir /c: permission denied`

This bug is somewhat related to #1713 

### Info

* docker-maven-plugin version : 0.43.4
* Maven version (`mvn -v`) :
```
Apache Maven 3.9.1 (2e178502fcdbffc201671fb2537d0cb4b4cc58f8)
Maven home: C:\Development\maven
Java version: 11.0.14.1, vendor: Eclipse Adoptium, runtime: C:\Program Files\Eclipse Adoptium\jdk-11.0.14.1+1
Default locale: en_US, platform encoding: Cp1252
```
* Docker version : Rancher Desktop 1.8.1, Podman 4.7.2
  1. Install and start Podman on Windows using this guide: https://github.com/containers/podman/blob/main/docs/tutorials/podman-for-windows.md
  2. If `podman pull nginx` does not work, you probably have to fix the DNS resolution in the WSL image with this fix: https://gist.github.com/ThePlenkov/6ecf2a43e2b3898e8cd4986d277b5ecf
  3. Clone the sample project
  4. Run `mvn verify`, it should fail with the `permission denied` error
* Sample project : *[GitHub Clone URL](https://github.com/bergerst/docker-plugin-windows-showcase.git)*

bergerst:
Fixes #1719 #1713

After my fix, the sample provided in #1719 works.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
+++ b/src/main/java/io/fabric8/maven/docker/access/ContainerCreateConfig.java
@@ -15,7 +15,6 @@ import java.util.Properties;
 import java.util.Set;
 
 import io.fabric8.maven.docker.config.Arguments;
-import io.fabric8.maven.docker.util.EnvUtil;
 import io.fabric8.maven.docker.util.JsonFactory;
 
 public class ContainerCreateConfig {
@@ -166,14 +165,14 @@ public class ContainerCreateConfig {
     }
 
     private String extractContainerPath(String volume) {
-        String path  = EnvUtil.fixupPath(volume);
-        if (path.contains(":")) {
-            String[] parts = path.split(":");
+        if (volume.contains(":")) {
+            // only split when : is not followed by \ (Windows)
+            String[] parts = volume.split(":(?!\\\\)");
             if (parts.length > 1) {
                 return parts[1];
             }
         }
-        return path;
+        return volume;
     }
 
     private void addEnvironment(Properties envProps) {
--- a/src/main/java/io/fabric8/maven/docker/access/ContainerHostConfig.java
+++ b/src/main/java/io/fabric8/maven/docker/access/ContainerHostConfig.java
@@ -25,7 +25,8 @@ public class ContainerHostConfig {
             JsonArray binds = new JsonArray();
 
             for (String volume : bind) {
-                if (volume.contains(":")) {
+                // only add binding if : is not followed by \ (Windows)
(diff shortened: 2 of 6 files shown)


bergerst:
@rohanKanojia I've fixed the test failures and tested the changes both on Linux- and Windows-based containers.

Can you please trigger the Github workflows again and review my changes?

bergerst:
Test coverage should be better now after the 2nd commit. I can squash my commits if you want, but I thought reviewing will be easier if the commits are split.

I converted the existing for-loop tests into ParameterizedTests since it's easier to read. The ulimit tests in ContainerHostConfigTest were broken since they only read `data[1]`, `data[2]` and so on instead of `data[i+1]` and `data[i+2]`

bergerst:
Sorry, now all tests should pass again.

rhuss:
Thanks! This looks good to me. Could you add some notes to https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/changelog.md for this bug ? Otherwise it looks good for me to merg.

rohanKanojia (approved):

rohanKanojia:
I'll add an entry to the changelog while preparing for the next release.

PR #1721, opened 2023-11-09
Fix Buildx build with Dockerfile outside of Docker build context

mruzicka:
Fixes a Buildx build failure when an external Dockerfile outside of the Docker build context directory is specified.
The Dockerfile (be it an external one or the generated one) is always stored at the root of the build archive created by  `DockerAssemblyManager.createDockerTarArchive()`.
Before this patch the code which constructs the full path of the Dockerfile to pass to Buildx didn't take that into account.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,6 +2,7 @@
 * **0.44-SNAPSHOT**:
   - Add new option "useDefaultExclusion" for build configuration to handle exclusion of hidden files ([1708](https://github.com/fabric8io/docker-maven-plugin/issues/1708))
   - The <noCache> option is now propagated down to the buildx command, if it is set in the <build> section. ([1717](https://github.com/fabric8io/docker-maven-plugin/pull/1717))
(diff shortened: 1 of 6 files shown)


rohanKanojia:
@mruzicka : Thanks for your pr, Is it possible to write some test to validate this fix?

mruzicka:
@rohanKanojia, I've added an integration test here: https://github.com/fabric8io/docker-maven-plugin/tree/30dc304119cae19eb493d651dc7957074a19699b/it/buildx-dockerfile_and_contextdir

rhuss:
Thanks! Test looks good, and the fix as well. Could you please add an entry to https://github.com/fabric8io/docker-maven-plugin/blob/master/doc/changelog.md for the fix, too ? I think we are good to merge then.

mruzicka:
Hi, @rhuss, I've added the changelog entry.

rohanKanojia (approved):

rohanKanojia:
@mruzicka : Thanks a lot!

mruzicka:
Thank you!

PR #1727, opened 2023-11-22
Refacto(src/main/java/io/fabric8/maven/docker/util/ImageName.java)r:Removed unorganized getters and orgnaized by Lombok getters

mdxabu:
I removed some unorganized getters and made them organized and efficient using Lombok ```@Getter```.

Signed-off-by: Mohamed Abdullah <abdullahfakrudeen2020@email.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -312,6 +312,14 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <version>1.18.30</version>
+      <scope>provided</scope>
+    </dependency>
+
+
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks for the PR, but please no Lombok, only for replacing getters. I'm not a big fan of adding an extra dependency only for a little syntactic sugar. The decision of introducing Lombok or not has a wider scope than just sneaking it in via a PR. If you feel inclined, feel free to open an issue to discuss, but please be aware that we could live without Lombok quite nicely over all those years, so would you need to have really good arguments.

I'm closing this PR, but please feel free to elaborate on what you mean by "unorganized" and what value that would bring to the project. But thanks again, really appreciate that you care about code quality.

PR #1730, opened 2023-11-30
feat: Adding support for `--cacheFrom` and `--cacheTo` buildX arguments

### LINKED ISSUE #1621: buildx cache-from / cache-to
Hello,

I am wondering how to set 'cache-from' and 'cache-to' when using buildx to build multi-platform images.

I have a maven project in which several Docker images are being built and some of them depend on each other. Unfortunately, when a depending Docker image is being built, Docker tries to download the base Docker image (as defined by 'FROM' in the Dockerfile) instead of reusing the already built one. This obviously fails.

I have already tested if using 'dockerStateDir' helps, but it did not work either.

Because of this I would like to use the 'cache-from' and 'cache-to' option of buildx but do not know how to do this with the docker-maven-plugin (and if it is possible at all).

Does anybody have an idea how to solve this issue?

### Info

* docker-maven-plugin version : 0.40.2
* Maven version (`mvn -v`) : 3.8.1
* Docker version : 20.10.21

orange-buffalo:
Closes #1621

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
@@ -42,6 +42,18 @@ public class BuildXConfiguration implements Serializable {
     @Parameter
     private String nodeName;
 
+    /**
+     * Value to be passed to {@code --cache-from} option of BuildX build.
+     */
+    @Parameter
+    private String cacheFrom;
+
+    /**
+     * Value to be passed to {@code --cache-to} option of BuildX build.
+     */
+    @Parameter
+    private String cacheTo;
+
     public String getBuilderName() {
         return builderName;
     }
@@ -56,6 +68,14 @@ public class BuildXConfiguration implements Serializable {
 
     public String getNodeName() {
         return nodeName;
+    }
+
+    public String getCacheFrom() {
+        return cacheFrom;
+    }
+
+    public String getCacheTo() {
+        return cacheTo;
     }
 
     public boolean isBuildX() {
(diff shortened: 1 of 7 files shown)


src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
```diff
@@ -42,6 +42,18 @@ public class BuildXConfiguration implements Serializable {
     @Parameter
     private String nodeName;
 
+    /**
+     * Value to be passed to {@code --cache-from} option of BuildX build.
+     */
+    @Parameter
+    private String cacheFrom;
```
rohanKanojia, line 49: Is format of `cacheFrom` used in buildx different from one used in legacy docker builds? I see  we already have `cacheFrom` configuration option in `build` > `cacheFrom`. If they both have similar format maybe we can see the existing configuration option.

rohanKanojia:
Thanks a lot for your PR! I just have a minor comment otherwise it looks good.

src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
orange-buffalo: I am not an expert in this area, but from [docs](https://docs.docker.com/engine/reference/commandline/build/#cache-from) of `docker build --cache-from` I understand that the argument can only be a registry image: "flag pointing to an image in the registry". 

While for `docker buildx build --cache-from` there are [multiple different options](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from): "Supported types are registry, local, gha and s3". 

So the BuildX flag covers the values from the legacy builder, and more. 

From my point of view as a plugin user, `cacheFrom` and `cacheTo` should go together in case of BuildX usage. We cannot add `cacheTo` to `build` section as legacy Docker build does not support this. Having to define `cacheFrom` in `build`, and `cacheTo` in `build / buildX` is confusing from user standpoint. For this reason, I would suggest to have both `cacheFrom` and `cacheTo` in `buildX`, even though there is some overlap with `build [cacheFrom]`.

PR #1731, opened 2023-12-06
Add support for using maven/system properties for an ARG used as the FROM image in a dockerfile

sdumitriu:
Scenario: build using a Dockerfile, I'm trying to configure the base image using an ARG, as in:

```
ARG baseImage=some/project:version
FROM $baseImage
...
```
This works as long as I don't want a different base image, or if I configure it in the image settings. In theory `docker.buildArg.baseImage=ghcr.io/some/otherProject:otherVersion` should allow changing that value, but that doesn't work since the `autoPullBaseImage` code only looks in the image buildArg configuration.

This PR adds support for looking at all the arg configurations, no matter where they're coming from.

### DIFF
--- a/it/dockerfile-base-as-arg/README.md
+++ b/it/dockerfile-base-as-arg/README.md
@@ -1,0 +1,13 @@
+## d-m-p sample using a Dockerfile
+
+This example shows how to use docker-maven-plugin together with a Dockerfile. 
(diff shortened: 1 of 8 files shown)


rohanKanojia (approved):

rohanKanojia:
@sdumitriu : Thanks a lot!

PR #1761, opened 2024-02-27
Feat: Automatically create parent directories of portPropertyFile path

Willena:
This adds a new option enabling automatic folder creation for the path specified in `portPropertyFile`.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,6 +1,7 @@
 # ChangeLog
 * **0.45-SNAPSHOT**:
-
+  - Automatically create parent directories of portPropertyFile path
+  
 * **0.44.0** (2024-02-17):
   - Add new option "useDefaultExclusion" for build configuration to handle exclusion of hidden files ([1708](https://github.com/fabric8io/docker-maven-plugin/issues/1708))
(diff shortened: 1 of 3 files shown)


rohanKanojia:
@Willena : Hello, do you think it's a good idea to always create parent directories by default?

Willena:
My initials thought were to add a new parameter set to "false" by default to avoid changing the current behaviour (do not create paths). This should be (unless I forgot something) the current behaviour in the PR.  

That beeing said I would personnaly prefer that the plugin always try to create parent paths (and thus I can remove the option and simplifies the code).

rohanKanojia:
@Willena : It's good to be defensive. Your changes look good to me. 

However, I get the feeling that exposing a new configuration parameter just to ensure whether parent directory of `portPropertyFile` exists is a bit overkill.

Willena:
> However, I get the feeling that exposing a new configuration parameter just to ensure whether parent directory of portPropertyFile exists is a bit overkill.

I fully agree with you. I've updated the PR to remove the option. Parent directories are not always created.

rohanKanojia (approved):

rohanKanojia:
@Willena : Thanks! Could you please add a line to doc/changelog.md regarding this change?

Willena:
Done !

PR #1772, opened 2024-03-18
Allow runCmds to change user.

JensGroenborgAndersen:
As discussed in https://github.com/fabric8io/docker-maven-plugin/issues/913#issuecomment-378725546, except the user tag is not limited to the first entry. A user tag will however disable the optimize option


src/test/java/io/fabric8/maven/docker/util/AuthConfigFactoryTest.java
```diff
@@ -101,7 +101,7 @@ public String decrypt(String password) {
 
     @BeforeEach
     void containerSetup() throws ComponentLookupException, SecDispatcherException {
-        Mockito.lenient().when(container.lookup(SecDispatcher.ROLE, "maven")).thenReturn(secDispatcher);
+//        Mockito.lenient().when(container.lookup(SecDispatcher.ROLE, "maven")).thenReturn(secDispatcher);
```
rohanKanojia, line 104: Why is this mockito expectation disabled?

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -669,18 +663,19 @@ private Server checkForServer(Server server, String id, String registry, String
...
+//            Object secDispatcher = container.lookup(SecDispatcher.ROLE, "maven");
+//            Method method = secDispatcher.getClass().getMethod("decrypt",String.class);
+//            synchronized(secDispatcher) {
+//                return (String) method.invoke(secDispatcher, password);
+//            }
+//        } catch (ComponentLookupException e) {
+//            throw new MojoExecutionException("Error looking security dispatcher",e);
+//        } catch (ReflectiveOperationException e) {
+//            throw new MojoExecutionException("Cannot decrypt password: " + e.getCause(),e);
+//        }
```
rohanKanojia, line 677: This code is used for decrypting password from maven settings.xml . Why is this commented out? Disabling this might cause problems for users who are using [encryption with maven](https://maven.apache.org/guides/mini/guide-encryption.html)

src/test/java/io/fabric8/maven/docker/util/AuthConfigFactoryTest.java
JensGroenborgAndersen: Ugh i guess I forgot about those. As i remember, it happened because the Security dispatcher stuff doesn't work with never versions of maven. I guess it should be enabled, so that whoever knows how it is supposed to work can make the appropriate fixes.

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
JensGroenborgAndersen: same as above

PR #1791, opened 2024-06-07
Remove usage of AnsiConsole

slawekjaranowski:
Init of Ansi system is not needed in Maven plugin.

Additionally, Maven 4 switch to JLine
https://issues.apache.org/jira/browse/MNG-7995
and this class will be not available anymore


rohanKanojia:
@slawekjaranowski : Thanks! Is this change backwards compatible with older versions of maven?

PR #1793, opened 2024-06-14
Gracefully handle 404 when removing container

sebastiankirsch:
If a container is to be removed - yet doesn't exist anymore, we can gracefully handle the 404 by treating it as though the container had been removed.

### DIFF
--- a/it/graceful-container-removal/pom.xml
+++ b/it/graceful-container-removal/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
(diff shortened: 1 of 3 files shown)


sebastiankirsch:
Note that I didn't add a test since there was none present for `removeContainer` in the first place, and with the test setup & mocking in place I would only be able to introduce a "signature test" adding (in a specific order) the additional HTTP code in the mock.

rohanKanojia:
@sebastiankirsch : Is it possible to add any integration test (test projects in `it/` directory)

sebastiankirsch:
@rohanKanojia I've added an IT as you suggested.
Execution of `mvn verify` fails with the current version 0.44.0 (at least on my machine :wink: ).

rohanKanojia (approved):

Issue #1711, opened 2023-10-05
Failed to build the docker image on using "chmod" option


lakshmi-dj:
### Info

* docker-maven-plugin version : 0.43.4
* Maven version (`mvn -v`) : 3.6.2
```

```
* Docker version : 
Client:                                     
 Cloud integration: v1.0.33                 
 Version:           24.0.2                  
 API version:       1.43                    
 Go version:        go1.20.4                
 Git commit:        cb74dfc                 
 Built:             Thu May 25 21:53:15 2023
 OS/Arch:           windows/amd64           
 Context:           default                 

Server: Docker Desktop 4.20.1 (110738)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:52:17 2023
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

* If it's a bug, how to reproduce :  Use COPY instruction in dockerfile with "--chmod" option
* If it's a feature request, what is your use case : bug

qerub:
I think setting `<buildOptions><version>2</version></buildOptions>` should unlock BuildKit for you; see https://dmp.fabric8.io/#config-image-build and https://docs.docker.com/engine/api/v1.38/#tag/Image/operation/ImageBuild. Having built-in support in docker-maven-plugin would be nice though!

rohanKanojia:
@qerub : Could you please open a new issue about this? Shall we enable this behavior by default or via some configuration option?

errodrigues:
> I think setting `<buildOptions><version>2</version></buildOptions>` should unlock BuildKit for you; see https://dmp.fabric8.io/#config-image-build and https://docs.docker.com/engine/api/v1.38/#tag/Image/operation/ImageBuild. Having built-in support in docker-maven-plugin would be nice though!

@qerub Can you please confirm that this workaround worked for you? I am facing this same issue and tried your suggestion but I still get same parser error.

I know my Dockerfile is fine because if I build it using podman-maven-plugin or directly using docker-compose, it works fine.

[...]

PR #1794, opened 2024-06-19
feat: add support for dockerd Builder-Version header

rroesch1:
* in docker 23 dockerd introduced the 'Builder-Version' header to expose the recommended version of the builder backend
* it is up to the client to follow this recommendation
* dockerd uses the deprecated 'classic builder' if clients do not set the 'version' build option explicitly
* d-m-p should always set the 'version' option as recommended by the dockerd

Refs: fabric8io/docker-maven-plugin#1711


rohanKanojia:
@rroesch1 : Some integration tests seem to be failing. Could you please take a look at these failures?

rroesch1:
> @rroesch1 : Some integration tests seem to be failing. Could you please take a look at these failures?

Well, i got that but it took me while to realize what's going wrong here:
* I am sorry to say it isn't that easy to get buildkit support …
* enabling buildkit actually works, but builds may fail in some cases
* The reason for this is, that buildkit expects to have some sort of "callback api" to the client. To provide this callback, the client has to initiate a gRPC session (see: [Initialize interactive session](https://docs.docker.com/engine/api/v1.45/#tag/Session/operation/Session) API) + the client has to implement several gRPC services.
* I couldn't find any documentation about these services, but the code for these [seem to here here](https://github.com/moby/buildkit/tree/master/session)
* Creating (and maintaining) a Java implementation for these gRPC services is probably a monstrous task. I am afraid to say that it's probably easier to stop using the Docker Engine API at all and use the CLI (`docker build(x)`) instead. At least for now i have no clue how to deal with that.

PR #1799, opened 2024-07-15
feature: Add the possibility to give secret to buildx build (#1798)

### LINKED ISSUE #1798: How to mount secrets on build?
Is there any way to mount secrets on build?

For example, `docker build --secret id=mytoken,src=$HOME/.aws/credentials .`

Reference: https://docs.docker.com/build/building/secrets/

kevinleturc:
Fix #1798

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
+++ b/src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
@@ -61,6 +61,12 @@ public class BuildXConfiguration implements Serializable {
     @Parameter
     private Map<String, String> driverOpts;
 
+    /**
+     * Secret to expose to the build
+     */
+    @Parameter
+    private SecretConfiguration secret;
+
     public String getBuilderName() {
         return builderName;
     }
@@ -86,7 +92,7 @@ public class BuildXConfiguration implements Serializable {
     }
 
     public boolean isBuildX() {
-        return !getPlatforms().isEmpty();
+        return !getPlatforms().isEmpty() || hasSecret();
     }
 
     @Nonnull
@@ -100,6 +106,14 @@ public class BuildXConfiguration implements Serializable {
 
     public Map<String, String> getDriverOpts() {
         return driverOpts;
+    }
+
+    public boolean hasSecret() {
+        return secret != null;
+    }
+
+    public SecretConfiguration getSecret() {
+        return secret;
     }
 
     public static class Builder {
@@ -159,7 +173,6 @@ public class BuildXConfiguration implements Serializable {
             return this;
         }
 
-
         public Builder cacheFrom(String cacheFrom) {
             config.cacheFrom = cacheFrom;
             if (cacheFrom != null) {
@@ -179,6 +192,14 @@ public class BuildXConfiguration implements Serializable {
         public Builder cacheTo(String cacheTo) {
             config.cacheTo = cacheTo;
             if (cacheTo != null) {
+                isEmpty = false;
+            }
+            return this;
+        }
+
+        public Builder secret(SecretConfiguration secret) {
+            config.secret = secret;
+            if (secret != null) {
                 isEmpty = false;
             }
             return this;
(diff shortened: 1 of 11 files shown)


rohanKanojia:
@kevinleturc : Could you please add a line to doc/changelog.md regarding this change?

Do you think it's possible to add an integration test for this feature?

kevinleturc:
> @kevinleturc : Could you please add a line to doc/changelog.md regarding this change?

Sure

> Do you think it's possible to add an integration test for this feature?

I will check that.

kevinleturc:
Hi @rohanKanojia, I added the changelog line and an it test. For the it test, I ran into the following error in E2E Tests at the beginning:
```
Error:  DOCKER> Unable to build image [dmp/alpine:0.45-SNAPSHOT] : "the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled"  ["the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled" ]
```

In order to fix this error, I needed to use BuildKit, and so edit the `BuildXConfiguration#isBuildX` method to enable the BuildKit driver to be created when there's secret element in configuration but not a platforms/platform one, like below:
```
              <build>
                <dockerFile>${project.basedir}/src/main/docker/Dockerfile</dockerFile>
                <buildx>
                  <secret>
                    <files>
                      <myFile>${project.basedir}/../README.md</myFile>
                    </files>
                  </secret>
                </buildx>
              </build>
```

Which, in my opinion, is a valid usage of the plugin.

But, I don't know if such case could work on the docker buildx driver with some GitHub runner configuration, and so revert the change on `BuildXConfiguration#isBuildX` method behavior.
Otherwise, if the change of behavior is considered legitimate, I'm thinking about other buildx configuration the plugin have, they might need the same behavior? For instance `buildx/attestations`, `buildx/cacheFrom`, etc...

What do think?

rohanKanojia:
> I'm thinking about other buildx configuration the plugin has, they might need the same behavior? 

@kevinleturc : You're right. Adding not null checks for each field might be difficult to keep up with in the future. Do you think we can introduce something like `isEmpty` at BuildXConfiguration level (just like we have in BuildXConfiguration.Builder? 

https://github.com/fabric8io/docker-maven-plugin/blob/e9801891092bd2760e52cffde18d2fe8bae2dc73/src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java#L108

src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java
```diff
@@ -86,7 +92,7 @@ public String getCacheTo() {
     }
 
     public boolean isBuildX() {
-        return !getPlatforms().isEmpty();
+        return !getPlatforms().isEmpty() || hasSecret();
```
rohanKanojia, line 95: This is okay for now but we need to improve this in the future.

rohanKanojia (approved):

kevinleturc:
> @kevinleturc : You're right. Adding not null checks for each field might be difficult to keep up with in the future. Do you think we can introduce something like `isEmpty` at BuildXConfiguration level (just like we have in BuildXConfiguration.Builder? 
> 
> https://github.com/fabric8io/docker-maven-plugin/blob/e9801891092bd2760e52cffde18d2fe8bae2dc73/src/main/java/io/fabric8/maven/docker/config/BuildXConfiguration.java#L108

Yes exactly, something like the current isBuildX but whitout having to state every fields. This should be possible if upper logic juste check buildx nullity to do the buildx things instead of checking if there's more than 1 platform. 

I can prepare a PR for this, maybe outside of this one?
Before changing that, I would check that every configuration in the buildx element require the buildkit engine. 

I can work on that by the beginning of the week, if everything is ok for you?

rohanKanojia:
@kevinleturc : Thanks a lot!

PR #1805, opened 2024-07-23
#1804 support buildx use of default builder

### LINKED ISSUE #1804: Cannot use buildx default driver with local (not multiarchitecture) platform builds
### Description
Our project would like to use a consistent buildx build both locally and on ci.  We believe this is a general pattern that could benefit other projects as well.  

Our general setup is as follows:
* multiple platforms configured (arm64, amd64)
* build phase is correctly invoking only the local architecture
  * NOTE: uses the `maven` `docker-container` buildx driver, resulting in slower builds due to I/O inherent with the `docker-container` driver.  We would like to avoid this extra I/O on local builds.
* We don't ever run the deploy phase locally - so we never build multiple platforms (and thus don't need the `docker-container` output later in the `docker-maven-plugin`'s lifecycle)
* For continuous integration, we use a `ci` Maven profile that skips the build phase, allowing the multiple platform builds that occur at this time to happen simultaneously (which speeds up most of our docker modules)

We'd like to be able to set the buildx`builder-name` to be `default` to allow the use of the default buildx driver on local builds.  This has the following benefits:
* Faster builds due to lower I/O and buildx improvements
* Better visual output
* Consistent docker execution with our ci approach (some modules seem to build differently when not using buildx - we have not narrowed down why yet, but it doesn't matter with this proposed change - but suspect we have some buildx-specific features).

We have a simple PR for this change with a corresponding unit test (will be updated in a moment).

### Info

* docker-maven-plugin version : 0.44.0
* Maven version (`mvn -v`) : 3.9.6
* Docker version : 
```
Client:
 Version:           26.1.0-rd
 API version:       1.44 (downgraded from 1.45)
 Go version:        go1.21.9
 Git commit:        cca8e72
 Built:             Tue Apr 23 16:25:03 2024
 OS/Arch:           darwin/arm64
 Context:           default

Server:
 Engine:
  Version:          25.0.5
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.21.10
  Git commit:       e63daec8672d77ac0b2b5c262ef525c7cf17fd20
  Built:            Sun May 12 07:25:43 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          v1.7.10
  GitCommit:        4e1fe7492b9df85914c389d1f15a3ceedbb280ac
 runc:
  Version:          1.1.12
  GitCommit:        51d5e94601ceffbbd85688df1c928ecccbfa4685
 docker-init:
  Version:          0.19.0
  GitCommit:
```

d-ryan-ashcraft:
Fix #1804

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
@@ -279,30 +279,49 @@ public class BuildXService {
     protected String createBuilder(Path configPath, List<String> buildX, ImageConfiguration imageConfig, BuildDirs buildDirs) throws MojoExecutionException {
         BuildXConfiguration buildXConfiguration = imageConfig.getBuildConfiguration().getBuildX();
         String builderName = Optional.ofNullable(buildXConfiguration.getBuilderName()).orElse("maven");
+
+        if ("default".equals(builderName)) {
+            logger.info("Using default builder with buildx - only single platforms will be supported");
+        } else {
+            createCustomBuilderIfNotExists(configPath, buildX, buildXConfiguration, builderName, buildDirs);
+        }
+        return builderName;
+    }
+
+    private void createCustomBuilderIfNotExists(Path configPath, List<String> buildX, BuildXConfiguration buildXConfiguration, String builderName, BuildDirs buildDirs) throws MojoExecutionException {
         String nodeName = buildXConfiguration.getNodeName();
         Path builderPath = configPath.resolve(Paths.get("buildx", "instances", builderName.toLowerCase()));
-        if(Files.notExists(builderPath)) {
+
+        if (Files.notExists(builderPath)) {
             List<String> cmds = new ArrayList<>(buildX);
             append(cmds, "create", "--driver", "docker-container", "--name", builderName);
+
             if (nodeName != null) {
                 append(cmds, "--node", nodeName);
             }
 
-            if (buildXConfiguration.getDriverOpts() != null && !buildXConfiguration.getDriverOpts().isEmpty()) {
-                buildXConfiguration.getDriverOpts().forEach((key, value) -> append(cmds, "--driver-opt", key + '=' + value));
-            }
+            addDriverOptions(cmds, buildXConfiguration);
+            addBuildConfig(cmds, buildXConfiguration, buildDirs);
 
-            String buildConfig = buildXConfiguration.getConfigFile();
-            if(buildConfig != null) {
-                append(cmds, "--config",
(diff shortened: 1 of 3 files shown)


src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -245,27 +245,31 @@ protected void createDirectory(Path cachePath) {
...
         BuildXConfiguration buildXConfiguration = imageConfig.getBuildConfiguration().getBuildX();
         String builderName = Optional.ofNullable(buildXConfiguration.getBuilderName()).orElse("maven");
-        String nodeName = buildXConfiguration.getNodeName();
-        Path builderPath = configPath.resolve(Paths.get("buildx", "instances", builderName));
-        if(Files.notExists(builderPath)) {
-            List<String> cmds = new ArrayList<>(buildX);
-            append(cmds, "create", "--driver", "docker-container", "--name", builderName);
-            if (nodeName != null) {
-                append(cmds, "--node", nodeName);
-            }
```
d-ryan-ashcraft, line 255: Docker does a nice job of providing meaningful error messages if you try to use the default builder with multiple architectures, etc.  For instance:
```
[INFO] DOCKER> ERROR: Multi-platform build is not supported for the docker driver.
[INFO] DOCKER> Switch to a different driver, or turn on the containerd image store, and try again.
[INFO] DOCKER> Learn more at https://docs.docker.com/go/build-multi-platform/
[ERROR] DOCKER> Error status (1) when building
```

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -245,27 +245,31 @@ protected void createDirectory(Path cachePath) {
     protected String createBuilder(Path configPath, List<String> buildX, ImageConfiguration imageConfig, BuildDirs buildDirs) throws MojoExecutionException {
         BuildXConfiguration buildXConfiguration = imageConfig.getBuildConfiguration().getBuildX();
         String builderName = Optional.ofNullable(buildXConfiguration.getBuilderName()).orElse("maven");
```
d-ryan-ashcraft, line 247: Recommend turning ignore whitespace on to better highlight the change

ewilkins-csi:
PR has been rebased on the latest from the main branch, with merge conflicts resolved.  Should be good to merge assuming there are no other issues and it passes the workflow checks.

d-ryan-ashcraft:
Added documentation and resolved SonarQube Quality Gate issue.

rohanKanojia (approved):

PR #1824, opened 2024-09-21
Pull FROM images in relative path Dockerfiles

### LINKED ISSUE #1823: FROM images in Dockerfiles specified by relative paths are not pulled when building
### Description
If you point to your Dockerfile using a relative path, then any FROM images in your Dockerfile will not be pulled when using the `docker:build` goal.

The documentation for `docker:build` states:
> All paths can be either absolute or relative paths (except when both dockerFileDir and dockerFile are provided in which case dockerFile must not be absolute). A relative path is looked up in ${project.basedir}/src/main/docker by default. You can make it easily an absolute path by using ${project.basedir} in your configuration.

My image's build configuration has not changed from version 0.43.4, but the FROM images in my Dockerfile are no longer pulled. `Dockerfile` and `files.xml` are both in the `src/main/docker` folder of my module.
```
<build>
    <dockerFile>Dockerfile</dockerFile>
    <assembly>
        <descriptor>files.xml</descriptor>
    </assembly>
</build>
```

This regression was introduced in e9801891092bd2760e52cffde18d2fe8bae2dc73. I tested with 0.44.0 and could not reproduce the issue.

### Info
* docker-maven-plugin version : 0.45.0
* Maven version (`mvn -v`) :
```
Apache Maven 3.9.9 (8e8579a9e76f7d015ee5ec7bfcdc97d260186937)
Maven home: /Applications/apache-maven-3.9.9
Java version: 17.0.12, vendor: Amazon.com Inc., runtime: /Library/Java/JavaVirtualMachines/amazon-corretto-17.jdk/Contents/Home
OS name: "mac os x", version: "13.7", arch: "aarch64", family: "mac"
```
* Docker version : `Docker version 27.2.0, build 3ab4256`

jrnorth:
Fixes #1823

### DIFF
--- a/src/test/java/io/fabric8/maven/docker/MojoTestBase.java
+++ b/src/test/java/io/fabric8/maven/docker/MojoTestBase.java
@@ -251,6 +251,7 @@ abstract class MojoTestBase {
         mojo.project = mavenProject;
         mojo.log = ansiLogger;
         mojo.outputDirectory= "target/docker";
+        mojo.sourceDirectory = "src/main/docker";
         mojo.authConfigFactory= authConfigFactory;
         mojo.session = session;
 
--- a/src/main/java/io/fabric8/maven/docker/service/BuildService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildService.java
@@ -370,7 +370,7 @@ public class BuildService {
     }
 
     static List<String> extractBaseFromDockerfile(BuildImageConfiguration buildConfig, MojoParameters mojoParameters, Map<String, String> buildArgs) {
-        if (buildConfig.getDockerFile() == null || !buildConfig.getDockerFile().exists()) {
+        if (buildConfig.getDockerFile() == null || !buildConfig.getAbsoluteDockerFilePath(mojoParameters).exists()) {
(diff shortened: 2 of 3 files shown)


src/test/java/io/fabric8/maven/docker/MojoTestBase.java
```diff
@@ -251,6 +251,7 @@ protected void givenMavenProject(AbstractDockerMojo mojo) {
         mojo.project = mavenProject;
         mojo.log = ansiLogger;
         mojo.outputDirectory= "target/docker";
+        mojo.sourceDirectory = "src/main/docker";
```
jrnorth, line 254: I added this to prevent a NullPointerException in the test [buildUsingConfiguredBuildxWithContext](https://github.com/fabric8io/docker-maven-plugin/blob/master/src/test/java/io/fabric8/maven/docker/BuildMojoTest.java#L116). It sets a contextDir on the BuildImageConfiguration, which causes a default dockerFileFile to be set. This causes the BuildService to attempt to get the FROM images from the Dockerfile, and the source directory needs to be set to calculate the absolute path to the file.

rohanKanojia (approved):

rohanKanojia:
@jrnorth : Thanks a lot!

PR #1412, opened 2020-12-02
Docker build cache friendly example utilizing Spring Boot Layered JAR and Maven Assembly Plugin

mabrarov:
Sample utilizing Spring Boot Layered JAR to make the build more friendly for Docker build cache and to avoid rebuilding all the layers of image when just application code changes.

PR #1828, opened 2024-10-03
Docker build cache friendly example updated

mabrarov:
Update of example introduced in #1412 to make it compatible with recent versions of Docker Engine (tested on 19.03.12 and 27.1.1).


rohanKanojia (approved):

rohanKanojia:
@mabrarov : Thanks a lot!

PR #1833, opened 2024-10-30
Fix for missing classes required on executeStopOnVMShutdown:true

Smiche:
I encountered the following issue:
image

Using:

```
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: C:\Program Files\apache-maven
Java version: 17.0.12, vendor: Eclipse Adoptium, runtime: C:\Program Files\Java\jdk-17.0.12+7
Default locale: en_GB, platform encoding: Cp1252
```
Adding the two required classes to the Class loader resolved the issue. I didn't go in too deep why that check using Strings occurs.

Here's the execution config which lead to this:
```xml
					<execution>
						<id>remove-it-database</id>
						<goals>
							<goal>stop</goal>
						</goals>
						<configuration>
							<executeStopOnVMShutdown>true</executeStopOnVMShutdown>
							<removeVolumes>true</removeVolumes>
							<skip>${skipTests}</skip>
						</configuration>
						<phase>pre-integration-test</phase>
					</execution>
```

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/StopMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/StopMojo.java
@@ -98,6 +98,11 @@ public class StopMojo extends AbstractDockerMojo {
                 // HttpDelete class is not used during start mojo, so we need to load it and initialize it. It is not
                 // possible to load classes in the shutdown hook as
                 Class.forName("org.apache.http.client.methods.HttpDelete", true, this.getClass().getClassLoader());
+
+                // com.google.common.base.Strings class from com.google.guava is required and it depends on Platforms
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks for spotting this! An alternative solution would be to remove Guava altogether from the code base, it doesn't add many benefits (except causing issues like these :)

@rohanKanojia wdyt ? On the other hand, this is a quick fix without any side-effects asfais, so would be happy to merge it, too.

Smiche:
Sounds like a better fix indeed. It was barely used for things that don't even need it.

rohanKanojia (approved):

PR #1838, opened 2024-11-10
Add tag to list of goals

arend-von-reinersdorff:


src/main/asciidoc/inc/_goals.adoc
```diff
@@ -48,6 +48,10 @@ in the next sections.
 |Save images to a file
 |
 
+|**<<{plugin}:tag>>**
+|Tag images
+|install
```
rohanKanojia, line 53:
```suggestion
|Tags an image so that it becomes part of a repository.
```

arend-von-reinersdorff: Improved description to be more detailed

rohanKanojia (approved):

PR #1847, opened 2025-01-01
fix: error already existing instance due to case sensitive Path

Amitrei:
### Description
We are using a per-module builder implementation, where the module name is used as the builder name:
`<builderName> ${project.artifactId} </builderName>
`
When specifying a buildx builder name with capital letters, for example: 
`<builderName> myBuilder </builderName>`,
an error occurs if the builder already exists:
`ERROR: existing instance for "myBuilder" but no append mode, specify --node to make changes for existing instances
`
### Cause
The `docker buildx create` command stores the builder configuration at the following path:
_$DOCKER_CONFIG/buildx/instances/mybuilder_ (Note: the builder name is stored in lowercase).

According to the [Docker buildx implementation](https://github.com/docker/buildx/blob/master/store/util.go#L37), **docker buildx always stores the builder configuration file name in lowercase.**

In the Maven Docker plugin, the following code checks if a builder exists by looking for the builder name inside the _instances_ folder. If the builder doesn't exist, it attempts to create it:

```

...
Path builderPath = configPath.resolve(Paths.get("buildx", "instances", builderName));
if (Files.notExists(builderPath)) {
    List<String> cmds = new ArrayList<>(buildX);
    append(cmds, "create", "--driver", "docker-container", "--name", builderName);
}
...
```

While some file system types are case-insensitive, others are not. This results in the plugin mistakenly attempting to create a builder that already exists. On my macOS, the file system ignores case sensitivity, so the code works fine. However, on our Linux-based CI runners, the "already exists" error is triggered due to case sensitivity.

### The Fix
Simplify the builder name lookup by converting the builderName to lowercase when checking the instances directory. This aligns with the behavior of the docker buildx create command, which resolves the issue of case sensitivity.


rohanKanojia:
@Amitrei : Thanks for the PR! Is it possible to add a unit test to verify path is getting converted to lowercase as expected?

Amitrei:
Hi, as requested I added a test making sure the builder name is lower cased when looking up for a builder config in the `buildx/instances` folder.

rohanKanojia:
@Amitrei : Thanks, your changes look good to me. could you please add a line regarding this change to doc/changelog.md ?

rohanKanojia (approved):

Amitrei:
Hi Rohan,
Thanks for the approval. I've added a note regarding the change in the doc/changelog.md file. Please let me know if you need anything else from me.

Thanks again!

rohanKanojia:
@Amitrei : Thanks a lot!

PR #1848, opened 2025-01-01
Move off direct SecDispatcher use

cstamas:
Use the "official" way instead, and it solves Maven 4 support as well.

Changes:
* align Maven deps (maven-compat was 3.0 while maven-core 3.8.1)
* since Maven 3.9 plexus util is NOT provided
* do not use directly SecDispatcher, use Maven APIs instead
* fix UTs

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
+++ b/src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
@@ -6,7 +6,6 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import com.google.gson.JsonObject;
 
-import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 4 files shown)


cstamas:
An alt fix could be to detect is Mojo running in Maven 4, and if so, simply "do nothing" (make decrypt method no-op), as Maven 4 on boot _decrypts everything in settings.xml_, and there is no need to decrypt anything (as long as thing comes from effective settings).

As explained on Maven ML https://lists.apache.org/thread/4fywo93s80fhmcj58cg9lvj5s61497tc

rohanKanojia (approved):

rohanKanojia:
@cstamas : Thanks a lot!

src/main/java/io/fabric8/maven/docker/util/AuthConfigFactory.java
```diff
@@ -670,16 +671,15 @@ private Server checkForServer(Server server, String id, String registry, String
...
-            synchronized(secDispatcher) {
-                return (String) method.invoke(secDispatcher, password);
-            }
+            SettingsDecrypter settingsDecrypter = container.lookup(SettingsDecrypter.class);
+            Server stub = new Server();
+            stub.setUsername("whatever");
+            stub.setPassword(password);
+
+            SettingsDecryptionResult result = settingsDecrypter.decrypt(new DefaultSettingsDecryptionRequest(stub));
+            return result.getServers().get(0).getPassword();
```
mfriedenhagen, line 680: What if there are multiple server credentials in `settings.xml`, e.g. against `docker.io` for the base image and e.g. `ghcr.io` for the derivated application image? And why is the username irrelevant? 
AFAIK the first a docker client will do is to login via basic auth (maybe by using a registry specific token) and retrieve a scoped bearer token (https://docs.docker.com/registry/#authentication) for further communication.

PR #1850, opened 2025-01-03
fix: respects network configuration for buildx

michaelmejaeger:
### Description

I have found that the configuration of the network (in Docker with the parameter `--network`) is not respected in multi-arch builds using `buildx`. One can configure the network as stated in the documentation, however, this has no effect on the commandline call of `docker`.

### Cause

The reason is that the configuration of `network` is simply not considered.

### Solution

I provided a unit tests which shows the problem and a fix that considers the configuration done via System properties, Maven properties, and in the configuration of the plugin.


shazzerim:
this fix would help us greatly!

michaelmejaeger:
Hello, is there anything more that I can/should do in this PR?

rohanKanojia:
ah, I see it's a fix, not a feature. Please ignore my previous comment.

rohanKanojia (approved):

rohanKanojia:
@michaelmejaeger : Thanks a lot!

PR #1852, opened 2025-01-07
docs: Correct removeVolumes description

jakub-bochenski:
I believe it works the same as v option https://docs.docker.com/reference/api/engine/version/v1.43/#tag/Container/operation/ContainerDelete so it doesn't remove "any" containers only the anonymous ones.


rohanKanojia:
@jakub-bochenski : Thanks a lot!

PR #1855, opened 2025-02-24
Allow non-standard build directory.

jdelker:
Currently, the general `outputDirectory` has a fixed default of `target/docker`, which breaks projects with a diverting `${project.build.directory}`.
This change simply adjusts the default, so this aligns correctly.


rohanKanojia:
@jdelker : Thanks, could you please add an entry to doc/changelog.md regarding this change?

jdelker:
> @jdelker : E2E tests seem to be failing. Could you please take a look?

Well, I'm lacking insight about what really happens in those integration tests to find the source of that failure.
The change is tiny, but it does effectively change the `outputDir` to contain an absolute path, rather than a relative.
From my understanding that should be perfectly fine, so I would assume that some parts do have problematic code not handling this correctly.

PR #1864, opened 2025-04-22
Fix #1863 by treating a buildArchiveOnly path as `true`

### LINKED ISSUE #1863: buildArchiveOnly with a path still builds the image
The `buildArchiveOnly` option implemented in #1297 does not behave [as documented](https://dmp.fabric8.io/#global-configuration) for the case when a path is provided as the value. The issue appears to be that the `shallBuildArchiveOnly` method [returns `false`](https://github.com/fabric8io/docker-maven-plugin/pull/1297/files) if the value is not literally the strings "true" or "false" (case-insensitive).

ewilkins-csi:
Adds a fall-through return for when `buildArchiveOnly` is not empty, but is not a boolean.  Adds unit tests for this logic and the default case to ensure `buildArchiveOnly` continues to default to `false` if not set.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/BuildMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/BuildMojo.java
@@ -137,6 +137,7 @@ public class BuildMojo extends AbstractDockerMojo {
                     buildArchiveOnly.equalsIgnoreCase("true")) {
                 return Boolean.parseBoolean(buildArchiveOnly);
             }
+            return true;
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


ewilkins-csi:
E2E failure looks like a transient Docker Hub 500 error:
```sh
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.fabric8:docker-maven-plugin:0.47-SNAPSHOT:build (build) on project dmp-it-docker-compose-dependon: Unable to pull 'postgres:16.3-alpine' : {"message":"Head \"[https://registry-1.docker.io/v2/library/postgres/manifests/16.3-alpine\](https://registry-1.docker.io/v2/library/postgres/manifests/16.3-alpine/)": EOF"} (Internal Server Error: 500)
```

Can the failed Linux job just be re-run?

PR #1868, opened 2025-05-16
Make all mojos threadsafe

meliz19:
## Description
When using `mvnd` to build Docker images with the `BuildMojo`, users get the following warning

```bash
[INFO] ----------------------------[ docker-build ]----------------------------
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but this         *
[WARNING] * project contains the following plugin(s) that have goals not  *
[WARNING] * marked as thread-safe to support parallel execution.          *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against Apache Maven.                           *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked as thread-safe in <modules>:
[WARNING]   io.fabric8:docker-maven-plugin:0.46.0
[WARNING] 
[WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
[WARNING] *****************************************************************
```

The purpose of this PR is to resolve this warning by making all of the Mojos thread safe.

## Notes
`testExtraHostsDoesNotResolve` method was broken, so I updated the test to resolve the issue.

### DIFF
--- a/src/test/java/io/fabric8/maven/docker/access/ContainerHostConfigTest.java
+++ b/src/test/java/io/fabric8/maven/docker/access/ContainerHostConfigTest.java
@@ -21,7 +21,7 @@ class ContainerHostConfigTest {
     @Test
     void testExtraHostsDoesNotResolve() {
         ContainerHostConfig hc = new ContainerHostConfig();
-        List<String> extraHosts = Collections.singletonList("database.pvt:ahostnamewhichreallyshouldnot.exist.zz");
+        List<String> extraHosts = Collections.singletonList("database.pvtahostnamewhichreallyshouldnot.exist.zz");
         Assertions.assertThrows(IllegalArgumentException.class, () -> hc.extraHosts(extraHosts));
     }
 
--- a/src/main/java/io/fabric8/maven/docker/BuildMojo.java
+++ b/src/main/java/io/fabric8/maven/docker/BuildMojo.java
@@ -35,7 +35,7 @@ import static io.fabric8.maven.docker.service.RegistryService.createCompleteAuth
  * @author roland
  * @since 28.07.14
  */
-@Mojo(name = "build", defaultPhase = LifecyclePhase.INSTALL, requiresDependencyResolution = ResolutionScope.TEST)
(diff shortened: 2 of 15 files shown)


src/test/java/io/fabric8/maven/docker/access/ContainerHostConfigTest.java
```diff
@@ -21,7 +21,7 @@ class ContainerHostConfigTest {
     @Test
     void testExtraHostsDoesNotResolve() {
         ContainerHostConfig hc = new ContainerHostConfig();
-        List<String> extraHosts = Collections.singletonList("database.pvt:ahostnamewhichreallyshouldnot.exist.zz");
+        List<String> extraHosts = Collections.singletonList("database.pvtahostnamewhichreallyshouldnot.exist.zz");
```
meliz19, line 24: `ContainerHostConfig.extraHosts` throws an `IllegalArgumentException` if an item in the `singletonList` has a length of 1. With the original value of `"database.pvt:ahostnamewhichreallyshouldnot.exist.zz"`, the exception isn't raised as the length is 2.

rhuss: Sorry, I can't follow.

meliz19: The purpose of the unit test appears to be checking that an exception is raised if a string without a colon is present. It does this by splitting the string and verifying that the length is 1 (ie, colon not present). It doesn't appear that the codebase around this test has changed recently, but I either missed something or a dependency changed. When you run mvn clean install on this unit test it fails saying the exception isn't wasn't raised as expected.

meliz19: I apologize for the delayed response. I started a new job this week. That said, I'm now using my personal computer and I'm on a different WiFi network. Between this combination, I am not able to recreate the issue now. I was using an M1 Mac on an AT&T fiber network. Now I'm using an Intel Mac. I'm not sure what the internet provider is. 

I'll remove this change from the PR, but it might be worth investigating if you have someone else run into this issue.

PR #1891, opened 2025-11-13
Allow saving multiple images in docker:save using new parameters saveNames and saveAliases.

Uklosk:
Docker API allows exporting several images to the same tar in the same request:

https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Image/operation/ImageGetAll

I needed this feature to save the same image with different tags (e.g. :latest and :version) to the same tar file.

The proposed solution is to add two new parameters to the save mojo, "saveNames" and "saveAliases", where you can put a list of image names or aliases separated with commas.

I tried to keep the original behavior as intact as possible, I also added a few tests.

I think this feature may be useful for some people, please feel free to provide any feedback or directly reject the PR.

Regards,
Uklosk

### DIFF
--- a/src/main/asciidoc/inc/_docker-save.adoc
+++ b/src/main/asciidoc/inc/_docker-save.adoc
@@ -41,11 +41,19 @@ Note that using overriding the default to use `docker` or `docker-%a` may lead t
 | Element | Description | Property
 
 | *saveName*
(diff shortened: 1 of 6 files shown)


rohanKanojia:
@Uklosk Thanks a lot!

PR #1913, opened 2026-04-29
fix #1872

### LINKED ISSUE #1872: Plugin does not work with AWS VPCE
We are trying to use the plugin in an environment with AWS VPCE using the docker login authentication. When pushing an image to our registry, the plugin tries to connect to api.ecr.eu-south-1.amazonaws.com instead of our VPCE api server.

Are we missing some configuration?

Thanks

mbarinc:
this PR extend VPC endpoint support

- Added constant ECR_ENDPOINT_PROPERTY = "docker.ecr.endpoint"
- Added endpoint field
- In createSignedRequest(), the host resolution logic is now:
 1. If endpoint is passed to the constructor → use that
 2. Otherwise read the system property docker.ecr.endpoint
 3. Otherwise fallback to the default api.ecr.<region>.amazonaws.com

Example of additional parameter: -Ddocker.ecr.endpoint=vpce-XXXXXXXXX.api.ecr.eu-south-1.vpce.amazonaws.com

### DIFF
--- a/src/main/asciidoc/inc/_authentication.adoc
+++ b/src/main/asciidoc/inc/_authentication.adoc
@@ -221,3 +221,14 @@ This extra dependency allows the usage of all link:https://docs.aws.amazon.com/s
 
 If the AWS SDK is found in the classpath, it takes precedence over the custom AWS credentials lookup mechanisms listed above.
 ====
+
+==== Custom ECR API Endpoint (VPC Endpoint)
+
+By default, the plugin contacts the ECR API at `api.ecr.<region>.amazonaws.com`.
+If your environment requires routing through a VPC Endpoint (e.g. in private subnets without internet access), you can override the ECR API endpoint using the system property `docker.ecr.endpoint`:
+
+.Example
+[source, sh, subs="+attributes"]
+----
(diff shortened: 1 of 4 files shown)


mbarinc:
I looked at the failed action logs.

The MacOS build fails because **actions/setup-java@v1** downloads an x64 JDK on an ARM64 runner (macos-14-arm64). 
This is unrelated to this PR, the runner image was updated on April 27 (**20260427.0019**) and likely dropped Rosetta 2 support:

```
2026-04-06T12:35:00.1949420Z ##[group]Runner Image
2026-04-06T12:35:00.1949730Z Image: macos-14-arm64
2026-04-06T12:35:00.1950030Z Version: 20260330.0199.1
2026-04-06T12:35:00.1950780Z Included Software: https://github.com/actions/runner-images/blob/macos-14-arm64/20260330.0199/images/macos/macos-14-arm64-Readme.md
2026-04-06T12:35:00.1951750Z Image Release: https://github.com/actions/runner-images/releases/tag/macos-14-arm64%2F20260330.0199
```

I think this is a common problem with all new developments, the fix would be upgrading to actions/setup-java@v4 in the workflow.

rohanKanojia (approved):

PR #1921, opened 2026-07-04
Honour <follow> configuration element for start/run/watch (#1797)

### LINKED ISSUE #1797: Follow settings ingored unless specified from command line
### Description
None of this works
```
            <properties>
                <docker.follow>true</docker.follow>
            </properties>
            <build>
                <pluginManagement>
                    <plugins>
                        <plugin>
                            <groupId>io.fabric8</groupId>
                            <configuration>
                                <follow>true</follow>
                            </configuration>
                        </plugin>
                    </plugins>
                </pluginManagement>
            </build>
```
the setting is only respected when specified from command line.

This is confusing. 
Also I can't have a nice `mvn -Prun` command working.

### Info

* docker-maven-plugin version : 0.44

JamBalaya56562:
### Problem

The `follow` flag (block and stream container logs until interrupted) could only be enabled through the `docker.follow` **system property**. A `<follow>` element in the plugin `<configuration>` was silently ignored, because:

- `StartMojo.follow` had **no `@Parameter` annotation**, and `StartMojo.followLogs()` read `System.getProperty("docker.follow")` directly.
- `WatchMojo` had its own private `follow()` doing the same.

So `docker:start` / `docker:run` / `docker:watch` did not honour `<follow>` in the POM (reported in #1797).

### Fix

- **StartMojo**: bind `follow` to `@Parameter(property = "docker.follow")`. It is a nullable `Boolean` (no `defaultValue`) so the per-goal default can differ: `followLogs()` resolves `null → false`.
- **RunMojo**: `followLogs()` now resolves `null → true`, preserving `docker:run`'s "follow by default" behaviour.
- **WatchMojo**: replace the ad-hoc `follow()` method with a plain `@Parameter(property = "docker.follow", defaultValue = "false")` field (same idiom as `LogsMojo`).

`-Ddocker.follow` keeps working unchanged, since Maven injects the user property into the annotated field (same mechanism `LogsMojo` already uses). The resolved value still flows into `StartContainerExecutor`, so the existing "show logs automatically when following" default is untouched.

### Tests

- New `StartMojoTest` covers the resolution matrix: `docker:start` defaults to `false`, `docker:run` defaults to `true`, and both honour an explicit `<follow>` value.
- Full unit-test suite green (`936` tests, `0` failures).

Documentation (`start` overview) and `doc/changelog.md` updated.

Closes #1797

### DIFF
--- a/src/test/java/io/fabric8/maven/docker/StartMojoTest.java
+++ b/src/test/java/io/fabric8/maven/docker/StartMojoTest.java
@@ -1,0 +1,49 @@
+package io.fabric8.maven.docker;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that the {@code follow} log-following flag can be driven from the {@code <follow>}
+ * configuration element (and not only the {@code docker.follow} system property), while keeping
+ * the historical defaults: {@code docker:start} does not follow, {@code docker:run} does. See #1797.
+ */
+class StartMojoTest {
+
+    @Test
+    @DisplayName("docker:start does not follow logs unless configured (default false)")
+    void startFollowDefaultsToFalse() {
+        Assertions.assertFalse(new StartMojo().followLogs());
+    }
+
+    @Test
+    @DisplayName("docker:start honours the <follow> configuration element")
+    void startHonoursConfiguredFollow() {
+        StartMojo mojo = new StartMojo();
+
+        mojo.follow = Boolean.TRUE;
+        Assertions.assertTrue(mojo.followLogs());
+
+        mojo.follow = Boolean.FALSE;
+        Assertions.assertFalse(mojo.followLogs());
+    }
+
+    @Test
+    @DisplayName("docker:run follows logs by default (default true)")
+    void runFollowDefaultsToTrue() {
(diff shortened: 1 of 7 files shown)


src/test/java/io/fabric8/maven/docker/StartMojoTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @DisplayName("docker:run honours the <follow> configuration element")
+    void runHonoursConfiguredFollow() {
+        RunMojo mojo = new RunMojo();
+
+        mojo.follow = Boolean.FALSE;
+        Assertions.assertFalse(mojo.followLogs());
+
+        mojo.follow = Boolean.TRUE;
+        Assertions.assertTrue(mojo.followLogs());
+    }
```
rohanKanojia: I don't think it's good to have `RunMojo` tests in `StartMojoTest`. Could you please move it to dedicated test class

JamBalaya56562: Good catch, thanks! I've moved the `RunMojo` assertions into a dedicated `RunMojoTest` class and kept `StartMojoTest` focused on `docker:start` (commit c3806d2). The branch is also rebased onto the latest `master`.

rohanKanojia (approved):

PR #1929, opened 2026-07-04
feat: add wslc:// transport for WSL Containers (wslc) (fixes #1928)

### LINKED ISSUE #1928: Support connecting to the Docker daemon running under WSL Containers (wslc://)
### Feature request
Add a new `wslc://` docker host transport so the plugin can talk to the Docker daemon that ships with **WSL Containers** (`wslc`, available in WSL 2.9.x and later, see https://learn.microsoft.com/windows/wsl/wsl-container).
Benefits: no more need to install podman or docker, only WSL !
### Problem
Unlike Docker Desktop or Podman, WSL Containers exposes **neither a Windows named pipe (`//./pipe/docker_engine`) nor a TCP port** on the host. The Docker daemon runs inside the wslc virtual machine and is only reachable through a stdio bridge:
  ```
  wslc system session run docker system dial-stdio
  ```
As a result, `docker-maven-plugin` cannot auto-detect or connect to a WSL Containers daemon today. On such a machine the connection detector falls through all existing providers and fails with:
  ```
  No <dockerHost> given, no DOCKER_HOST environment variable,
  no read/writable '/var/run/docker.sock' or '//./pipe/docker_engine' ...
  ```
### Proposed solution
Introduce a `wslc://` scheme, analogous to how `DOCKER_HOST=ssh://` works: the transport spawns the bridge command above and lets Apache HttpClient drive its stdin/stdout as if it were an ordinary socket (one bridge process per HTTP connection, reused across requests via HTTP keep-alive).
Concretely:
  1. **New transport** (`wslc://`) built on the existing
     `AbstractNativeClientBuilder` / `AbstractNativeSocketFactory` abstractions (same pattern as the `unix://` and `npipe://` transports). A process-backed `Socket` bridges HttpClient to `docker system dial-stdio`.
  2. **Auto-detection**: add a Windows-only provider that selects `wslc://` when the `wslc` CLI is present. Priority is placed **below** the Docker/Podman named pipe, so an existing Docker Desktop / Podman endpoint always wins and wslc is only used as a fallback. New discovery order:
     1. explicit `dockerHost`
     2. `DOCKER_HOST` env var
     3. `/var/run/docker.sock` (Unix)
     4. `//./pipe/docker_engine` (Windows named pipe)
     5. **`wslc://` if the `wslc` CLI is available (Windows)** ← new
  3. **Executable override**: the wslc executable can be pointed at a non-default location via the `WSLC_EXECUTABLE` environment variable (kept out of the URL so Windows paths with spaces/backslashes don't have to survive URI parsing).
### Notes / trade-offs
  - Auto-detection probes with `wslc version` (a cheap metadata call that does **not** start the container VM). It only checks that the CLI is present, so if `wslc` is installed but no session/daemon is ready, wslc is still selected and the failure surfaces as a connection error at build time.
  - Like the named-pipe transport, the process-backed socket has no socket-level read timeout; a hung daemon blocks the read until the bridge process ends.
  - There is an open change request against WSL itself [microsoft/wsl#40976](https://github.com/microsoft/wsl/issues/40976) that, if implemented, would have WSL Containers expose the Docker daemon through a host-visible endpoint (e.g. a Windows named pipe or TCP port) natively.
If/when that lands, the `wslc://` stdio-bridge proposed here becomes largely redundant: the daemon would be reachable through the plugin's **existing** `npipe://` (or `tcp://`) transport and would be picked up by the current named-pipe auto-detection with no wslc-specific code path. In other words, this feature fills a gap that exists in today's `wslc` releases; its usefulness will diminish once WSL provides a native endpoint.
### PR
I have a working implementation (new `wslc://` transport + auto-detection + unit tests + docs), PR assisted with Claude Code.

DavidTavoularis:
WSL Containers (wslc, shipping with WSL 2.9.x+) exposes neither a Windows named pipe nor a TCP port; its Docker daemon is only reachable through a stdio bridge (`wslc system session run docker system dial-stdio`). Add a new `wslc://` Docker host transport, mirroring the existing unix:// and npipe:// transports:

- WslcProcessSocket: a Socket backed by the bridge child process, so Apache HttpClient (incl. hijacked exec/attach/log streams) drives it normally.
- WslcConnectionSocketFactory / WslcClientBuilder + `wslc` scheme wiring in DockerAccessWithHcClient.
- WslcDockerHostProvider: auto-detects wslc as a fallback (priority 45, below the docker_engine named pipe), so an existing Docker Desktop / Podman endpoint always wins.
- WSLC_EXECUTABLE env override for non-default wslc locations.
- Unit tests and <dockerHost> documentation.

Fixes #1928

### DIFF
--- a/src/main/asciidoc/inc/_global-configuration.adoc
+++ b/src/main/asciidoc/inc/_global-configuration.adoc
@@ -62,8 +62,20 @@ a| The URL of the Docker Daemon. If this configuration option is not given, then
 depending on whether plain HTTP communication is enabled or SSL should
 be used. Alternatively the scheme could be `tcp` in which case the
 protocol is determined via the IANA assigned port: 2375 for `http`
-and 2376 for `https`. Finally, Unix sockets are supported by using
+and 2376 for `https`. Unix sockets are supported by using
 the scheme `unix` together with the filesystem path to the unix socket.
+Windows named pipes are supported with the scheme `npipe`. Finally,
+https://learn.microsoft.com/windows/wsl/wsl-container[WSL Containers (wslc)]
+are supported with the scheme `wslc`: since wslc exposes neither a Windows
+named pipe nor a TCP port, the plugin reaches the Docker daemon running inside
+the wslc virtual machine through a stdio bridge
+(`wslc system session run docker system dial-stdio`). The URL path, if given,
+is the wslc executable to invoke, e.g. `wslc:///wslc.exe` (defaults to `wslc.exe`
+on the `PATH`; can also be overridden with the `WSLC_EXECUTABLE` environment
+variable). This requires a WSL installation that provides `wslc` (WSL 2.9.x or
+later) with an available session. Note that auto-detection only checks that the
+`wslc` CLI is present, so if `wslc` is installed but no session/daemon is ready
+it is still selected and the failure surfaces as a connection error at build time.
 The discovery sequence used by the docker-maven-plugin to determine
 the URL is:
 
@@ -72,6 +84,7 @@ the URL is:
 . the value of the environment variable `DOCKER_HOST`.
 . `/var/run/docker.sock` if it is a readable socket (Unix & OS X).
 . `//./pipe/docker_engine` if it is a readable named pipe (Windows)
+. `wslc://localhost` if `wslc` (WSL Containers) is installed and available (Windows)
 | `docker.host`
 
 | *filter*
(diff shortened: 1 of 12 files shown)


DavidTavoularis:
### Testing
  - Manually verified end-to-end on wslc 2.9.3.0 with `DOCKER_HOST` unset (and podman machine stopped), the plugin auto-detects `wslc://localhost` and `docker:build` builds & tags a real image over the bridge.

rohanKanojia (approved):

rohanKanojia:
@DavidTavoularis : Thanks a lot! Much appreciated 🙏

PR #1952, opened 2026-07-26
Apply <labels> when building from a provided Dockerfile (buildx --label + classic build option)

### LINKED ISSUE #1914: <labels> block silently ignored when <buildx> is configured
## Summary

When an image's `<build>` configuration includes both a `<labels>` element and a `<buildx>` element (multi-platform build), the labels are silently dropped — they appear nowhere in the resulting image's `Config.Labels`. Classic (non-buildx) builds in the same project correctly honour `<labels>`.

## Versions

- Reproduced on **0.45.1**
- Source-verified still present on **0.48.1** (current latest at time of writing)

## Reproduction

Minimal `pom.xml` snippet:

```xml
<image>
  <name>example/foo:latest</name>
  <build>
    <buildx>
      <platforms>
        <platform>linux/arm64</platform>
      </platforms>
    </buildx>
    <dockerFileDir>${project.basedir}/src/main/docker</dockerFileDir>
    <labels>
      <org.opencontainers.image.title>Example</org.opencontainers.image.title>
      <org.opencontainers.image.vendor>Example GmbH</org.opencontainers.image.vendor>
      <org.opencontainers.image.version>1.0.0</org.opencontainers.image.version>
    </labels>
  </build>
</image>
```

Run `mvn docker:build`, then inspect:

```bash
docker inspect example/foo:latest --format '{{json .Config.Labels}}'
```

**Expected:** the three `org.opencontainers.image.*` labels appear in the output.
**Actual:** `null` (no `Labels` key emitted at all).

Removing the `<buildx>` element so the same image goes through the classic build path makes the labels appear as expected.

## Root cause

`io.fabric8.maven.docker.service.BuildXService.buildX(...)` constructs the `docker buildx build` command line and emits flags for `--platform`, `--tag`, `--build-arg`, `--no-cache`, `--sbom`, `--provenance`, `--cache-from`, `--cache-to`, `--secret`, `--squash` — but never calls `BuildImageConfiguration.getLabels()` and never appends `--label` flags. Confirmed by source inspection on the `v0.48.1` tag.

## Expected behaviour

`<labels>` should be propagated to `docker buildx build` as one `--label key=value` flag per entry, matching the behaviour of the classic build path. This is important for `org.opencontainers.image.*` metadata (title / vendor / source / revision / version / licenses) that registry tooling, CVE scanners, and supply-chain audit tools all read from image-config labels.

## Workaround

For now, projects either:
1. Add `LABEL` instructions in the `Dockerfile` with `ARG`-plumbed values (fragmenting the label set away from the rest of the d-m-p config), or
2. Accept that buildx-built images go unlabelled.

Happy to put together a PR if there's appetite — the fix looks like a small addition to the buildx command construction in `BuildXService`.

JamBalaya56562:
Closes #1914

## Root cause (slightly different from the issue report)

`<labels>` has exactly one implementation: they are written as a `LABEL` instruction into the Dockerfile **that this plugin generates** — `DockerAssemblyManager#createDockerFileBuilder` → `.labels(buildConfig.getLabels())`, reached only from the `else` branch of `isDockerFileMode()`.

So the deciding factor is **not buildx**, it is whether a Dockerfile is provided. With `<dockerFile>` / `<dockerFileDir>` / `<contextDir>` the user's Dockerfile is used verbatim and the labels are silently dropped — **for the classic build too**, not only for buildx.

I verified this on Docker 29.6.2 / buildx v0.35.0 with the plugin built from `master`, using the same `<labels>` in four configurations:

| Build | Dockerfile | `Config.Labels` before |
|---|---|---|
| classic | generated | ✅ all three labels |
| classic | **provided** | ❌ `null` |
| buildx | generated | ✅ all three labels |
| buildx | **provided** (the config in #1914) | ❌ `null` |

The buildx command line confirms it: `--platform`, `--tag`, `--file=…` are emitted, `--label` never is. And the classic path never sends the `labels` build option to the daemon (`BuildOptions` had no `labels`, so the Engine API `labels` parameter went unused).

So there were two independent gaps, and fixing only buildx would have left classic + provided Dockerfile broken.

## Change

- `BuildXService#buildX`: emit `--label key=value` per entry, mirroring the neighbouring `--build-arg` block. No signature change — `buildX` already receives the `ImageConfiguration`.
- `BuildOptions#labels(Map)`: new option that serialises to the Engine API `labels` parameter, following the existing `buildArgs(Map)` pattern.
- `BuildService`: pass `buildConfig.getLabels()` into the `BuildOptions` chain.

Both are emitted **only when labels are configured**, so command lines and build options are unchanged for projects without `<labels>`. For a generated Dockerfile the labels are now applied twice (Dockerfile `LABEL` + flag/option) with identical values, which is idempotent.

## Verification

Re-ran the same four configurations with the patched plugin:

| Build | Dockerfile | `Config.Labels` after |
|---|---|---|
| classic | generated | ✅ |
| classic | **provided** | ✅ (was `null`) |
| buildx | generated | ✅ |
| buildx | **provided** | ✅ (was `null`) |

```
docker image inspect repro/file-buildx:latest --format '{{json .Config.Labels}}'
{"org.opencontainers.image.title":"Example","org.opencontainers.image.vendor":"Example GmbH","org.opencontainers.image.version":"1.0.0"}
```

Label values containing spaces survive correctly (`Example GmbH`) — the buildx command is executed via `ProcessBuilder`, so no shell quoting is involved.

Tests added:
- `BuildXServiceTest#testBuildXPassesLabels` / `#testBuildXOmitsLabelFlagWhenNoLabelsConfigured`
- `BuildServiceTest#testLabelsArePropagatedToBuildOptions` / `#testLabelsAbsentFromBuildOptionsWhenNotConfigured`

The negative tests guard the exact-command-line assertions in `BuildMojoTest`, `RegistryServiceTest` and `PushMojoBuildXTest`, which would break if the flag were emitted unconditionally.

Full build and test suite passes on Temurin 21.


src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -175,6 +175,16 @@ protected void buildX(List<String> buildX, String builderName, BuildDirs buildDi
...
             });
         }
+
+        // Labels are only baked into the Dockerfile when it is generated by this plugin, so they have
+        // to be passed explicitly for builds using a provided Dockerfile (dockerFile/contextDir mode).
+        Map<String, String> labels = buildConfiguration.getLabels();
+        if (labels != null) {
+            labels.forEach((key, value) -> {
+                cmdLine.add("--label");
+                cmdLine.add(key + '=' + value);
```
rohanKanojia: A null label value becomes the literal string `null` here.
```suggestion
                cmdLine.add(key + '=' + (value != null ? value : ""));
```

src/test/java/io/fabric8/maven/docker/service/BuildServiceTest.java
```diff
@@ -443,6 +444,59 @@ void testTargetAbsentFromBuildOptionsWhenNotConfigured() throws DockerAccessExce
...
+            .build();
+        File buildArchive = buildService.buildArchive(imageConfig, buildContext, "");
+
+        // When
+        buildService.buildImage(imageConfig, null, false, false, Collections.emptyMap(), buildArchive);
+
+        // Then
+        Mockito.verify(docker).buildImage(Mockito.any(), Mockito.any(),
+            Mockito.argThat((BuildOptions options) ->
+                "{\"org.opencontainers.image.title\":\"Example\"}".equals(options.getOptions().get("labels"))));
```
rohanKanojia: This couples the assertion to Gson's exact output. Parsing the value back into a map would survive a serializer or ordering change and would let the test cover more than one entry

src/test/java/io/fabric8/maven/docker/service/BuildServiceTest.java
```diff
@@ -443,6 +444,59 @@ void testTargetAbsentFromBuildOptionsWhenNotConfigured() throws DockerAccessExce
...
+            .build();
+        File buildArchive = buildService.buildArchive(imageConfig, buildContext, "");
+
+        // When
+        buildService.buildImage(imageConfig, null, false, false, Collections.emptyMap(), buildArchive);
+
+        // Then
+        Mockito.verify(docker).buildImage(Mockito.any(), Mockito.any(),
+            Mockito.argThat((BuildOptions options) -> !options.getOptions().containsKey("labels")));
+    }
```
rohanKanojia, line 544: The four new tests all build with a plugin-generated Dockerfile. They correctly verify the propagation mechanism, but the scenario from #1914 `<dockerFile>` / `<contextDir>` isn't directly exercised anywhere.

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
JamBalaya56562: Right, and it is worth saying why the `--build-arg` loop just above is safe from
the same thing: build args go through `prepareBuildArgs`, which collects them
into a Guava `ImmutableMap` and so refuses null values outright. Labels have no
such gate.

Taken, with one addition — the classic path had the same hole in a different
spelling. `BuildOptions.labels` handed the map straight to Gson, so a null value
arrived at the daemon as a JSON null. Both now send an empty label, which is
what a valueless label means.

A test on each path, both confirmed to fail without the change: the classic one
on `UnsupportedOperation JsonNull`, this one on the missing
`org.opencontainers.image.title=` argument.

src/test/java/io/fabric8/maven/docker/service/BuildServiceTest.java
JamBalaya56562: Agreed. It now reads the labels back out of the captured build options into a
map and compares that, so neither the serialiser nor the ordering matters, and
it carries two entries rather than one.

src/test/java/io/fabric8/maven/docker/service/BuildServiceTest.java
JamBalaya56562: You are right, and that is the more important of the three. Every one of the new
tests built with a Dockerfile generated by the plugin, while the issue is about a
provided one.

The propagation test now sets `<dockerFile>`, so the reported scenario is what
is pinned. The mechanism itself is not conditional on the mode — `.labels(...)`
and `--label` are added either way — so the old tests were not wrong, but they
would not have caught a change that made propagation depend on how the Dockerfile
was obtained. Now they would.

PR #1964, opened 2026-08-12
Report an unusable log output file instead of failing with a NullPointerException

JamBalaya56562:
> Builds on #1963 (tests only). GitHub shows both commits until that one merges; the fix itself is the second commit.

While writing the tests in #1963 I ran into three latent problems in `AnsiLogger`. That PR deliberately left them alone and only noted them; this one fixes them.

## The main one: an unusable `<outputFile>` fails with a NullPointerException

The constructor caught `FileNotFoundException`, printed it to stderr and carried on — leaving `outputFile` non-null while `pw` stayed null. The next log call then died on the field:

```
[ERROR] Failed to execute goal ... Cannot invoke "java.io.PrintWriter.println(String)" because "this.pw" is null
```

The actual cause only appeared as an unformatted stack trace, unconnected to the failure. This is also the only `e.printStackTrace()`-and-continue in the tree; everywhere else an `IOException` becomes a typed exception.

**Fix, in two steps:**

1. **Create the parent directory first.** `AbstractDockerMojo` builds the file with a plain `new File(outputFile)`, so `-DoutputFile=target/logs/build.log` fails purely because `target/logs` does not exist yet — the most likely way to hit this. `DefaultLogCallback` already creates parent directories for container log files, so this only brings the two in line.
2. **If it still cannot be opened, say so.** The failure surfaces as an `UncheckedIOException` (the same unchecked conversion `DockerFileUtil` and `ContainerCreateConfig` use for `FileNotFoundException`), which `AbstractDockerMojo` turns into a `MojoExecutionException` naming the path — matching how `SaveMojo` and `EnvUtil` report directories they cannot create.

Constructor signatures are unchanged, so none of the ~35 call sites needed touching.

**Ordering matters here:** opening the file now happens *before* `initializeColor()`. That call flips the global `Ansi` state, and `AbstractDockerMojo` restores it from a `finally` block that a failure at construction time never reaches — so throwing after it would have left the setting flipped for the rest of the build.

`logOrPrintToFile` also checks `pw` rather than `outputFile` now, so a logger in an unexpected state degrades to console output instead of NPE-ing inside the error handling that is trying to report the original problem.

## Two smaller ones

- **`progressUpdate()` without a preceding `progressStart()` threw NPE** — the `ThreadLocal`s had no initial value. They now hold one. Callers in the plugin always start first, so this was latent, but it is a public method.
- **`progressFinished()` cleared `imageLines` but not `updateCount`**, unlike `progressStart()` which clears both. A following run therefore continued the old hash counter. Now symmetric.

## Verified against Docker 29.6.2

| `-DoutputFile=` | Before | After |
|---|---|---|
| `target/logs/nested/build.log` (parent missing) | `Cannot invoke "java.io.PrintWriter.println(String)" because "this.pw" is null` | directory created, log written |
| a non-empty directory | the same NullPointerException | `Cannot write to output file target\nonempty: (access denied)` |

```
./mvnw clean install     # Temurin 21
→ Tests run: 1010, Failures: 0, Errors: 0, Skipped: 6  —  BUILD SUCCESS
```

Five tests added, including one that pins the ordering fix by asserting the global `Ansi` state is untouched when construction fails. `StopMojoTest` and the other Mojo tests that go through `execute()` pass unchanged.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
+++ b/src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
@@ -51,8 +51,8 @@ public class AnsiLogger implements Logger, Closeable {
 
 
     // Map remembering lines
-    private ThreadLocal<Map<String, Integer>> imageLines = new ThreadLocal<>();
-    private ThreadLocal<AtomicInteger> updateCount = new ThreadLocal<>();
+    private ThreadLocal<Map<String, Integer>> imageLines = ThreadLocal.withInitial(HashMap::new);
+    private ThreadLocal<AtomicInteger> updateCount = ThreadLocal.withInitial(AtomicInteger::new);
 
     // Whether to use ANSI codes
     private boolean useAnsi;
@@ -70,6 +70,10 @@ public class AnsiLogger implements Logger, Closeable {
         this(log, useColor, verbose, batchMode, prefix, null);
     }
 
+    /**
+     * @param outpufFile file to write the log to instead of the console, or <code>null</code> for the console
+     * @throws UncheckedIOException if the output file cannot be opened for writing
+     */
     public AnsiLogger(Log log, boolean useColor, String verbose, boolean batchMode, String prefix, File outpufFile) {
         this.log = log;
         this.prefix = prefix;
@@ -80,12 +84,10 @@ public class AnsiLogger implements Logger, Closeable {
             this.batchMode = true;
         }
         checkVerboseLoggingEnabled(verbose);
+        // Open the output file before initializeColor() touches the global Ansi state: if opening fails,
+        // the caller aborts and would otherwise leave that state switched for the rest of the build.
+        initializePrintWriter();
         initializeColor(useColor);
-        try {
-            initializePrintWriter();
-        } catch (FileNotFoundException e) {
-            log.error(prefix + "Cannot write log output to " + outputFile, e);
-        }
     }
 
     /** {@inheritDoc} */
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/maven/docker/util/AnsiLogger.java
```diff
@@ -70,6 +70,10 @@ public AnsiLogger(Log log, boolean useColor, String verbose, boolean batchMode,
         this(log, useColor, verbose, batchMode, prefix, null);
     }
 
+    /**
+     * @param outpufFile file to write the log to instead of the console, or <code>null</code> for the console
+     * @throws UncheckedIOException if the output file cannot be opened for writing
+     */
```
rohanKanojia, line 76: this javadoc doesn't look right?

JamBalaya56562: Good catch, though the odd part is not the javadoc — `outpufFile` is the actual
parameter name, and has been since the constructor was added, so `@param`
matched it. Adding the javadoc just brought the typo into view.

Renamed the parameter to `outputFile` in a follow-up commit and lined the
javadoc up with it. The name appeared only in this constructor, which this PR
already touches.

If the javadoc itself is what looked wrong — one parameter documented out of
six, where the other constructors have none — say the word and I will drop the
`@param` and keep only the `@throws`, which is the part this PR actually needed.

rohanKanojia (approved):

PR #1976, opened 2026-08-25
Gate the Sonar PR publish on a maintainer approval instead of trusting fork code

JamBalaya56562:
### What

Every run of `Sonar PR Analysis Publish` fails, and has done for a while — the last eight are all `failure` or `cancelled`:

```
##[error]Refusing to check out fork pull request code from a 'workflow_run' workflow.
This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache
scope, and runner access. Fetching and executing a fork's code in that trusted context
commonly leads to "pwn request" vulnerabilities.
To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set
'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

Nothing is broken about builds or tests — the effect is that pull requests from forks get no Sonar analysis.

### Why `actions/checkout` is right

The job holds `SONAR_TOKEN` and a `GITHUB_TOKEN`, checks out `refs/pull/N/head` and runs `./mvnw clean install` over it. A pull request from a fork controls that code, so it could read those secrets out of a test or a Maven plugin. Setting `allow-unsafe-pr-checkout: true` on its own would just reopen the hole that was closed.

### How

The fork's code is built only once a human who can merge it has looked at the very commit that is about to run. A `gate` job:

1. resolves the pull request from the head repository and branch,
2. requires its head to *still* be the commit this run was triggered for, and
3. requires an `APPROVED` review from an `OWNER`, `MEMBER` or `COLLABORATOR` carrying **that same commit id**.

Anything short of that **skips** the analysis rather than failing it, with a notice saying which condition was not met. Because the approval is tied to the commit, pushing after being approved closes the gate again. The checkout that follows is pinned to that commit and asserted with `git rev-parse HEAD` afterwards, so `allow-unsafe-pr-checkout` is only reached on the far side of a human decision.

**Pull requests from branches of this repository skip the check entirely** — only someone with write access can push there, so there is nothing to gate. Dependabot's pull requests keep being analysed automatically, as before.

The gate calls three endpoints: `GET /pulls?head=…` to find the pull request, `GET /pulls/{n}/reviews` to read its reviews, and `GET /repos/{owner}/{repo}/collaborators/{username}/permission` for each login that approved the head commit.

"Maintainer" means write or admin as the repository itself reports it, not the review's `author_association` — `MEMBER` only says the reviewer is in the organisation, and `COLLABORATOR` is just as true of read or triage access. An approval that resolves to anything else, or a lookup the API declines to answer, leaves the gate closed.

The REST reference for that permission endpoint says the caller needs push access, which reads like more than this job should hold. That line applies to user-to-server and classic tokens; for a workflow token the [GitHub Apps permission reference](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps) lists it as `Metadata: read`, which every `GITHUB_TOKEN` already has. Checked on a workflow declaring exactly the `contents: read` and `pull-requests: read` above:

```
declared permissions: contents: read + pull-requests: read
== own repo, the pushing user ==
OK -> admin
== own repo, an unrelated login ==
OK -> read
```

### Two other faults in the same file

* **The concurrency group cancelled unrelated pull requests.** It keyed on `github.ref`, which in a `workflow_run` event resolves to the default branch — every publish run of this workflow reports `head_branch=master`, so all pull requests shared one group with `cancel-in-progress: true`. That matches the `cancelled` entries in the run history. It now keys on the head repository and branch.
* **The pull request number was resolved without any check.** It came from `sender.login` plus a branch name, taking `.[0].number`. Run 32851681701 shows the result: a run for branch `fix-ansilogger-robustness` (PR #1964) carried `PR_NUMBER=1973`. The new lookup uses the head repository owner and only accepts a pull request whose head SHA matches the commit being analysed.

The debug step dumping the entire event payload is gone; the gate reports what it decided and why.

### Verification

A fork cannot exercise `workflow_run` against this repository, so the gate script was extracted from the YAML and run locally against **recorded GitHub API responses**. The fork case uses the real reviews on #1964 (two `COMMENTED`, one `MEMBER` and one `CONTRIBUTOR`); the approval cases are those same objects with `state` and `commit_id` altered, so the schema is genuine.

| scenario | `analyse` |
| --- | --- |
| head is a branch of this repository (Dependabot #1971) | `true` — gate skipped |
| fork, approved by a `MEMBER`, commit id matches | `true` |
| fork, approved by a `MEMBER`, but for an earlier commit | `false` |
| fork, only `COMMENTED` reviews (the real state of #1964) | `false` |
| fork, `APPROVED` by a `CONTRIBUTOR` | `false` |
| no open pull request whose head is this commit | `false` |

The workflow was also parsed to confirm the job graph, `needs`/`if` wiring and permissions, and `sonar-pr-request.yml` is untouched.

**What still needs checking after merge**, since it cannot be exercised from a fork: that a fork pull request without approval shows the `sonar` job as skipped, that approving it makes the next run analyse, and that a Dependabot pull request passes straight through.

### DIFF
--- a/.github/workflows/sonar-pr-analysis-publish.yml
+++ b/.github/workflows/sonar-pr-analysis-publish.yml
@@ -10,42 +10,115 @@ on:
       - completed
 
 concurrency:
-  # Only run once for latest commit per ref and cancel other (previous) runs.
-  group: ci-sonar-docker-maven-plugin-${{ github.ref }}
+  # In a workflow_run event github.ref resolves to the default branch, so keying on it would put
+  # every pull request into one group and let them cancel each other. Key on the head instead.
+  group: ci-sonar-docker-maven-plugin-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
   cancel-in-progress: true
 
 permissions:
   contents: read
+  pull-requests: read
 
 jobs:
+  gate:
+    name: Check whether the commit may be analysed
+    runs-on: ubuntu-latest
+    outputs:
+      analyse: ${{ steps.check.outputs.analyse }}
+      pr: ${{ steps.check.outputs.pr }}
+      sha: ${{ steps.check.outputs.sha }}
+    steps:
+      # This workflow holds the base repository's secrets, and the job below builds and tests the
+      # commit it checks out. For a fork that is only safe once a human who can merge the pull
+      # request has looked at the very code that is about to run, so establish that first.
+      - name: Check
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BASE_REPO: ${{ github.repository }}
+          HEAD_REPO: ${{ github.event.workflow_run.head_repository.full_name }}
+          HEAD_OWNER: ${{ github.event.workflow_run.head_repository.owner.login }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+        run: |
+          set -eu
+
+          skip() {
+            echo "analyse=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::$1"
+            exit 0
+          }
+
+          echo "sha=$HEAD_SHA" >> "$GITHUB_OUTPUT"
+
+          # Identify the pull request by its head, and only accept one that still points at the
+          # commit this run was triggered for.
+          pr=$(gh api "repos/$BASE_REPO/pulls?head=$HEAD_OWNER:$HEAD_BRANCH&state=open" \
+                 --jq '[.[] | select(.head.sha == env.HEAD_SHA)] | .[0].number // empty')
+          if [ -z "$pr" ]; then
+            skip "No open pull request has $HEAD_SHA as its head; nothing to analyse."
+          fi
+          echo "pr=$pr" >> "$GITHUB_OUTPUT"
+
+          # A branch in this repository can only be pushed by someone who already has write access,
+          # so there is no untrusted code here to gate - this covers Dependabot as well.
+          if [ "$HEAD_REPO" = "$BASE_REPO" ]; then
+            echo "analyse=true" >> "$GITHUB_OUTPUT"
+            echo "::notice::PR #$pr comes from this repository, analysing $HEAD_SHA."
+            exit 0
+          fi
+
+          # The approval has to name this exact commit, so pushing after being approved closes the
+          # gate again rather than smuggling code past it.
+          approvers=$(gh api "repos/$BASE_REPO/pulls/$pr/reviews" --paginate \
+            --jq '.[]
+                  | select(.state == "APPROVED" and .commit_id == env.HEAD_SHA)
+                  | select(.author_association == "OWNER"
+                        or .author_association == "MEMBER"
+                        or .author_association == "COLLABORATOR")
+                  | .user.login')
+          approver=$(printf '%s\n' "$approvers" | head -n 1)
+          if [ -z "$approver" ]; then
+            skip "PR #$pr is from a fork and $HEAD_SHA carries no approval from a maintainer, skipping the analysis."
+          fi
+
+          echo "analyse=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


.github/workflows/sonar-pr-analysis-publish.yml
```diff
@@ -10,42 +10,115 @@ on:
...
+          fi
+
+          # The approval has to name this exact commit, so pushing after being approved closes the
+          # gate again rather than smuggling code past it.
+          approvers=$(gh api "repos/$BASE_REPO/pulls/$pr/reviews" --paginate \
+            --jq '.[]
+                  | select(.state == "APPROVED" and .commit_id == env.HEAD_SHA)
+                  | select(.author_association == "OWNER"
+                        or .author_association == "MEMBER"
+                        or .author_association == "COLLABORATOR")
```
rohanKanojia: Could we tighten the trust boundary here? `author_association` is broader than the “human who can merge” described in the comment above. In particular, `COLLABORATOR` and `MEMBER` don't necessarily imply write/maintain access to this repository.

JamBalaya56562: Good catch — `author_association` doesn't answer the question the comment above it claims to answer. `MEMBER` only says the reviewer is in the org, and `COLLABORATOR` is just as true of read or triage access; neither can merge this PR.

Pushed a follow-up that drops the association check and resolves every login that approved the exact head commit through `repos/{repo}/collaborators/{login}/permission` instead, accepting only `write` (what the API reports for `maintain` too) or `admin`. Anything else, including a lookup the API declines to answer, leaves the gate closed.

.github/workflows/sonar-pr-analysis-publish.yml
```diff
@@ -10,42 +10,135 @@ on:
...
-  # Only run once for latest commit per ref and cancel other (previous) runs.
-  group: ci-sonar-docker-maven-plugin-${{ github.ref }}
+  # In a workflow_run event github.ref resolves to the default branch, so keying on it would put
+  # every pull request into one group and let them cancel each other. Key on the head instead.
+  group: ci-sonar-docker-maven-plugin-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
   cancel-in-progress: true
 
 permissions:
   contents: read
+  pull-requests: read
```
rohanKanojia, line 20: Question about the token here, and I might be misreading the docs.

`GET /repos/{owner}/{repo}/collaborators/{login}/permission` is documented as needing push access on the caller. The job's token has `contents: read` + `pull-requests: read`; would that count as push access? My reading is that it wouldn't, and the call would 403, but I haven't tried it.

JamBalaya56562: It is not called, and the description was misleading about that — now reworded,
with the two endpoints the gate actually uses stated up front.

The gate reads only `GET /pulls?head=…` and `GET /pulls/{n}/reviews`, both
covered by the `pull-requests: read` it declares.

Your reading of the docs matches mine, and it is exactly why that endpoint is
avoided: it wants push access on the caller, which is more than this job should
hold to make a read-only decision. So the check settles for the review's
`author_association` instead, with the trade-off that `MEMBER` means "in the
fabric8io organisation" rather than strictly "has write access here".

JamBalaya56562: Correction to my previous answer: it **is** called. I checked a stale local copy
of the branch, from before the follow-up that replaced `author_association` with
that endpoint, and told you the opposite. Sorry for the detour — the description
has been fixed too, it carried the same claim.

To your actual question, the answer is no, it does not 403. The "push access"
line is in the REST reference, and it applies to user-to-server and classic
tokens; for a workflow token the [GitHub Apps permission reference](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps)
lists this endpoint as `Metadata: read`, which every `GITHUB_TOKEN` has already.

Ran it on a workflow declaring exactly the two permissions this job does:

```
declared permissions: contents: read + pull-requests: read
== own repo, the pushing user ==
OK -> admin
== own repo, an unrelated login ==
OK -> read
```

So the gate can ask the repository what access an approver really has, which is
what you asked for in the other thread, without the job holding write anywhere.
And if a lookup ever is declined, that path is already handled: it logs a warning
and leaves the gate closed rather than falling back to something looser.

PR #1977, opened 2026-08-31
Add a retry configuration option for docker start

geoand:
This is done because in Quarkus we've seen flakiness in CI when starting a docker container

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/RunService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/RunService.java
@@ -34,6 +34,7 @@ import java.util.Set;
 import java.util.StringJoiner;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicReference;
 
 import io.fabric8.maven.docker.access.ContainerCreateConfig;
 import io.fabric8.maven.docker.access.ContainerHostConfig;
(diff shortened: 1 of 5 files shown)


src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -187,15 +187,69 @@ public String createAndStartContainer(ImageConfiguration imageConfig,
...
-        if (portMapping.needsPropertiesUpdate()) {
-            updateMappedPortsAndAddresses(id, portMapping);
+    /**
+     * Create and start a container with the given image configuration, retrying on transient failures.
+     *
+     * @param imageConfig image configuration holding the run information and the image name
+     * @param portMapping container port mapping
+     * @param gavLabel label to tag the started container with
+     * @param properties properties to fill in with dynamically assigned ports
+     * @param defaultContainerNamePattern pattern to use for naming containers. Can be null in which case a default pattern is used
```
rohanKanojia, line 204: `@param baseDir` looks like it's missing from the new overload's javadoc.

src/main/java/io/fabric8/maven/docker/service/RunService.java
```diff
@@ -187,15 +187,69 @@ public String createAndStartContainer(ImageConfiguration imageConfig,
...
+                                          PortMapping portMapping,
+                                          GavLabel gavLabel,
+                                          Properties properties,
+                                          File baseDir,
+                                          String defaultContainerNamePattern,
+                                          Date buildTimestamp,
+                                          int startRetries) throws DockerAccessException {
+        // 0: the original attempt, 1..startRetries: possible retries.
+        int attempt = 0;
+        while (true) {
```
rohanKanojia: We already pull in Failsafe and use it right below in `updateMappedPortsAndAddresses` with `.withBackoff(10, 100, ChronoUnit.MILLIS)`. Do you think we can reuse that here so both retry paths in this class look the same?

geoand:
Thanks. I'll have a look in a few days when I'm back

geoand:
@rohanKanojia PR updated

rohanKanojia (approved):

PR #1978, opened 2026-08-31
Make the buildx driver configurable and support Docker Build Cloud

romank8k:
### What

Makes the buildx driver configurable instead of hard-coding `docker-container`, and adds first-class support for building on [Docker Build Cloud](https://docs.docker.com/build-cloud/).

New `<buildx>` option:

```xml
<buildx>
  <driver>cloud</driver>
  <builderName>my-org/my-builder</builderName>
  <driverOpts>
    <key>value</key>
  </driverOpts>
</buildx>
```

### Why

`createCustomBuilderIfNotExists` always passed `--driver docker-container`, so there was no way to target any other buildx driver. Docker Build Cloud builders in particular could not be used at all, even though they are just another driver from the CLI's point of view.

### Details

- `<driver>` defaults to `docker-container`, preserving today's behaviour exactly when unset.
- For `driver=cloud`, `docker buildx create` takes the endpoint as a **positional** argument rather than `--name`, so `<builderName>` is treated as the cloud endpoint and must be in `<org>/<name>` form; a clear `MojoExecutionException` is thrown otherwise.
- Docker derives the local builder name for a cloud builder as `cloud-<org>-<name>`, so that name is used for the instance-path existence check and passed on to the build.
- The `cloud` driver manages its own nodes and build configuration, so `--node` and the buildx config file are not applied to it.
- `<driver>` and `<driverOpts>` are wired through `ConfigKey`/`PropertyConfigHandler`, and can additionally be overridden at invocation time with `-Ddocker.buildx.driver` and `-Ddocker.buildx.driverOpts.<key>=<value>` (per-key merge over the XML config). `-Ddocker.buildx.builderName` is honoured on the same path.
- `createConfigJson` now runs *before* `createBuilder`, so builder creation can see registry credentials. Creating a cloud builder contacts the registry, which failed when `config.json` was only written afterwards.

### Documentation

`src/main/asciidoc/inc/build/_buildx.adoc` and `inc/external/_property_configuration.adoc` document the new element and properties.

### Tests

New `BuildXServiceCreateBuilderTest` (11 cases) covers the default driver, an explicit non-cloud driver, cloud endpoint handling and derived naming, the `<org>/<name>` validation failure, driver-opt merging, and the system-property overrides. `BuildXServiceTest` and `PropertyConfigHandlerTest` pass unchanged.

All commits are signed off per `CONTRIBUTING.md`.

### DIFF
--- a/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
@@ -15,6 +15,7 @@ import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -293,8 +294,8 @@ public class BuildXService {
 
     protected String createBuilder(Path configPath, List<String> buildX, ImageConfiguration imageConfig, BuildDirs buildDirs) throws MojoExecutionException {
         BuildXConfiguration buildXConfiguration = imageConfig.getBuildConfiguration().getBuildX();
-        String builderEndpoint = Optional.ofNullable(buildXConfiguration.getBuilderName()).orElse("maven");
-        String driver = buildXConfiguration.getDriver();
+        String builderEndpoint = resolveBuilderName(buildXConfiguration);
+        String driver = resolveDriver(buildXConfiguration);
 
         // For the cloud driver Docker derives the local builder name as "cloud-<org>-<name>"
         String builderName = isCloudDriver(driver)
@@ -304,21 +305,35 @@ public class BuildXService {
         if ("default".equals(builderName)) {
             logger.info("Using default builder with buildx - only single platforms will be supported");
         } else {
-            createCustomBuilderIfNotExists(configPath, buildX, buildXConfiguration, builderName, builderEndpoint, buildDirs);
+            if (isCloudDriver(driver) && !builderEndpoint.contains("/")) {
+                throw new MojoExecutionException(
+                    "Cloud builder identifier must be in <org>/<name> form, got: " + builderEndpoint);
+            }
+            createCustomBuilderIfNotExists(configPath, buildX, buildXConfiguration, builderName, builderEndpoint, driver, buildDirs);
         }
         return builderName;
+    }
+
+    private static String resolveBuilderName(BuildXConfiguration buildXConfiguration) {
+        String sysProp = System.getProperty("docker.buildx.builderName");
+        return sysProp != null ? sysProp : Optional.ofNullable(buildXConfiguration.getBuilderName()).orElse("maven");
+    }
+
+    // System property takes priority over XML config, matching the pattern used by ConfigHelper.isNoCache / getNetwork
+    private static String resolveDriver(BuildXConfiguration buildXConfiguration) {
+        String sysProp = System.getProperty("docker.buildx.driver");
+        return sysProp != null ? sysProp : buildXConfiguration.getDriver();
     }
 
     private static boolean isCloudDriver(String driver) {
         return "cloud".equals(driver);
     }
 
-    private void createCustomBuilderIfNotExists(Path configPath, List<String> buildX, BuildXConfiguration buildXConfiguration, String builderName, String builderEndpoint, BuildDirs buildDirs) throws MojoExecutionException {
+    private void createCustomBuilderIfNotExists(Path configPath, List<String> buildX, BuildXConfiguration buildXConfiguration, String builderName, String builderEndpoint, String driver, BuildDirs buildDirs) throws MojoExecutionException {
         Path builderPath = configPath.resolve(Paths.get("buildx", "instances", builderName.toLowerCase()));
 
         if (Files.notExists(builderPath)) {
             List<String> cmds = new ArrayList<>(buildX);
-            String driver = buildXConfiguration.getDriver();
 
             if (isCloudDriver(driver)) {
                 // Cloud driver: endpoint is a positional argument; --name and --node are not used
@@ -342,9 +357,16 @@ public class BuildXService {
     }
 
     private void addDriverOptions(List<String> cmds, BuildXConfiguration buildXConfiguration) {
-        if (buildXConfiguration.getDriverOpts() != null && !buildXConfiguration.getDriverOpts().isEmpty()) {
-            buildXConfiguration.getDriverOpts().forEach((key, value) -> append(cmds, "--driver-opt", key + '=' + value));
+        Map<String, String> opts = new LinkedHashMap<>();
+        if (buildXConfiguration.getDriverOpts() != null) {
+            opts.putAll(buildXConfiguration.getDriverOpts());
         }
+        // -Ddocker.buildx.driverOpts.<key>=<value> overrides individual XML-configured entries
+        Map<String, String> sysOpts = EnvUtil.extractFromPropertiesAsMap("docker.buildx.driverOpts", System.getProperties());
+        if (sysOpts != null) {
+            opts.putAll(sysOpts);
+        }
+        opts.forEach((key, value) -> append(cmds, "--driver-opt", key + '=' + value));
     }
 
     private void addBuildConfig(List<String> cmds, BuildXConfiguration buildXConfiguration, BuildDirs buildDirs) {
--- a/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
+++ b/src/main/java/io/fabric8/maven/docker/service/BuildXService.java
@@ -15,13 +15,13 @@ import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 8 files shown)


src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -293,31 +294,61 @@ protected void createDirectory(Path cachePath) {
...
+                    "Cloud builder identifier must be in <org>/<name> form, got: " + builderEndpoint);
+            }
+            createCustomBuilderIfNotExists(configPath, buildX, buildXConfiguration, builderName, builderEndpoint, driver, buildDirs);
         }
         return builderName;
     }
 
-    private void createCustomBuilderIfNotExists(Path configPath, List<String> buildX, BuildXConfiguration buildXConfiguration, String builderName, BuildDirs buildDirs) throws MojoExecutionException {
-        String nodeName = buildXConfiguration.getNodeName();
+    private static String resolveBuilderName(BuildXConfiguration buildXConfiguration) {
```
rohanKanojia: These same three keys get wired into PropertyConfigHandler in this PR, so `-Ddocker.buildx.driver` now resolves twice through different machinery. The ValueProvider path honours PropertyMode ... someone on Fallback or Skip expects XML to win, but resolveDriver makes the `-D` win unconditionally. Was that intentional, or would leaning entirely on the ConfigKey wiring be simpler?

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -326,9 +357,16 @@ private void createCustomBuilderIfNotExists(Path configPath, List<String> buildX
...
 
     private void addDriverOptions(List<String> cmds, BuildXConfiguration buildXConfiguration) {
-        if (buildXConfiguration.getDriverOpts() != null && !buildXConfiguration.getDriverOpts().isEmpty()) {
-            buildXConfiguration.getDriverOpts().forEach((key, value) -> append(cmds, "--driver-opt", key + '=' + value));
+        Map<String, String> opts = new LinkedHashMap<>();
+        if (buildXConfiguration.getDriverOpts() != null) {
+            opts.putAll(buildXConfiguration.getDriverOpts());
+        }
+        // -Ddocker.buildx.driverOpts.<key>=<value> overrides individual XML-configured entries
+        Map<String, String> sysOpts = EnvUtil.extractFromPropertiesAsMap("docker.buildx.driverOpts", System.getProperties());
```
rohanKanojia: Should this go through `EnvUtil.getPropertiesWithSystemOverrides(project)` rather than `System.getProperties()`?

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -78,10 +79,10 @@ protected <C> void useBuilder(ProjectPaths projectPaths, ImageConfiguration imag
             copyBuildXToConfigPathIfBuildXBinaryInDefaultDockerConfig(configPath);
         }
 
-        String builderName = createBuilder(configPath, buildX, imageConfig, buildDirs);
         Path configJson = configPath.resolve("config.json");
         try {
             createConfigJson(configJson, authConfig);
+            String builderName = createBuilder(configPath, buildX, imageConfig, buildDirs);
```
rohanKanojia, line 89: This affects every buildx build, not just cloud ... is that deliberate?

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -293,31 +294,61 @@ protected void createDirectory(Path cachePath) {
...
+        // For the cloud driver Docker derives the local builder name as "cloud-<org>-<name>"
+        String builderName = isCloudDriver(driver)
+            ? "cloud-" + builderEndpoint.replace("/", "-").toLowerCase()
+            : builderEndpoint;
 
         if ("default".equals(builderName)) {
             logger.info("Using default builder with buildx - only single platforms will be supported");
         } else {
-            createCustomBuilderIfNotExists(configPath, buildX, buildXConfiguration, builderName, buildDirs);
+            if (isCloudDriver(driver) && !builderEndpoint.contains("/")) {
```
rohanKanojia: Is `contains("/")` strict enough? `a/b/c`, `org/` and bare `/ `all pass.

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
```diff
@@ -293,31 +294,61 @@ protected void createDirectory(Path cachePath) {
...
+    private static String resolveDriver(BuildXConfiguration buildXConfiguration) {
+        String sysProp = System.getProperty("docker.buildx.driver");
+        return sysProp != null ? sysProp : buildXConfiguration.getDriver();
+    }
+
+    private static boolean isCloudDriver(String driver) {
+        return "cloud".equals(driver);
+    }
+
+    private void createCustomBuilderIfNotExists(Path configPath, List<String> buildX, BuildXConfiguration buildXConfiguration, String builderName, String builderEndpoint, String driver, BuildDirs buildDirs) throws MojoExecutionException {
```
rohanKanojia, line 325: Does the cloud path survive a second run? The default state dir lives under the build directory, so after mvn clean this is notExists again and we re-run buildx create ... for a cloud endpoint that already exists remotely, is that idempotent or does it error?

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
romank8k: Not intentional as a policy decision, just me adding a shortcut on top of the wiring — and you are right that it breaks `PropertyMode`. Removed in 29ea295: `resolveDriver`, `resolveBuilderName` and the `driverOpts` system-property merge are gone, and `BuildXService` reads only `buildXConfiguration`. `docker.buildx.driver` / `docker.buildx.driverOpts.KEY` now resolve solely through `ConfigKey`/`PropertyConfigHandler`, so `Override`/`Fallback`/`Skip` and `_combine` all behave as they do for every other key. `docker.buildx.builderName` is back to its pre-PR behaviour (no system-property read at all).

Added `PropertyConfigHandlerTest#testBuildXDriverAndDriverOpts` and `#testBuildXDriverOptsMergeWithConfiguredOnes` to cover that path, and dropped the three system-property tests from `BuildXServiceCreateBuilderTest`.

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
romank8k: Moot now — that whole block is gone in 29ea295. `addDriverOptions` is back to the original XML-only version, and the merge of `docker.buildx.driverOpts.KEY` over `<driverOpts>` is handled by `ValueProvider.getMap` with `ValueCombinePolicy.Merge`, which reads `EnvUtil.getPropertiesWithSystemOverrides(project)` via the config handler.

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
romank8k: Deliberate, and it needs to be: creating a `cloud` builder contacts the registry, so `config.json` has to exist before `docker buildx create` runs.

The blast radius on the other drivers is nil, though: `createConfigJson` only writes the file into the per-build state dir, it does not touch the daemon or the builder, and `docker buildx create --driver docker-container` never reads registry credentials. So for existing builds the only change is that the file is on disk a few milliseconds earlier. `removeConfigJson` is still in the same `finally`, and moving `createBuilder` inside the `try` also means the file is cleaned up if builder creation fails, which was not the case before.

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
romank8k: No, it is not — good catch. 29ea295 replaces it with an anchored pattern requiring exactly two non-empty segments:

```java
private static final Pattern CLOUD_ENDPOINT_PATTERN =
    Pattern.compile("[A-Za-z0-9][A-Za-z0-9._-]*/[A-Za-z0-9][A-Za-z0-9._-]*");
```

`cloudDriverRequiresOrgSlashNameFormat` is now a `@ParameterizedTest` over `invalid-no-slash`, `myorg/`, `/default`, `/`, `myorg/team/default` and `my org/default`, and it asserts on the message rather than just the exception type.

src/main/java/io/fabric8/maven/docker/service/BuildXService.java
romank8k: Verified that it's idempotent if the cloud endpoint already exists. `docker buildx create --driver cloud <org>/<name>` does not provision anything remotely — the cloud builder is created out-of-band in Docker Build Cloud, and `create` only registers a local reference to it in the buildx instance store. That store lives under the `--config` directory the plugin passes, i.e. the state dir under the build directory, so after `mvn clean` there is no local instance, `Files.notExists` is true, and re-registering an endpoint that exists remotely succeeds.

romank8k:
Hi @rohanKanojia,
I have addressed your feedback.
If it looks good, I can squash the commits to get it ready for merging.

rohanKanojia (approved):

